PR #87, opened 2022-08-03
Dapps catalog

IvoPaunov:
Added dapps catalog  default lists, service and hook


superKalo (approved):
I'm not sure if the mobile implementation will be compatible with this flow (based on what we discussed with Emo), but as soon as 1) we have the designs ready 2) we validate if the same (or similar) flow will be used on the web app; we're going to dive into it.

So far, with the latest small improvements we discussed offline with @IvoPaunov , it looks all right to me 👍

PR #117, opened 2022-09-08
Feature / Portfolio hook v2

gergana95:
Implementation of portfolio v2 uses velcro-v2. The main purpose here is to speed up the balances loading and to include regular prices fetching from coingecko.

This is the issue in ambire-wallet for [portfolio-v2](https://github.com/AmbireTech/ambire-wallet/issues/848).

Second goal with this feature change is adding and displaying pending balances. [Issue](https://github.com/AmbireTech/ambire-wallet/issues/853)
We implement this by using the balance oracle with 3 requests and two states:
1 - latest - fetching latest balance
2 - pending - fetching pending balance for all network tokens
3 - pending (unsigned transactions) - fetching pending balance for all network tokens with passed unsigned transactions.

Implement Hide Collectibles - [Issue](https://github.com/AmbireTech/ambire-wallet/issues/777)

Change order of requests and order them in ambire-wallet by dateAdded, to do that we add timestamp for dateAdded on each addRequest

Implement caching with useCacheStorage hook on ambire-wallet with indexedDB

All related changes can be found [here](https://github.com/AmbireTech/ambire-wallet/pull/1050), since [this](https://github.com/AmbireTech/ambire-wallet/pull/926) one was deprecated earlier.


jordan-enev:
After 5-10 minutes on Dashboard (without routing to other tabs), the portfolio v2 branch had a crash with the following error:

jordan-enev (commented):
Really good and solid work! :bow: :fire: :1st_place_medal: 

The way you simplified `usePortfolio` in smaller hooks + caching derived data with `useMemo` (instead of using `useEffect` + state updates) results in a much simple and better performing code.

:clap: :clap: :clap: 

From what I've seen, the only hook that is still a bit more complicated is [src/hooks/usePortfolio/usePortfolioFetch.ts](https://github.com/AmbireTech/ambire-common/pull/117/files#diff-00ae3e263957bf8dbcaaaee5eeb5da1872fcdb2e140609895f4fcfa0be46e6b3) and as a first step we can try document its flow and use-cases (its logic is really crazy). Later, we can try figurring out how we can simplify it.

jordan-enev:
For some reason, the Protocol chart is not showing 100% (it works correctly on Production):

image

src/hooks/usePortfolio/usePortfolio.ts
```diff
@@ -671,41 +142,60 @@ export default function usePortfolio({
...
+    const refreshInterval = setInterval(() => {
+      updateCoingeckoAndSupplementData(currentAssets)
+    }, 20000)
     return () => clearInterval(refreshInterval)
-  }, [fetchSupplementTokenData, tokensByNetworks])
+  }, [requests, eligibleRequests, pendingTransactions])
+
+  useEffect(() => {
+    fetchAndSetSupplementTokenData(currentAssets)
+  }, [requests, `${eligibleRequests}`, `${pendingTransactions}`])
```
jordan-enev, line 169: Why string interpolation?

src/hooks/usePortfolio/useTransactions.ts
```diff
@@ -0,0 +1,23 @@
...
+    // Pending transactions which aren't executed yet
+    const allPending = useMemo(() => data && data.txns.filter(x => !x.executed && !x.replaced)
+    , [data?.txns])
+
+    useEffect(() => {
+        forceRefresh()
+    },[requests, sentTxn])
+
+    return {
+        pendingTransactions: allPending?.length ? [allPending] : [],
```
jordan-enev, line 21: Why array in array?

PR #122, opened 2022-09-15
Bug / Remove `package-lock.json` (since project uses yarn)

superKalo:
No need for `package-lock.json` file (outdated anyways), since the project uses yarn and has an `yarn.lock` file anyways.

Plus, fixes the deps ordering (which is the default one that yarn makes).


PetromirDev (approved):
Looks good to me!

PR #127, opened 2022-09-23
Swappin humanizer

Maskln:
Added Swappin humanizer


superKalo (approved):
With @Maskln we discussed that spreading the `...paymentSrcToken` and the `...paymentToken` could cause troubles, because these can be not only objects, but also null, string or number. However, this is a consistent issue with all humanizers, not just this one, so we will handle this in another PR.

Looks good to me 👍

PR #139, opened 2022-10-05
Relayer wallet rewards

stojnovsky:
- remove useConstants from useClaimableWalletToken
- added useRelayerData in useClaimableWalletToken
- get wallets reward data from relayer (including current claimable tokes per user, root and signed root)


src/hooks/useClaimableWalletToken/types.ts
```diff
@@ -1,9 +1,9 @@
 import { Account } from '../useAccounts'
-import { UseConstantsReturnType } from '../useConstants'
 import { UseNetworkReturnType } from '../useNetwork'
 
 export type UseClaimableWalletTokenProps = {
-  useConstants: () => UseConstantsReturnType
+  relayerURL: string,
+  useRelayerData: any,
```
superKalo, line 6: The type here should be:

```
useRelayerData: (props: Omit<UseRelayerDataProps, 'fetch'>) => UseRelayerDataReturnType
```

src/hooks/useClaimableWalletToken/useClaimableWalletToken.ts
```diff
@@ -9,30 +9,40 @@ import { getProvider } from '../../services/provider'
...
   const initialClaimableEntry = useMemo(() => {
-    if (!constants?.WALLETInitialClaimableRewards) {
+    if (!claimableRewardsData) {
       return null
     }
 
-    return constants.WALLETInitialClaimableRewards.find((x) => x.addr === accountId)
-  }, [accountId, constants?.WALLETInitialClaimableRewards])
+    return claimableRewardsData //.find((x) => x.addr === accountId)
+  }, [accountId, claimableRewardsData])
```
superKalo, line 45: Then `accountId` is an unnecessary dependency.

src/hooks/useClaimableWalletToken/useClaimableWalletToken.ts
```diff
@@ -9,30 +9,40 @@ import { getProvider } from '../../services/provider'
...
   )
   const initialClaimableEntry = useMemo(() => {
-    if (!constants?.WALLETInitialClaimableRewards) {
+    if (!claimableRewardsData) {
       return null
     }
 
-    return constants.WALLETInitialClaimableRewards.find((x) => x.addr === accountId)
-  }, [accountId, constants?.WALLETInitialClaimableRewards])
+    return claimableRewardsData //.find((x) => x.addr === accountId)
```
superKalo, line 44: I see no need to have `claimableRewardsData` then. Simply:

```
const claimableRewardsData = rewardsData?.data?.claimableRewardsData || null
```

Will be the same now.

superKalo (approved):
Looks good to me + it solves the mobile app issues with `claimableNow` being very small amount (like `0.0000000002`) 👍

PR #146, opened 2022-10-19
Signed messages storage support

0xdex18:
This PR enables support for signed messaged storage

(In pair with PR https://github.com/AmbireTech/ambire-wallet/pull/1010)

### DIFF
--- a/src/hooks/useSignMessage/types.ts
+++ b/src/hooks/useSignMessage/types.ts
@@ -14,7 +14,7 @@ export type UseSignMessageProps = {
     approveQuickAcc?: (confCode: number) => void
   ) => Promise<any>
   onLastMessageSign: () => void
-  getHardwareWallet: (device?: any) => any,
+  getHardwareWallet: (device?: any) => any
 }
 
 export type UseSignMessageReturnType = {
@@ -32,9 +32,9 @@ export type UseSignMessageReturnType = {
   isTypedData: boolean
   confirmationType: 'email' | 'otp' | null
   dApp: {
-    name: string,
-    description: string,
-    url: string,
-    icons: string[],
+    name: string
+    description: string
+    url: string
+    icons: string[]
   }
 }
--- a/src/hooks/useSignMessage/types.ts
+++ b/src/hooks/useSignMessage/types.ts
@@ -1,3 +1,5 @@
+import { UseStorageType } from 'hooks/useStorage'
+
 import { NetworkType } from '../../constants/networks'
 import { UseAccountsReturnType } from '../useAccounts'
 import { UseToastsReturnType } from '../useToasts'
@@ -15,6 +17,7 @@ export type UseSignMessageProps = {
   ) => Promise<any>
   onLastMessageSign: () => void
   getHardwareWallet: (device?: any) => any
+  useStorage: UseStorageType
 }
 
 export type UseSignMessageReturnType = {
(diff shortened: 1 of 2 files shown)


src/hooks/useSignMessage/useSignMessage.ts
```diff
@@ -271,11 +279,24 @@ const useSignMessage = ({
 
         await verifySignature(toSign, sig, requestedNetwork?.id)
 
-        resolve({ success: true, result: sig })
         addToast('Successfully signed!')
+
+        setSignedMessages( [...signedMessages, {
+            account: account.id,
```
superKalo, line 285: Maybe name this `accountId`, instead of `account`?

It drives me crazy in the web app, when `account` is used sometimes for id (string), sometimes for object (full account).

src/hooks/useSignMessage/types.ts
```diff
@@ -31,4 +32,5 @@ export type UseSignMessageReturnType = {
   requestedChainId: NetworkType['chainId']
   isTypedData: boolean
   confirmationType: 'email' | 'otp' | null
+  dApp: any
```
superKalo, line 35: Please add the type for this return prop. Should be:

```
  accountId: Account['id']
  networkId: NetworkType['id],
  ... and so on :)
```

src/hooks/useSignMessage/types.ts
```diff
@@ -14,7 +14,8 @@ export type UseSignMessageProps = {
     approveQuickAcc?: (confCode: number) => void
   ) => Promise<any>
   onLastMessageSign: () => void
-  getHardwareWallet: (device?: any) => any
+  getHardwareWallet: (device?: any) => any,
+  connections: any[]
```
superKalo, line 18: New prop, `connections`, is added to the hook props. But it's not used anywhere. Is this still needed?

src/hooks/useSignMessage/useSignMessage.ts
```diff
@@ -271,11 +279,24 @@ const useSignMessage = ({
...
+            message: toSign.txn,
+            signature: sig,
+            dApp,
+          }]
+        )
+
         if (everythingToSign.length === 1) {
           !!onLastMessageSign && onLastMessageSign()
         }
+        resolve({ success: true, result: sig })
```
superKalo, line 299: You moved the `resolve()` part here, because on the web app it unmounts the web component and skips executing the logic afterwards, correct?

If so - please add a comment here. So that 1) we don't forget and 2) we give context to the other brave souls who will break or debug this one day.

src/hooks/useSignMessage/useSignMessage.ts
```diff
@@ -318,8 +339,9 @@ const useSignMessage = ({
         // Unfortunately that isn't possible, because isValidSignature only takes a bytes32 hash; so to sign this with
         // a personal message, we need to be signing the hash itself as binary data such that we match 'Ethereum signed message:\n32<hash binary data>' on the contract
 
-        const sig = await (toSign.type === 'eth_signTypedData_v4' ||
-        toSign.type === 'eth_signTypedData'
+        const isTypedData = toSign.type === 'eth_signTypedData_v4' || toSign.type === 'eth_signTypedData'
```
superKalo, line 342: 'isTypedData' is already declared in the upper scope, maybe name it something else 🙄

src/hooks/useSignMessage/useSignMessage.ts
0xdex18: totally Agree!

src/hooks/useSignMessage/useSignMessage.ts
0xdex18: you're right, I thought that it had been removed. I'll just pass isTypedData from upstream

superKalo (approved):

PR #147, opened 2022-10-21
badges-rewards-apy

0xdex18:
Update : added SVG images + commented ADX badge, until we know what we do with it

in concordance with PR 
https://github.com/AmbireTech/ambire-wallet/pull/1015

adding ADX staking rewards badge

adds some logic for multiplier badges (until now only static multipliers),  having a possibility to display APY (see wallet logic), passing APY data through multiplier badge id.

If we want more custom logic for badges in the future, we should refactor the wallet logic as well.
I adjusted with a minimal code change for the moment

### DIFF
--- a/src/constants/multiplierBadges.ts
+++ b/src/constants/multiplierBadges.ts
@@ -35,7 +35,7 @@ export const multiplierBadges: MultiplierBadge[] = [
     link: 'https://blog.ambire.com/win-a-cryptotesters-nft-with-ambire-and-get-into-one-of-the-hottest-web3-communities-c9d7185760b1',
     icon_svg: 'rewards-crypto-tester'
   },
-  {
+  /*{
     id: 'gasTankNft',
     name: 'GasTankNFT',
     icon: '⛽',
@@ -43,7 +43,7 @@ export const multiplierBadges: MultiplierBadge[] = [
     multiplier: 1.25,
     link: 'https://blog.ambire.com/ambire-gas-tank-launches-with-exclusive-nft-drop-2a4eb29f2f07',
     icon_svg: 'rewards-gas-tank'
-  },
+  },*/
   {
     id: 'powerUserMultiplier',
     name: 'Power User',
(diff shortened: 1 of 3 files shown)


src/constants/multiplierBadges.ts
```diff
@@ -14,40 +15,53 @@ export const multiplierBadges: MultiplierBadge[] = [
...
+    icon_svg: 'reward-power-user'
+  },
+  /*{
+    id: 'adxStakingApy',
+    name: 'ADX staking rewards',
+    icon: 'ADX',
+    color: '#28879c',
+    multiplier: 0,
+    link: 'https://blog.ambire.com/wallet-rewards-mechanism-explained-start-accumulating-value-before-the-token-is-launched-5e9ee36cefdd'
+  },*/
```
superKalo, line 64: Please add some context here in the code why this multiplier is commented out.

src/constants/multiplierBadges.ts
```diff
@@ -14,40 +15,53 @@ export const multiplierBadges: MultiplierBadge[] = [
     icon: '🧪',
     color: '#6000FF',
     multiplier: 1.25,
-    link: 'https://blog.ambire.com/announcing-the-wallet-token-a137aeda9747'
+    link: 'https://blog.ambire.com/announcing-the-wallet-token-a137aeda9747',
+    icon_svg: 'rewards-beta-tester'
```
superKalo, line 19: I see no point in adding `icon_svg` in the common codebase.

1. It's web app specific, so why in common code?

1. On the web app you can wire-up these icons by their `id`. Why introduce another prop?

superKalo: 1. Unfortunately, no. Web SVGs don't work out of the box for native apps.

2. Okay, I see. Please either use "id" or map the `icon_svg` prop in ambire-wallet instead.

superKalo (approved):

PR #168, opened 2022-11-10
Config / Migrate `devDependencies` to `peerDependencies`

superKalo:
These packages should be all listed as `peerDependencies`.

The reason some of them were added as well under `devDependencies` was because they are needed (imported) by your editor when you make changes to the ambire-common.

But still, ambire-common should not require or install these packages. It should always use the parent ones.


package.json
```diff
@@ -2,26 +2,21 @@
...
-    "validator": "^13.7.0"
-  },
-  "devDependencies": {
-    "@types/react": "17.0.39",
-    "@types/validator": "^13.7.0",
-    "@typescript-eslint/eslint-plugin": "5.5.0",
-    "@typescript-eslint/parser": "5.5.0",
-    "adex-protocol-eth": "git+https://git@github.com/AmbireTech/adex-protocol-eth.git#ef2f88f26824c64c1eb12c02a389baec2b2c3889",
-    "eslint": "8.3.0",
+    "eslint": "^7.11.0",
```
superKalo, line 19: Unfortunately, moving those as `peerDependencies` conflicts with the web app deps. Investigating.

PR #175, opened 2022-11-25
Refactor / useSignMessages Hook

sonytooo:


src/hooks/useAccSignMsgStatus/types.ts
```diff
@@ -0,0 +1,17 @@
+import { NetworkId } from 'constants/networks'
+import { Account } from 'hooks/useAccounts'
```
superKalo, line 2: Both should have relative paths: '../'

superKalo (approved):

PR #180, opened 2022-12-02
Config / Store network id under "networkId" key in the (cache) storage

superKalo:
Otherwise, we also have network (which is the full network object) and they conflict.


jordan-enev:
A bit more context here (we had an internal discussion in the office). `network` is currently used/reserved and it contains only the current network id value, not the full network object.

But in terms of Chrome extension, we always have a need of the full network object, but `network` naming/key is already reserved.

So the best naming option is to rename the current key to `networkId` and to introduce a new one `network` for holidng all the network data.

There is one negligible trade-off. If we rename the current key - then we are going to reset the current network of all users to `ethereum`.

Is this fine for you @stojnovsky?

superKalo:
Update: I spoke with Emo, he's fine with this change. We move it forward!

PR #186, opened 2023-01-03
FIx: added missing param for token function in movr.js

Maskln:


src/services/humanizers/Movr.ts
```diff
@@ -20,7 +20,7 @@ const MovrMapping = (humanizerInfo: HumanizerInfoType) => {
       return [
         `Transfer ${token(humanizerInfo, middlewareInputToken, amount)} to ${getNetwork(
           toChainId
-        )} for ${token(humanizerInfo, tokenToBridge)}`
+        )} for ${token(humanizerInfo, tokenToBridge, null)}`
```
jordan-enev, line 23: If we omit the param (`amount`), then its value will be still falsy (`undefined`). Therefore `token` function will return the same output, no matter we pass `null` or `undefined`.

```
// token
if (!amount) return !extended ? assetInfo[0] : extendedToken
```

Did I miss something?

PR #196, opened 2023-02-07
Feature/ dapp search in description and name at the same time

PetromirDev:
Changes:

- search in description and name at the same time, instead of the name only

Closes https://github.com/AmbireTech/ambire-wallet/issues/1145

### DIFF
--- a/src/hooks/useDapps/useDapps.ts
+++ b/src/hooks/useDapps/useDapps.ts
@@ -169,7 +169,9 @@ export default function useDapps({ useStorage, fetch }: UseDappsProps): UseDapps
             match = categoryFilter.filter(item, favorites)
           }
           if (search && match) {
-            match = item.name.toLowerCase().includes(search?.toLowerCase())
+            const matchedName = item.name.toLowerCase().includes(search?.toLowerCase())
+            const matchedDescription = item?.description.toLowerCase().includes(search?.toLowerCase())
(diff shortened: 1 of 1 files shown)


jordan-enev (approved):
@gergana95, @superKalo let's not create a new tag for that one, because it is not a high priority one and we don't rush it. 

Please include it in the next minor/major tag you are releasing.

src/hooks/useDapps/useDapps.ts
```diff
@@ -169,7 +169,9 @@ export default function useDapps({ useStorage, fetch }: UseDappsProps): UseDapps
             match = categoryFilter.filter(item, favorites)
           }
           if (search && match) {
-            match = item.name.toLowerCase().includes(search?.toLowerCase())
+            const matchedName = item.name.toLowerCase().includes(search?.toLowerCase())
+            const matchedDescription = item?.description.toLowerCase().includes(search?.toLowerCase())
```
jordan-enev, line 173: @PetromirDev looking into the above line `item.name....` it looks like that item is alwasys an object. Having this in mind, shouldn't we use the optional chaining operator after the `.description` field, as follows `item.description?.toLowerCase()`

Currently it's working, because all the dApps are having description field, but it would be better to refactor it.

jordan-enev (approved):

PR #197, opened 2023-02-08
Feature / Checks if token is eligible for adding as extra token

superKalo:
Refactors a bit the `onAddExtraToken` method by splitting the check if token is eligible for adding as extra token in a separate method.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.21.4",
+  "version": "0.21.5",
   "name": "ambire-common",
   "description": "Common ground for the Ambire apps",
   "peerDependencies": {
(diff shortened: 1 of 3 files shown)


src/hooks/usePortfolio/useExtraTokens.ts
```diff
@@ -28,21 +29,50 @@ export default function useExtraTokens({
...
+        }
+
       if (
         constants?.tokenList &&
         Object.values(constants.tokenList)
           .flat(1)
-          .map(({ address }: any) => address)
-          .includes(extraTokenAddress)
+          .map((t) => t.address)
+          .includes(address)
```
superKalo, line 49: @gergana95 , here maybe we should do `address.toLowerCase()` and in general - for all similar checks - we should always lowercase the incoming address input. What do you think?

PR #218, opened 2023-03-16
Add: isOffline param in useRelayerData

Maskln:


src/hooks/useRelayerData/useRelayerData.tsx
```diff
@@ -12,12 +12,14 @@ const RESET_DATA_AFTER = 250
...
   url,
-  initialState = null
+  initialState = null,
+  useOfflineStatus = null
 }: UseRelayerDataProps): UseRelayerDataReturnType {
   const [isLoading, setLoading] = useState<boolean>(true)
   const [data, setData] = useState<any>(initialState)
   const [err, setErr] = useState<any>(null)
   const prevUrl = useRef('')
+  const isOffline = (!!useOfflineStatus && useOfflineStatus()) || false
```
PetromirDev, line 22: Why pass the entire hook, call it and get the value here, when you can just pass isOffline and give it a default value of false? `isOffline=false`

You are importing it in the file that you call `useRelayerData` anyway, so just call it there, get the value of isOffline and pass it to the hook.

PetromirDev (changes requested):

PR #222, opened 2023-03-20
Config / OKX and Sepolia chains ⛓️

stojnovsky:
- Added OKX and Sepolina chains
- Added coingeckoId for native assets on all chains
- change adex-protocol-eth version

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.22.9",
+  "version": "0.22.10",
   "name": "ambire-common",
   "description": "Common ground for the Ambire apps",
   "peerDependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "0.22.9",
+  "version": "0.22.10",
```
jordan-enev, line 2: @superKalo would it be better if we create a new minor tag here 0.23, because 0.22.9 > 0.22.10 (it's confusing)

superKalo: Fine by me, we have a couple of tags where we go beyond x.x.10 :)

PR #223, opened 2023-04-20
Feature/contract testing

borislav-itskov:
Change log:
* add hardhat as a dev dependency to have a local node for testing
* add a multisig test

How to run tests:
* npx hardhat node
* npx hardhat test


test/MultiSig.test.js
```diff
@@ -0,0 +1,49 @@
+const { ethers } = require("ethers")
+const { expect } = require("chai")
+const pk1 = "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
+const pk2 = "0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d"
+const addressOne = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
```
Ivshti, line 5: can we use consistent style in regards to quotes (singular quotes)

Ivshti (changes requested):

Ivshti:
@borislav-itskov perhaps also it's a good idea to have a test/AmbireAccount/ directory

contracts/libs/SignatureValidatorV2.sol
```diff
@@ -64,7 +64,8 @@ library SignatureValidator {
 			// s := schnorr signature
 			// parity := public key y-coord parity (27 or 28)
 			// last uint8 is for the Ambire sig mode - it's ignored
-			(bytes32 px, bytes32 e, bytes32 s, uint8 parity,) = abi.decode(sig, (bytes32, bytes32, bytes32, uint8, uint8));
+			sig.trimToSize(sig.length - 1);
+			(bytes32 px, bytes32 e, bytes32 s, uint8 parity) = abi.decode(sig, (bytes32, bytes32, bytes32, uint8));
```
borislav-itskov, line 68: @Ivshti This is a change in the solidity code I'm requesting. It just doesn't work otherwise.  
We remove the Ambire sig mode with trimToSize and afterwards proceed to decoding

contracts/AmbireAccount.sol
```diff
@@ -140,7 +140,12 @@ contract AmbireAccount {
 			require(privileges[signerKey] != bytes32(0), 'INSUFFICIENT_PRIVILEGE');
 		}
 
+		// we increment the nonce to prevent reentrancy
+		// also, we do it here as we want to reuse the previous nonce
+		// and respectively hash upon recovery / canceling
+		nonce = currentNonce + 1;
```
Ivshti, line 146: you can add a comment that doing this after sig verification is fine because sig verification can only do STATICCALLS

contracts/libs/SignatureValidatorV2.sol
Ivshti: sorry, that was a bug - I did think about it but I totally forgot to fix it; agreed with the change

Ivshti (changes requested):

contracts/AmbireAccountFactory.sol
borislav-itskov, line 1: @Ivshti Perhaps we no longer need `deployAndCall` as it targets the quick account manager that we will no longer use?

Ivshti (approved):

PR #229, opened 2023-05-04
Feature/deployless generate json

borislav-itskov:
Add a `compileFromContracts` method that takes a contract's name from the /contracts folder as an argument and generates at runtime:
* abi
* bytecode (also called bin)
* deployBytecode (also called binRuntime)


borislav-itskov:
@Ivshti I made the js implementation a typescript one

PR #244, opened 2023-05-17
Feature / useDapps `toggleFavorite` method optimization

superKalo:
Tweaks a bit the `toggleFavorite` method, which otherwise re-renders way too many times.


superKalo:
Released as v0.21.9. Please sync with latest from `develop` and merge later on.

PR #249, opened 2023-06-06
Feature / Keystore Signer

sonytooo:
* Added KeystoreSigner interface
* Add keystore.getSigner() function
* Add getSigner tests


v2/libs/keystore/keystore.ts
```diff
@@ -21,67 +20,80 @@ const SUPPORTED_KEY_TYPES = ['internal', 'trezor', 'ledger', 'lattice']
...
 
 type MainKey = {
-	key: Uint8Array;
-	iv: Uint8Array;
+	key: Uint8Array
+	iv: Uint8Array
 }
 
-type Key = {
+export type SupportedKeyTypes = 'internal' | 'trezor' | 'ledger' | 'lattice'
```
Ivshti, line 48: we shouldn't restrict those to a hardcoded list

v2/libs/keystore/keystore.ts
```diff
@@ -183,16 +225,52 @@ export class Keystore {
...
-		this.storage.set('keystoreKeys', keys.filter(x => x.id !== id))
+		if (!keys.find((x) => x.id === id)) throw new Error(`keystore: trying to remove key that does not exist ${id}}`)
+		this.storage.set(
+			'keystoreKeys',
+			keys.filter((x) => x.id !== id)
+		)
+	}
+
+	async getSigner(keyId: string) {
+		const storedKey: StoredKey = (await this.storage.get('keystoreKeys', [])).find((x: StoredKey) => x.id === keyId)
```
Ivshti, line 244: even though i gave you the example, maybe this should be a two-liner, the first line doing the `await`, so that we don't have `(await ..)...`

v2/libs/keystore/keystore.ts
```diff
@@ -183,16 +225,52 @@ export class Keystore {
...
+			meta,
+			isExternallyStored: type !== 'internal'
+		}
+
+		const signerInitializer = this.keystoreSigners[key.type]
+		if (!signerInitializer) throw new Error('keystore: unsupported singer')
+		if (key.type === 'internal' && !this.isUnlocked()) throw new Error('keystore: not unlocked')
+
+		if (key.type === 'internal') {
+			const encryptedBytes = getBytes(storedKey.privKey as string)
```
Ivshti, line 262: you can put `if (!this.unlocked()) throw new Error` in this branch instead of having a separate check, it will clean up the logic

v2/libs/keystore/keystore.ts
```diff
@@ -183,16 +225,52 @@ export class Keystore {
...
+		const key = {
+			id,
+			label,
+			type,
+			meta,
+			isExternallyStored: type !== 'internal'
+		}
+
+		const signerInitializer = this.keystoreSigners[key.type]
+		if (!signerInitializer) throw new Error('keystore: unsupported singer')
```
Ivshti, line 258: typo + should be `unsupported signer type`

Ivshti (changes requested):

Ivshti (approved):

PR #256, opened 2023-06-08
Feature / V2 UID and exportKeyWithPasscode in keyStore

stojnovsky:
#### added following tests:
- should export key backup, create wallet and compare public address
- should return uid
- should not return uid until there are no secrets yet

### DIFF
--- a/src/libs/keystore/keystore.ts
+++ b/src/libs/keystore/keystore.ts
@@ -271,6 +271,7 @@ export class Keystore {
     const storedKey: StoredKey = keys.find((x: StoredKey) => x.id === keyId)
 
     if (!storedKey) throw new Error('keystore: key not found')
+    if (storedKey.type !== 'internal') throw new Error('keystore: key does not have privateKey')
 
     const encryptedBytes = getBytes(storedKey.privKey as string)
     // @ts-ignore
--- a/src/libs/keystore/keystore.ts
+++ b/src/libs/keystore/keystore.ts
@@ -96,6 +96,12 @@ export class Keystore {
     return (await this.getMainKeyEncryptedWithSecrets()).length > 0
   }
 
+  async getKeyStoreUid() {
+    const uid = await this.storage.get('uid', null)
+    if (!uid) throw new Error('keystore: adding secret before get uid')
+    return uid
+  }
+
   // @TODO time before unlocking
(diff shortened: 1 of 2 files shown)


src/libs/keystore/keystore.ts
```diff
@@ -179,6 +185,12 @@ export class Keystore {
     })
     // Persist the new secrets
     await this.storage.set('keystoreSecrets', secrets)
+
+    // produce uid if not exist (should created when first secret added)
```
Ivshti, line 189: fix the english: "produce uid if one doesn't exist (should be created when the first secret is added)"

src/libs/keystore/keystore.ts
```diff
@@ -179,6 +185,12 @@ export class Keystore {
...
+    if (!(await this.storage.get('uid', null))) {
+      const uid = keccak256(randomBytes(32)).slice(2, 34)
```
Ivshti, line 191: why not just create this when the first secret is actually added (not unlocked) and why not have it be a hash of the mainKey

src/libs/keystore/keystore.ts
```diff
@@ -251,6 +264,27 @@ export class Keystore {
...
+    if (!this.isUnlocked()) throw new Error('keystore: not unlocked')
+
+    const keys = await this.storage.get('keystoreKeys', [])
+    const storedKey: StoredKey = keys.find((x: StoredKey) => x.id === keyId)
+
+    if (!storedKey) throw new Error('keystore: key not found')
+    if (storedKey.type !== 'internal') throw new Error('keystore: key does not have privateKey')
+
+    const encryptedBytes = getBytes(storedKey.privKey as string)
+    // @ts-ignore
```
Ivshti, line 277: what are we ignoring and why?

src/libs/keystore/keystore.ts
```diff
@@ -251,6 +264,27 @@ export class Keystore {
...
+    const counter = new aes.Counter(this.#mainKey.iv)
+    // @ts-ignore
```
Ivshti, line 279: again, what are we ignoring and why

Ivshti (changes requested):

Ivshti (approved):

PR #257, opened 2023-06-09
Feature / Portfolio Discovery fallback

jordan-enev:
- Add: Velcro Discovery fallback. Enrich portfolio hints with the previously found and cached hints, especially in the case Velcro discovery fails

### DIFF
--- a/src/controllers/portfolio.ts
+++ b/src/controllers/portfolio.ts
@@ -13,15 +13,21 @@ type AccountId = string
 // @TODO fix the any
 type PortfolioState = Map<AccountId, Map<NetworkId, any>>
 
-class PortfolioController {
+export class PortfolioController {
   latest: PortfolioState
   pending: PortfolioState
   private portfolioLibs: Map<string, Portfolio>
+  // @TODO - typescript interface
+  // It's Object, instead of Map, because Map can't be serialized in the storage
+  private previousHints: { [name: string]: any }
+  private storage: any
 
   constructor(storage: Storage) {
     this.latest = new Map()
     this.pending = new Map()
     this.portfolioLibs = new Map()
+    this.previousHints = {}
+    this.storage = storage
   }
   // NOTE: we always pass in all `accounts` and `networks` to ensure that the user of this
   // controller doesn't have to update this controller every time that those are updated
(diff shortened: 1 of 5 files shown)


src/libs/portfolio/portfolio.ts
```diff
@@ -88,7 +89,29 @@ export class Portfolio {
...
+        erc721s: {},
+        prices: {},
+        hasHints: false,
+        error
+      }
+    }
+
+    // Enrich hints with the previously found and cached hints, especially in the case the Velcro discovery fails.
+    if (opts.previousHints) {
+      hints = { ...hints, ...opts.previousHints }
```
Ivshti, line 112: this way of merging previous hints actually overrides, doesn't deep-merge; which is a problem in case we discover stuff through simulations (not implemented yet)

src/libs/portfolio/portfolio.ts
```diff
@@ -88,7 +89,29 @@ export class Portfolio {
...
-    const hints = await this.batchedVelcroDiscovery({ networkId, accountAddr, baseCurrency })
+
+    // Make sure portfolio lib still works, even in the case Velcro discovery fails.
+    // Because of this, we fall back to Velcro default response.
+    let hints
+    try {
+      hints = await this.batchedVelcroDiscovery({ networkId, accountAddr, baseCurrency })
+    } catch (error) {
+      hints = {
+        networkId: networkId,
```
Ivshti, line 100: we should move this cruft to the type definiitions and just do ` { ...EmptyHints, error }`

Ivshti (changes requested):
awesome work but it requires hints merging changes

src/controllers/portfolio.ts
```diff
@@ -13,15 +13,25 @@ type AccountId = string
 // @TODO fix the any
 type PortfolioState = Map<AccountId, Map<NetworkId, any>>
 
-class PortfolioController {
+interface Hints {
+  erc20s: string[]
+  erc721s: { [name: string]: { tokens: string[] } }
```
Ivshti, line 18: this is probably wrong, there's no `enumerable`

Ivshti (changes requested):

Ivshti (approved):

PR #528, opened 2024-01-19
Temporary / Email vault 

jordan-enev:
I'm adding these changes to the EmailVault feature in a temporary branch. Since the base branch (email-vault) includes URL changes to Relayer calls that have not been deployed to Relayer yet, it won't work.

Once the Relayer is deployed, we can merge `temporary/email-vault` into `email-vault`.

**Here are the changes:**
- Add: Keep keyStoreUid in the Controller state.
- Add: Enable KeyStore recovery banner.
- Fix: `emailVault.#recoverKeyStore` - it should recall `#recoverKeyStore`, when the magic link is confirmed.


src/controllers/emailVault/emailVault.ts
```diff
@@ -441,12 +443,39 @@ export class EmailVaultController extends EventEmitter {
...
+  get banners(): Banner[] {
     const banners: Banner[] = []
 
-    Object.keys(this.emailVaultStates?.email || {}).forEach((email) => {
+    if (!this.hasKeystoreRecovery) {
+      banners.push({
+        id: 'keystore-secret-backup',
+        topic: 'WARNING',
+        title: 'Enable local key store recovery with Email Vault',
+        text: "Email Vault recovers your Device Password. It is securely stored in Ambire's infrastructure cloud.",
```
JIOjosBG, line 468: I think the first part should be changed to 
"The Email Vault recovery allows you to change you device password."

JIOjosBG (dismissed):

src/controllers/emailVault/emailVault.ts
jordan-enev: I'll sync with Miro about the copy, as I get it from the design.

PR #541, opened 2024-01-23
Replace etherface with 4bytes api

JIOjosBG:
We used etherface as part of our fallback humanization module. Everytime we get unparsable txn with unknown sigHash we should query an api for the unknown sighash.

Etherface seems to have a problem so we now use 4bytes


borislav-itskov:
Great, let's merge

PR #543, opened 2024-01-24
Feature: show basic acc simulation

borislav-itskov:
Add: show the basic account simulation


Ivshti (changes requested):
some changes requested on slack

PR #551, opened 2024-01-26
Feature / Portfolio controller critical errors as banners

superKalo:
Handles:

- Add: Display the portfolio critical errors as banners for the networks on which the currently selected account has assets
- Fix: The Portfolio controller `criticalError`s were getting set as an object with error attr (not an Error instance), so normalize them in order to fit the Error instance type (with a message)

Mentioned as missing in https://github.com/AmbireTech/ambire-app/issues/1567


Ivshti (approved):

src/libs/banners/banners.ts
```diff
@@ -177,3 +178,44 @@ export const getNetworksWithFailedRPCBanners = ({
...
+
+      const networkData = networks.find((n: NetworkDescriptor) => n.id === network)
+      if (!networkData) return []
+
+      return [
+        {
+          id: `${networkData.id}-${new Date().getTime()}`,
+          topic: 'WARNING',
+          title: `Failed to retrieve the portfolio data for ${networkData.name}`,
+          text: `Affected features(${
```
Ivshti, line 212: no space before the (

src/libs/banners/banners.ts
```diff
@@ -177,3 +178,44 @@ export const getNetworksWithFailedRPCBanners = ({
...
+  return Object.keys(selectedAccPortfolio)
+    .filter((networkId) => portfolio.networksWithAssets.includes(networkId))
+    .flatMap((network) => {
+      const portfolioForNetwork = selectedAccPortfolio[network]
+
+      const criticalError = portfolioForNetwork?.criticalError
+      if (!criticalError) return []
+
+      const networkData = networks.find((n: NetworkDescriptor) => n.id === network)
+      if (!networkData) return []
```
Ivshti, line 205: shouldn't we return an issue in this case

Ivshti (changes requested):

src/libs/banners/banners.ts
superKalo: Ah, I made this on purpose, to be consistent with the RPC error. I changed it in both places.

Screenshot 2024-01-29 at 19 35 26

and

Screenshot 2024-01-29 at 21 56 09

What do you think, should we try to shorten them even more?

src/libs/banners/banners.ts
superKalo: yes, good catch, error log added ✅

PR #554, opened 2024-01-29
Feature: better simulation and errors

borislav-itskov:
Fixes:
- Estimation.sol was reverting when associatedKeys were passed but none of them could generate a valid spoof. Fixed, now they return a SPOOF_ERROR that translates in the UI as "Your signer address is not authorized"
- ERC-4337 estimation was broken as an old version of accountOp that did not contain the userOperation properties was passed down to SignAccountOp.ts. Fixed
- Humanizer: ERC-20 approvals were regarded as NFT approvals and were not humanized to tokens. Fixed

Error handling:
- errors starting with an error signature of 0x08c379a0 get translated to string messages and displayed in the UI if there's an estimation failure
- added 0x5bf6f916 which translates to "Swap expired". This error will be shown in the simulation and estimation if an uniswap swap has expired

Humanizer:
- latest entry point humanization
- uniswap universal router humanization for the avalanche chain


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -287,6 +288,8 @@ export class SignAccountOpController extends EventEmitter {
 
     if (estimation) this.#estimation = estimation
 
+    if (accountOp) this.accountOp = accountOp
```
Ivshti, line 291: was this discussed with Dancho? makes sense but do we have internal mutability of accountOp in signAccountOp? we probably do when setting fees and etc. 

so this sounds like a potential bug

src/libs/deployless/deployless.ts
```diff
@@ -19,6 +19,8 @@ const deployErrorSig = '0xb4f54111'
 const errorSig = '0x08c379a0'
 // Signature of Panic(uint256)
 const panicSig = '0x4e487b71'
+// uniswap swap expired
+const expiredSig = '0x5bf6f916'
```
Ivshti, line 23: this error should NOT be parsed here, it's way too specific

Or we should make another library for error parsing that deployless uses

borislav-itskov: I haven't forgotten about this, will fix

PR #556, opened 2024-01-31
Fix/hide deadline option in uniswap humanizer

borislav-itskov:
In benzina, swap deadline started showing up for swaps from uniswap which is not correct as the transaction has already finished.  
This hides the deadline from the uniswap module if that is passed in options


JIOjosBG:
This is not the way to fix the benzina deadline issue. Additional option arguments and repetetive checks should be avaoided. The desired solution would be to have option in the front end component that is responsible for showing the humanization. In the component we cycle trough the items and their types and display each accordingly. 
When we reach 
`if(i.type==='deadline **&& options.displayDeadline)**` we should display the deadline

JIOjosBG:
Updates should be here
https://github.com/AmbireTech/ambire-app/pull/1580/files

borislav-itskov:
:+1: 

Logic moved to FE

PR #558, opened 2024-01-31
Change/ Scope pinned tokens to a network

PetromirDev:
## Changes:

- Changed pinnedTokens data structure and type (added type in interfaces too)
- Added the ability for pinned gas tank tokens

### Linked with:
https://github.com/AmbireTech/ambire-app/pull/1654

### DIFF
--- a/src/controllers/portfolio/portfolio.ts
+++ b/src/controllers/portfolio/portfolio.ts
@@ -19,8 +19,10 @@ import {
   AdditionalAccountState,
   GetOptions,
   Hints,
+  PinnedTokens,
   PortfolioControllerState,
-  PortfolioGetResult
+  PortfolioGetResult,
+  TokenResult
 } from '../../libs/portfolio/interfaces'
 import { Portfolio } from '../../libs/portfolio/portfolio'
 import { relayerCall } from '../../libs/relayerCall/relayerCall'
@@ -60,7 +62,7 @@ export class PortfolioController extends EventEmitter {
 
   #callRelayer: Function
 
-  #pinned: string[]
+  #pinned: PinnedTokens
 
   #networksWithAssetsByAccounts: {
     [accountId: string]: NetworkDescriptor['id'][]
@@ -68,7 +70,7 @@ export class PortfolioController extends EventEmitter {
 
   #minUpdateInterval: number = 20000 // 20 seconds
 
-  constructor(storage: Storage, providers: RPCProviders, relayerUrl: string, pinned: string[]) {
+  constructor(storage: Storage, providers: RPCProviders, relayerUrl: string, pinned: PinnedTokens) {
     super()
     this.latest = {}
     this.pending = {}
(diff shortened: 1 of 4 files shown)


borislav-itskov (changes requested):

src/libs/portfolio/portfolio.ts
```diff
@@ -163,7 +163,13 @@ export class Portfolio {
     }
 
     const tokenFilter = ([error, result]: [string, TokenResult]): boolean =>
-      (result.amount > 0 || pinned!.includes(result.address)) &&
+      (result.amount > 0 ||
+        !!pinned.find((pinnedToken) => {
+          // Pinned tokens from the humanizer don't have a networkId
+          if (pinnedToken.networkId === null) return pinnedToken.address === result.address
```
Ivshti, line 169: we need to change the humanizer perhaps because this could be a security risk in very weird circumstances/edge cases

src/controllers/portfolio/portfolio.ts
```diff
@@ -185,13 +187,64 @@ export class PortfolioController extends EventEmitter {
...
     }))
+
+    let pinnedGasTankTokens: TokenResult[] = []
+
+    if (res.data.gasTank.availableGasTankAssets) {
+      const availableGasTankAssets = res.data.gasTank.availableGasTankAssets
+
+      pinnedGasTankTokens = availableGasTankAssets.reduce((acc: TokenResult[], token: any) => {
+        const isGasTankToken = !!gasTankTokens.find(
+          (gasTankToken: TokenResult) =>
```
Ivshti, line 198: shouldn't we make this a closure, since we're using it a lot in this function

Ivshti (changes requested):

src/libs/portfolio/portfolio.ts
PetromirDev: @JIOjosBG @borislav-itskov

src/controllers/portfolio/portfolio.ts
```diff
@@ -185,13 +187,64 @@ export class PortfolioController extends EventEmitter {
...
+
     accountState.gasTank = {
       isReady: true,
       isLoading: false,
       errors: [],
       result: {
         updateStarted: start,
-        tokens: gasTankTokens,
+        tokens: [...gasTankTokens, ...pinnedGasTankTokens],
         total: getTotal(gasTankTokens)
```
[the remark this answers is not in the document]
PetromirDev: I don't think that's necessary since the amount of a pinned gas tank token is always `0n`

src/controllers/portfolio/portfolio.ts
PetromirDev: What should we make a closure exactly and what are the pros of making the change?

src/libs/portfolio/portfolio.ts
JIOjosBG: Here is a proposal for the future structure for humanizerMeta. 
My initial idea is to store whether it is pinned in the humanizerMeta, but there might be better and simpler solution
https://github.com/AmbireTech/ambire-app/issues/1662

PR #560, opened 2024-02-01
Suggestion: last block refetch on gas price

borislav-itskov:
Try refetching the block 5 times before declaring failure during gas price fetching


borislav-itskov:
Merging as tests are failing because of RPC failure.  
Also, I couldn't find a way to tell ethers to just "refetch" the request a couple of times instead of this implementation

PR #571, opened 2024-02-06
Feature / Email Vault - new state for tracking secret upload

jordan-enev:
Add: Flag (#isUploadingSecret) for tracking the upload state of keystore secret.

### DIFF
--- a/src/controllers/emailVault/emailVault.ts
+++ b/src/controllers/emailVault/emailVault.ts
@@ -22,6 +22,7 @@ import { KeystoreController } from '../keystore/keystore'
 export enum EmailVaultState {
   Loading,
   WaitingEmailConfirmation,
+  UploadingSecret,
   Ready
 }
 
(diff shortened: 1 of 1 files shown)


JIOjosBG:
We discussed that a better approach might be to check if we have new secret with type for keystore recovery and id = the device's keystoer uid added to the email vault controller state. This concern that lead to implementation in this PR was that we might not have actually recorded the secret in the relayer, but we might have it in the controller state. This is not realistically the case, because the state of the secrets is determined by the success messages and emailVaultInfo route in the relayer.

JIOjosBG (commented):
We discussed that a better approach might be to check if we have new secret with type for keystore recovery and id = the device's keystoer uid added to the email vault controller state. This concern that lead to implementation in this PR was that we might not have actually recorded the secret in the relayer, but we might have it in the controller state. This is not realistically the case, because the state of the secrets is determined by the success messages and emailVaultInfo route in the relayer.

jordan-enev:
We agreed that we need to have a flag that tracks if we are uploading a secret in the moment, in order to show a Loader component.

PR #575, opened 2024-02-07
WIP: Feature: paymaster sign unprotected

borislav-itskov:
Change log:
* allow the paymaster to sign unprotected

Main benefits for this are gas cost improvements:
* instead of `recoverAddrImpl`, we call directly `recoverAddrAllowUnprotected`
* instead of going through the EIP-712 setter logic, we directly pass an ethereum signed message hash for signature validation

For discussion:
* we have `bool isValidSig = sigAddr == relayer;` in the contract which can become a one liner instead. I've left it like this as it's easier to understand what's happening


borislav-itskov:
TODO: remove the eth prefix and sing with a library that supports clear signing

Ivshti:
We are dropping this because we can just use a different library to sign without the prefix

PR #578, opened 2024-02-08
WIP: Enhancement/one place for contract error messages

borislav-itskov:
Fix:
* deployless.test.ts eslint issues
* add errors from estimation to deployless and call `parseErr` to decode them
* if one of the inner reestimation requests fails, it returns an error object to `reestimate`. If `reestimate` fails two times with an error, it returns this error message to the user
* slow request reestimates are moved from 15 to 8 seconds


Ivshti (changes requested):
discussed on slack

PR #583, opened 2024-02-09
Fix: EOA simulation

borislav-itskov:
Change log:
* override the nonce for EOA during simulation;
* throw an error if the afterNonce is equal to or less to the beforeNonce;
* throw an error if the afterNonce is smaller or equal to the last account op nonce, if any set

Noticeable changes:
* instead of always setting the nonce to 0 for EOA during simulation, we set it to `0x1000000000000000000000000000000000000000000000000000000000000000`. This is because some RPCs complain when we try to do a state override with a number that has leading zeros.
* we throw an error if `afterNonce < simulatedAccountOpLastNonce + 1` for EOAs as well as the "fake" EOA nonce should have moved as well after simulation

### DIFF
--- a/src/libs/deployless/deployless.ts
+++ b/src/libs/deployless/deployless.ts
@@ -49,12 +49,17 @@ const defaultOptions: CallOptions = {
 
 export class Deployless {
   private iface: Interface
+
   // the contract deploy (constructor) code: this is the code that tjhe solidity compiler outputs
   private contractBytecode: string
+
   private provider: JsonRpcProvider | Provider
+
   // We need to detect whether the provider supports state override
   private detectionPromise?: Promise<void>
+
   private stateOverrideSupported?: boolean
+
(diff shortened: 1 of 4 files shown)


Ivshti (changes requested):
check slack
1) actually the `<=` check should be reverted
2) we should introduce a test that catches the bug we just introduced by changing `<` to `<=` (I know it was my idea but I didn't think of the pending w/o accountOps edge case)
3) the case we wanted to catch, which is accountOps passed with wrong nonce and not simulated, will be caught by the `lastAccOpNonce` case
4) **should we introduce some rules/checks/etc.** for whether the accountOps need to be passed in nonce order? cause the contract expects this. We can sort it in `getOnchainBalances` but if that's the case then we need to use `Math.max` rather than getting the last accountOp nonce

Ivshti (approved):

PR #585, opened 2024-02-12
Fix: Account related pinned tokens

borislav-itskov:
Until now, the pinned tokens that were added to the portfolio controller did not have an account relationship.  
This meant that once a pinned token got saved in the portfolio controller, that token was fetched for each account.  

This PR separates the tokens in two:
* without accountId, belonging to each account (the preset)
* with accountId, added on simulation

There's a test for this in the PR - once we add the pinned token to an account, it does not get fetched for another.  
Also, a discussion point that @PetromirDev found. Once a pinned token is added, it goes to hints.erc20 and gets cached. So even if the balance of the token is zero (the user makes a simulation but does not conclude his txn), it will continue to be fetched for this account in the future. This is also proven in the test in this PR.  
We should probably think how to handle this.

### DIFF
--- a/src/controllers/portfolio/portfolio.test.ts
+++ b/src/controllers/portfolio/portfolio.test.ts
@@ -28,6 +28,13 @@ describe('Portfolio Controller ', () => {
     }
   }
 
+  const EOA = {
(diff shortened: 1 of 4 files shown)


PetromirDev (approved):
Good job :fire:

PR #594, opened 2024-02-15
Bug/ Fee speeds error is displayed for a split second if fee speeds take longer do show

PetromirDev:
How to test:
- Go to the network settings of the background process
- Set throttling to Regular 3G
- Try to sign a transaction

A loading indicator should be visualized and no errors should be displayed until the indicator is replaced with the available fee speeds.


superKalo (approved):
Tested all cases that I experienced before, all good!

PR #596, opened 2024-02-15
Bug/ SignAccountOp can't estimate when the user doesn't have the native token in his portfolio

PetromirDev:
## Changes:

- Added the native token to pinned if it's missing in portfolio

## Before:
image
## After:
image


borislav-itskov (approved):

Ivshti:
lgtm

PR #597, opened 2024-02-15
Add expiry to magic link in email vault

JIOjosBG:


src/controllers/emailVault/emailVault.ts
```diff
@@ -228,7 +226,7 @@ export class EmailVaultController extends EventEmitter {
   getMagicLinkKeyByEmail(email: string): MagicLinkKey | null {
     const result = this.#magicLinkKeys[email]
     if (!result || !result.confirmed) return null
-    if (new Date().getTime() - result.requestedAt.getTime() > this.#magicLinkLifeTime) return null
+    if (Date.now() > result.expiry.getTime()) return null
```
jordan-enev, line 229: should it be `>=`?

JIOjosBG: technically yes, but it doesn't make much of a difference since its in milliseconds. Will update it anyway

PR #598, opened 2024-02-15
Feature / Rich JSON

sonytooo:
* rename and replace bigintJson with richJson
* add proper handling of the Error object when stringifying or parsing


src/libs/richJson/richJson.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+export function stringify(obj: any): string {
+  return JSON.stringify(obj, (key, value) => {
+    if (typeof value === 'bigint') {
+      return { $bigint: value.toString() }
+    }
+
+    if (value instanceof Error) {
+      const error: any = {}
```
Ivshti, line 24: won't this work with spread, Object.assign or something

Ivshti (changes requested):

src/libs/richJson/richJson.ts
sonytooo: tried spreading and Object.assign but it doesn't work and after stringifying the result is an empty object.

only 
`    
  const error: any = {}
  Object.getOwnPropertyNames(value).forEach((propName) => {
     error[propName] = value[propName]
  })
  return error
`
and
`
  return Object.getOwnPropertyNames(value).reduce((error: any, propName: string) => {
    error[propName] = value[propName];
    return error;
  }, {});
`
seem to do the trick

PR #599, opened 2024-02-15
Fix/display relayer error

borislav-itskov:
Change:
* display the bug from the relayer, if any
* make the fee tokens list up-to-date


Ivshti:
lgtm

PR #615, opened 2024-02-20
Refactoring / Keystore Controller - accessing and persisting internal controller state

jordan-enev:
:warning: At first glance, it doesn't seem like we'll break anything because the changes are insignificant, but we should test it carefully. Especially in cases of extension updates (i.e., when we already have persistent items in the storage) or when adding new accounts through different flows.

**Here's what I've done in the PR:**

1. Refactor `keystoreSecrets, keystoreKeys, keyStoreUid` class properties to be accessed and persisted in the storage by the convention we use in other controllers. I.e. everytime we need to retrieve a class property - we should always access it from the controller state (i.e. this.#keystoreSecrets), instead of the storage. In the case we want to update the property - first we are updating the property itself and after that we are persisting it in the storage.
2. Togerher with that we have to introduce `initialLoadPromise` (again it's a convention). It holds the initial hydratation of the storage <=> controller's state.
3. Add a Keystore getter for checking if the keystore has a password secret set.
4. Reuse the above flag for the EV banner visibility logic.


src/controllers/keystore/keystore.ts
```diff
@@ -621,11 +607,17 @@ export class KeystoreController extends EventEmitter {
     this.emitUpdate()
   }
 
+  get hasPasswordSecret() {
+    return !!this.#keystoreSecrets.find((x) => x.id === 'password')
```
superKalo, line 611: Minor: you can use the `some` method instead of `find` to return a boolean directly without needing to use `!!`, like so:

```
return this.#keystoreSecrets.some((x) => x.id === 'password');
```

superKalo (approved):
Great refactoring for the internal state! 👏 

I tested by connecting a Trezor, Ledger, then setting up Device Password and importing a seed. All good 👍

PR #620, opened 2024-02-21
Fix / [object Object] Error in Protfolio

sonytooo:
Fixed: [object Object] error in the portfolio ctrl when logged in dev mode in background.ts


superKalo (approved):
Since it is a silent error anyway (mostly for dev purposes and debugging) and since the user doesn't see it, I think this is fine.

But is the underlying issue (`[object Object]`) resolved in the `error: e` part?

PR #635, opened 2024-02-28
Feature: Paymaster for native & Optimism fee optimisation

borislav-itskov:
Changes:
* pay ERC-4337 native fees wtih a paymaster
* calculate the L1 Fee data with "payment in native" and "transfer payment" as there could be 10 cent difference depending on what option we choose
* calculate `maxPriorityFeePerGas` for optimism as a small percentage of the base fee as transaction included in a block have too random `maxPriorityFeePerGas` and result differ greatly
* remove the extra optimism fee as it becomes too expensive with it


Ivshti:
lgtm

PR #637, opened 2024-02-29
Feature / Token Preferences

gergana95:
Changelog:

- Pass additionalHints to updateSelectedAccount in order to discover them
- Store tokenPreferences in portfolio controller and in storage
- Store validTokens in portfolio controller and use them to check for erc20 validity
- Pass in tokenPreferences to portfolio lib and return them as well


src/controllers/portfolio/portfolio.ts
```diff
@@ -158,6 +189,21 @@ export class PortfolioController extends EventEmitter {
     this.#additionalHints = []
   }
 
+  async updateTokenValidationByStandard(token: any, accountId: AccountId) {
```
PetromirDev, line 192: replace any

src/libs/portfolio/helpers.ts
```diff
@@ -37,6 +41,44 @@ export function getFlags(
   }
 }
 
+export const validateERC20Token = async (
+  token: { address: string; networkId: NetworkId },
+  accountId: string,
+  provider: RPCProvider
+) => {
+  const erc20 = new Contract(token?.address, IERC20.abi, provider)
```
PetromirDev, line 49: token is required `?` is redundant

src/libs/portfolio/helpers.ts
```diff
@@ -37,6 +41,44 @@ export function getFlags(
...
+
+  const type: string = 'erc20'
+  let isValid: boolean = true
```
PetromirDev, line 52: Typescript assigns these types automatically

PetromirDev (changes requested):

PetromirDev:
Also, it'd be good to write additional portfolio tests for this feature

PR #640, opened 2024-03-01
WIP: Feature/gas tank 4337

borislav-itskov:


borislav-itskov:
Closing this as there's close to no point for this optimization for now - fees are pretty low on all networks, even Ethereum is great compared to the past. When the time comes to focus on gas price optimization, we can look at it again

PR #645, opened 2024-03-05
gas tank call: ignore calls to address(0)

Ivshti:
## Summary
Ignores all calls to address(0), allowing to use them as "notes" or "inscriptions"


borislav-itskov (commented):
Giga top

PR #649, opened 2024-03-13
Enhancement/ Manage account keys from the UI

PetromirDev:
## Linked with:
https://github.com/AmbireTech/ambire-app/pull/1901


superKalo (approved):

superKalo:
I think that was left as a legacy from the time when the keys had only addr, but not a type associated. Good catch 👍

PR #650, opened 2024-03-13
Fix errors for EV shouldStopConfirmationPolling checks

JIOjosBG:
There was this problem on cancelation for email confirmation
image

### DIFF
--- a/src/controllers/emailVault/emailVault.ts
+++ b/src/controllers/emailVault/emailVault.ts
@@ -307,17 +307,18 @@ export class EmailVaultController extends EventEmitter {
     let result: Boolean | null = false
     let magicKey = await this.#getMagicLinkKey(email)
 
-    if (!this.#shouldStopConfirmationPolling) {
+    if (!magicKey?.key && !this.#shouldStopConfirmationPolling) {
+      await this.handleMagicLinkKey(email, async () => {
+        magicKey = await this.#getMagicLinkKey(email)
+      })
+    }
+
+    if (this.#shouldStopConfirmationPolling) {
       this.#isUploadingSecret = false
       this.emitUpdate()
       return
     }
 
-    if (!magicKey?.key) {
-      await this.handleMagicLinkKey(email, async () => {
-        magicKey = await this.#getMagicLinkKey(email)
-      })
-    }
     if (magicKey?.key) {
       this.#isUploadingSecret = true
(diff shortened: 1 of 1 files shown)


src/controllers/emailVault/emailVault.ts
```diff
@@ -306,11 +306,19 @@ export class EmailVaultController extends EventEmitter {
...
       await this.handleMagicLinkKey(email, async () => {
         magicKey = await this.#getMagicLinkKey(email)
       })
     }
+
+    if (this.#shouldStopConfirmationPolling) {
+      this.#isUploadingSecret = false
+      this.emitUpdate()
+      return
+    }
```
jordan-enev, line 320: Why don't we put this short-circuit at the very beginning of async handleMagicLinkKey and stop caring for it in the rest of the function?

src/controllers/emailVault/emailVault.ts
```diff
@@ -214,7 +214,7 @@ export class EmailVaultController extends EventEmitter {
       fn && (await fn())
       this.storage.set(MAGIC_LINK_STORAGE_KEY, this.#magicLinkKeys)
       this.#requestSessionKey(email)
-    } else if (this.#shouldStopConfirmationPolling)
+    } else if (!this.#shouldStopConfirmationPolling)
```
jordan-enev, line 217: here we can do it again with short-circuit before ` if (ev && !ev.error) {`, because it's a bit misleading in the else if

jordan-enev:
one more thing, in `this.#wrapEmailVaultPublicMethod` we should not update the status to DONE and emit an update, when we stop the polling. Otherwise, when we click 'cancel', a success modal would be shown (I'm currently working on this)

src/controllers/emailVault/emailVault.ts
JIOjosBG: Because we want to exit after the polling has been cancelled or it has failed. There is no realistic case where the cancel button has been pressed before the polling has started. The point of this if is to return right before emittingErrors

JIOjosBG:
> one more thing, in `this.#wrapEmailVaultPublicMethod` we should not update the status to DONE and emit an update, when the polling is stopped. Otherwise, when we click 'cancel', a success modal would be shown (I'm currently working on this)

Not sure is DONE should be accepted as READY or as SUCCESS

jordan-enev (approved):

PR #656, opened 2024-03-18
Enhancement/ ENS/UD reverse lookup

PetromirDev:
## Changes:
- Add: reverse lookup helpers (both ENS and UD)
- Update: `@unstoppabledomains/resolution` to ^9.2.2, because reverse lookup wasn't possible in the previous version
- Add: domains controller and test
## Linked with:
https://github.com/AmbireTech/ambire-app/pull/1919

### DIFF
--- a/src/controllers/domains/domains.ts
+++ b/src/controllers/domains/domains.ts
@@ -1,0 +1,94 @@
+import { getAddress } from 'ethers'
+
+import { RPCProviders } from '../../interfaces/settings'
+import { reverseLookupEns } from '../../services/ensDomains'
+import { reverseLookupUD } from '../../services/unstoppableDomains'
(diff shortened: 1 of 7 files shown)


src/controllers/domains/domains.ts
```diff
@@ -0,0 +1,107 @@
...
+
+interface Domains {
+  [address: string]: {
+    ens: string | null
+    ud: string | null
+    savedAt: number
+  }
+}
+
+const DAY_IN_MS = 24 * 60 * 60 * 1000
```
superKalo, line 16: Note: To be discussed.

PetromirDev: Changed to 15 minutes

PR #662, opened 2024-03-21
Fix: not all gas tank tokens found

JIOjosBG:
The copy-pasted token list for gastank assets was from the relayer route that does not return all addresses as checksummed. issue here: 
- https://github.com/AmbireTech/relayer/issues/659

Also updated the getFlags function

Works with this:
 - https://github.com/AmbireTech/ambire-app/pull/1933

This PR closes: 
- https://github.com/AmbireTech/ambire-app/pull/1927


gergana95 (approved):
Looks good

PR #677, opened 2024-03-28
Fix: state overrides so the bundler estimates correctly

borislav-itskov:
fix: state overrides so the bundler estimates correctly


Ivshti:
lgtm

PR #683, opened 2024-04-01
Fix: nonce discrepancy issues when there's a pending txn

borislav-itskov:
Change log:
* rename `nonce` to `currentAccountNonce` in `estimate.ts` for more readability. `EstimateResult` should always return the current nonce, or the one the user will commit to in case of a `execute()` authrorization
* if there's a discrepancy in the accountOp.nonce and the (outcome.nonce - 1), we 1) return an error in `signAccountOp` and do not allow the user to sign 2) set the found nonce from the estimation in `accountOp` as the true nonce and wait for simulation & estimation reruns


Ivshti (approved):

Ivshti:
lgtm

PR #690, opened 2024-04-04
Feature / Add Temporary Tokens

gergana95:
In order to fetch data for tokenPreferences before adding it to the list I create this temporaryTokens state which will be used to store them.

This fixes adding tokens in additionalHints and bug found when being discovered with balance.


src/controllers/portfolio/portfolio.ts
```diff
@@ -85,6 +86,14 @@ export class PortfolioController extends EventEmitter {
 
   validTokens: any = { erc20: {}, erc721: {} }
 
+  temporaryTokens: {
+    [networkId: NetworkDescriptor['id']]: {
+      isLoading: boolean
+      errors: ExtendedError[]
```
borislav-itskov, line 92: Isn't this error collection heavy for live usage? Are we planning to disable this?  
Also, the ExtendedError is not added as a package in ambire-common. If we plan to use it, we'll need to add it

src/controllers/portfolio/portfolio.ts
```diff
@@ -207,6 +217,67 @@ export class PortfolioController extends EventEmitter {
...
+    // 2. The network RPC URL has changed
+    if (
+      !this.#portfolioLibs.has(key) ||
+      this.#portfolioLibs.get(key)?.network?.rpcUrl !==
+        // eslint-disable-next-line no-underscore-dangle
+        providers[network.id]?._getConnection().url
+    ) {
+      this.#portfolioLibs.set(key, new Portfolio(fetch, providers[network.id], network))
+    }
+    const portfolioLib = this.#portfolioLibs.get(key)!
```
borislav-itskov, line 237: We can move the code from line 224:237 to a separate function and reuse it here and in `updateSelectedAccount` as it's the same piece of code (if I'm not mistaken)

borislav-itskov (changes requested):
We have two small issues to discuss, everything else lgtm

borislav-itskov (approved):
:fire: :fire: :fire:

PR #693, opened 2024-04-04
Feature/eoa estimation.sol

borislav-itskov:
Make two estimations for EOA:
* one from Estimation.sol using a state override that makes the EOA a smart account
* one from eth_estimateGas

We compare both estimations and do the following:
* take the native balance the user has from Estimation.sol. The balance from there will be the balance after the pending txns have been simulated. So if the user tries to send all his native, his balance for the fee should be 0. We cover this case
* if the EOA is trying to do a normal transfer of native, trust eth_estimateGas as it should return 21K in most cases
* if the EOA is trying to do a contract call, compare the gas of eth_estimateGas and of Estimation.sol and use the higher one

Cases where the above shouldn't work as expected:
* if the simulation from Estimation.sol does not complete successfully, the native returned from Estimation.sol will be the current user balance instead the one after a successful simulation


src/libs/estimate/reestimate.ts
```diff
@@ -0,0 +1,31 @@
+export async function reestimate(fetchRequests: Function, counter: number = 0): Promise<any> {
```
Ivshti, line 1: should be renamed, this is actually `estimateWithRetries`

src/libs/estimate/reestimate.ts
```diff
@@ -0,0 +1,31 @@
...
+  // stop the execution on 5 fails;
+  // the below error message is not shown to the user so we are safe
+  if (counter >= 5)
+    return new Error(
+      'Estimation failure, retrying in a couple of seconds. If this issue persists, please change your RPC provider or contact Ambire support'
+    )
+
+  const estimationTimeout = new Promise((resolve) => {
+    setTimeout(() => {
+      resolve('Timeout reached')
```
Ivshti, line 11: you can do this more elegantly without having to check for typeof

the way it's done is
```
// allocate an object that will be resulted from the promise that can then be compared by reference
const santinelTimeoutErr = {}

...
resolve(santinelTimeoutErr)
....

if (result === santinelTimeoutErr)

```

src/libs/estimate/reestimate.ts
```diff
@@ -0,0 +1,31 @@
...
+  // if the request reaches the timeout, it cancels it and retries
+  let result = await Promise.race([Promise.all(fetchRequests()), estimationTimeout])
+
+  if (typeof result === 'string') {
+    const incremented = counter + 1
+    result = await reestimate(fetchRequests, incremented)
+  }
+
+  // if one of the calls returns an error, return it
+  if (Array.isArray(result)) {
```
Ivshti, line 25: then if my suggestion is done this `Array.isArray` shouldn't be needed, it will just be `if (result === santinelTimeoutErr) ... else {`

PR #702, opened 2024-04-11
Refactor/fee speeds

borislav-itskov:
1. `feeSpeeds` get changed and from an object that represented an array of slow, medium, fast and ape for the chosen `availableFeeOption` (fee token + paidBy), they now represent a collection of all the speeds for all available `availableFeeOptions`

As a result:
* we do not need to recalculate the `feeSpeeds` when the user chooses another payment option
* we can easily determine on the FE which option has enough funds to cover the fee and disable it if it doesn't; and we can easily determine whether the user has enough funds for the chosen speed as well (maybe he does for medium but not for fast, in which case we should disable fast and ape)

2. `getFeeSpeedIdentifier` was introduced as an ID for each `feeSpeed`. It represents a combination of the fee token + who is paying the fee + is it gas tank.

Also, a unique identifier was placed for EOAs paying for SA. EOAs can only pay in native and it doesn't matter which EOA does the final broadcast, the `feeSpeed` calculation for it will be the same. So we calculate it once and the next time we encounter an EOA, we just skip the calculation as it's already done.  
Example: user has 5 EOAs => 5 `availableFeeOptions`. Calculation will be done for the first option. For the next 4, it will see that  this calculation already done and skip on ahead, improving performance for users with a lot of EOAs

3. `selectedOption` was introduced as a property in `signAccountOp.ts`. Basically, it is as it's name suggests: it's the `availableFeeOption` the user has chosen. At controller initialization, it is `undefined`. It gets set on `update` once `estimation`, `paidBy` and `feeTokenResult` are available.  
This is just a little extra as `selectedOption` could be calculated at any given time having the required properties. But caching it remove one layer of copy/pasting the same find function across the controller

4. `isInitialized` - `isInitialized` checked for 4 properties before declaring true: account, accountOp, network and estimation. Of those 4, the first 3 are set in the constructor. So there is a non-existent chance of them not being set. We leave only the estimation there and set isInitialized to true once we have the estimation

5. `feeSpeedsLoading` - we remove `paidBy` and `feeTokenResult` as required properties for this as `feeSpeeds` now calculate all the speeds for all `availableFeeOptions`. They do not depend on `paidBy` and `feeTokenResult`


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -489,125 +505,125 @@ export class SignAccountOpController extends EventEmitter {
...
-          simulatedGasLimit,
-          amount,
-          amountFormatted: formatUnits(amount, Number(this.feeTokenResult!.decimals)),
-          amountUsd: getTokenUsdAmount(this.feeTokenResult!, amount) ?? '',
-          gasPrice,
-          maxPriorityFeePerGas: BigInt(speedValue.maxPriorityFeePerGas)
-        })
+      const nativeRatio = this.#getNativeToFeeTokenRatio(option.token)
+      if (!nativeRatio) {
+        this.feeSpeeds[identifier] = []
```
jordan-enev, line 536: why [], instead of simply a short-circuit with `return`

jordan-enev (changes requested):

PR #709, opened 2024-04-17
Fix: native fee payment option for custom network first SA txn

borislav-itskov:
Fix: native fee payment option was filtered for custom networks's non-erc-4337 requests (the first smart account txn)


Ivshti (commented):
lgtm

PR #710, opened 2024-04-17
Bug/ Portfolio lib swallows `priceIn` errors

PetromirDev:
## Changes:
- Added `errors` array to the return value of the portfolioLib
- Deleted `hintsError`, which is now stored in `errors`
- Separated portfolio error banners into two:
  - critical error banners- the portfolio doesn't work at all
  - missing prices error banners- tokens are displayed without prices
- Removed param reassigns of `priceIn`

## How to test:
- Block requests to `cena.ambire.com`
- Block requests to `velcro`
- Block requests to `invictus`

Combine the failure of different services or test them separately. The extension should work smoothly and display feedback to the user if a service fails.

### DIFF
--- a/src/libs/portfolio/portfolio.ts
+++ b/src/libs/portfolio/portfolio.ts
@@ -247,8 +247,6 @@ export class Portfolio {
 
     const oracleCallDone = Date.now()
 
-    const tokensWithPriceErrors: PortfolioGetResult['tokenErrors'] = []
-
     // Update prices and set the priceIn for each token by reference,
     // updating the final tokens array as a result
     const tokensWithPrices = await Promise.all(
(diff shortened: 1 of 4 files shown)


Ivshti (changes requested):
we still don't preserve the error from the catch anywhere

src/libs/portfolio/portfolio.ts
```diff
@@ -264,16 +273,40 @@ export class Portfolio {
...
+            })
+          }
+        }
+
+        if (!priceIn.length) {
+          tokensWithPriceErrors.push({
+            error: `Failed to fetch price data from cena.ambire.com for ${token.symbol} on ${networkId}`,
+            address: token.address
+          })
+        }
```
PetromirDev, line 305: This may be redundant, because the only scenario in which we store such error is when the token doesn't exist in coingecko, which would be obvious if `priceIn` is empty and there is no error for the whole network.

Ivshti (approved):

PR #718, opened 2024-04-19
Bug/ Portfolio controller state prepare racing condition

PetromirDev:
## The bug:
We determine if the portfolio is loading based on the flags `network.isReady`, `network.isLoading` and `network.criticalError` of every network. The problem is that this state is missing in the beginning so the flag `isAllReady`'s value changes like this: false, false, false, true, `(gasTank and rewards keys are added here) ---->` false, `(gasTank and rewards load here --->` true, true

## The fix:
The fix is obvious- prepare the state for all networks in the beginning, emit an update and start fetching.

## How to reproduce:
Add `await wait(3000)` to `getAdditionalPortfolio`

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -660,11 +660,6 @@ export class MainController extends EventEmitter {
       ? { forceUpdate, additionalHints }
       : { forceUpdate }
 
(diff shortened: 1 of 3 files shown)


PetromirDev:
@Ivshti cc

PR #730, opened 2024-04-24
Fix: declare submitted op failure if not found

borislav-itskov:
Problem:  
Sometimes the RPC fails to broadcast the transaction even if a success is returned from it OR the transaction enters the mempool for a given period of time only to be dropped after.

Fix: if the **transaction** (not the receipt) is not found 15 minutes after broadcast, declare it a failure

This was working like this for user operations as sometimes there's an error on the bundler side and the bundler doesn't broadcast the transaction without returning any feedback.  
Now, this is implemented for normal transactions as well.  
This gets rid of the "transaction broadcast" banner. I broadcast 1 txn on base, it got dropped and the banner got eternally stuck in the extension until this fix


jordan-enev (approved):
Nice! Can you please include the PR's description in the code you've changed because it's really worth and explanatory.

PR #734, opened 2024-04-24
Enhancement/ Reusable controller status logic in `EventEmitter`

PetromirDev:
## Summary:
The method `wrap[InsertControllerName]Action` was used in main, email-vault, keystore and settings. Making a change in one of the controllers always requires a change in every other implementation. 
This PR aims to reduce repetition by defining a reusable, enhanced `withStatus` method in the `eventEmitter`, because all controllers extend it.
The new implementation replaces `latestMethodCall` and `status` with `statuses`, an object in which every key is a method name and its corresponding value is the status of the method. This is done, because methods in the main controller should be callable in parallel, contrary to sub-controllers.

## Example of `statuses`:
```
{
  getEmailVaultInfo: 'INITIAL',
  uploadKeyStoreSecret: 'INITIAL',
  recoverKeyStore: 'INITIAL',
  requestKeysSync: 'INITIAL',
  finalizeSyncKeys: 'INITIAL'
}
```
## Front-end usage:
```
mainControllerState.status === 'SUCCESS' &&
mainControllerState.latestMethodCall === 'onAccountAdderSuccess'
```
is replaced with
```
mainControllerState.statuses.onAccountAdderSuccess === 'SUCCESS'
```

### DIFF
--- a/src/controllers/eventEmitter/eventEmitter.ts
+++ b/src/controllers/eventEmitter/eventEmitter.ts
@@ -94,7 +94,14 @@ export default class EventEmitter {
       this.forceEmitUpdate()
     } catch (error: any) {
       this.status = 'ERROR'
-      this.emitError(error)
+      if ('message' in error && 'level' in error && 'error' in error) {
+        this.emitError(error)
+      }
+      this.emitError({
+        message: error?.message || 'An unexpected error occurred',
+        level: 'major',
+        error
+      })
       this.forceEmitUpdate()
     }
 
--- a/src/controllers/eventEmitter/eventEmitter.ts
+++ b/src/controllers/eventEmitter/eventEmitter.ts
@@ -12,15 +12,9 @@ export type ErrorRef = {
   error: Error
 }
 
-export type Statuses<T extends Array<string>> = {
-  [key in T[number]]: 'INITIAL' | 'LOADING' | 'SUCCESS' | 'ERROR'
+export type Statuses<T extends string> = {
+  [key in T]: 'INITIAL' | 'LOADING' | 'SUCCESS' | 'ERROR'
 }
-
-export const getDefaultStatuses = (methods: string[]) =>
-  methods.reduce((acc, method) => {
-    acc[method] = 'INITIAL'
-    return acc
-  }, {} as Statuses<string[]>)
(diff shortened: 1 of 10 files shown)


Ivshti (approved):

src/controllers/keystore/keystore.ts
```diff
@@ -175,16 +199,21 @@ export class KeystoreController extends EventEmitter {
     const aesCtr = new aes.ModeOfOperation.ctr(derivedKey, counter)
     const mac = keccak256(concat([macPrefix, aesEncrypted.ciphertext]))
     if (mac !== aesEncrypted.mac) {
-      // Throw, because that's handled as a form field error
-      throw new Error('keystore: wrong secret')
+      this.errorMessage = 'Invalid Device Password.'
+      throw new EmittableError({
```
sonytooo, line 203: here we need only the form-level error and not the error toast so omit emitting an error and emit only an update

src/controllers/eventEmitter/eventEmitter.ts
```diff
@@ -77,6 +83,54 @@ export default class EventEmitter {
     for (const cb of this.#errorCallbacks) cb(error)
   }
 
+  protected async withStatus(callName: string, fn: Function, allowMultipleActions = false) {
```
sonytooo, line 86: consider renaming **allowMultipleActions** to **allowConcurrentActions** or **allowConcurrentActionCalls** and add some comments on what this prop is responsible for in its default state

src/controllers/settings/settings.ts
```diff
@@ -363,18 +366,25 @@ export class SettingsController extends EventEmitter {
...
       return
     }
-
     // make sure the network has not been added already
     const chainIds = this.networks.map((net) => net.chainId)
     if (chainIds.indexOf(BigInt(customNetwork.chainId)) !== -1) {
-      throw new Error('settings: addCustomNetwork chain already added')
+      throw new EmittableError({
+        message:
+          'Failed to detect network, perhaps an RPC issue. Please change the RPC and try again.',
```
JIOjosBG, line 374: Shouldn't the error message be about a network having already been added?

PR #736, opened 2024-04-25
Update / Portfolio Hints Structure

gergana95:
Proposed changes for portfolio hints structure, to be discussed better approaches where needed
[Issue](https://github.com/AmbireTech/ambire-app/issues/2109)


src/controllers/portfolio/portfolio.ts
```diff
@@ -28,42 +29,32 @@ import {
...
-    .map((token) => token.address)
+const TRESHOLD = 10
+
+/**
+ * Updates the previous hints storage with the latest portfolio get result.
+ * @param result - The portfolio get result.
+ * @param storagePreviousHints - The previous hints storage.
+ * @param key - The key to update in the previous hints storage.
+ * @returns The updated previous hints storage.
+ */
```
PetromirDev, line 50: Not needed, because the project uses typescript

src/controllers/portfolio/portfolio.ts
```diff
@@ -72,10 +63,25 @@ function getHintsWithBalance(
...
+    storagePreviousHints.learnedTokens = {
+      ...(storagePreviousHints.learnedTokens || {}),
+      [networkId]: {
+        ...((storagePreviousHints?.learnedTokens &&
+          storagePreviousHints?.learnedTokens[networkId]) ??
+          {}),
+        [address]: Date.now().toString()
+      }
+    }
+  })
```
PetromirDev, line 82: Pretty sure this code can be written in a more readable and performant way. Example(haven't tested it):
``` 
if (!storagePreviousHints.learnedTokens[networkId]) {
  storagePreviousHints.learnedTokens[networkId] = {}
}
erc20s.forEach((address) => {
  storagePreviousHints.learnedTokens[networkId][address] = Date.now().toString()
})
```

src/controllers/portfolio/portfolio.ts
```diff
@@ -28,42 +29,32 @@ import {
...
-    //     // Delete pinned tokens' hints if the user has > 1 non-zero tokens
-    //     (keepPinned &&
-    //       PINNED_TOKENS.find(
-    //         (pinnedToken) =>
-    //           pinnedToken.address === token.address && pinnedToken.networkId === token.networkId
-    //       ))
-    //   )
-    // })
-    .map((token) => token.address)
+const TRESHOLD = 10
```
PetromirDev, line 42: `THRESHOLD` typo

src/controllers/portfolio/portfolio.ts
```diff
@@ -28,42 +29,32 @@ import {
...
+
+/**
+ * Updates the previous hints storage with the latest portfolio get result.
+ * @param result - The portfolio get result.
+ * @param storagePreviousHints - The previous hints storage.
+ * @param key - The key to update in the previous hints storage.
+ * @returns The updated previous hints storage.
+ */
+function updatePreviousHintsStorage(
+  result: PortfolioGetResult,
```
PetromirDev, line 52: move to helpers

src/controllers/portfolio/portfolio.ts
```diff
@@ -664,6 +726,91 @@ export class PortfolioController extends EventEmitter {
...
+  async learnTokens(tokens: string[], networkId: NetworkId) {
+    const storagePreviousHints = await this.#storage.get('previousHints', {})
+
+    const learnedTokens = storagePreviousHints.learnedTokens || {}
+    const networkLearnedTokens = learnedTokens[networkId] || {}
+
+    // Get the limit for the network, refactor this maybe, this seems like a hack
+    const keyWithNetwork = [...this.#portfolioLibs.keys()].find((k) => k.includes(networkId))
+
+    const limit = this.#portfolioLibs[keyWithNetwork]?.value.deploylessTokens.isLimitedAt24kbData
```
PetromirDev, line 744: This part looks weird.

src/controllers/portfolio/portfolio.ts
```diff
@@ -664,6 +726,91 @@ export class PortfolioController extends EventEmitter {
...
+      }
+    })
+
+    const updatedPreviousHintsStorage = {
+      ...storagePreviousHints,
+      learnedTokens: {
+        ...storagePreviousHints.learnedTokens,
+        [networkId]: learnedTokens
+      }
+    }
```
PetromirDev, line 765: may be more performant like this:
```
const updatedPreviousHintsStorage = {...storagePreviousHints}
updatedPreviousHintsStorage.learnedTokens[networkId] = learnedTokens
```

PetromirDev (changes requested):

src/controllers/portfolio/portfolio.ts
PetromirDev: The same applies to all other functions

src/controllers/portfolio/portfolio.ts
gergana95: Yes, this part needs to be rethinked how to be changed

src/controllers/portfolio/portfolio.ts
```diff
@@ -746,7 +746,7 @@ export class PortfolioController extends EventEmitter {
       : LIMITS.deploylessStateOverrideMode
 
     // Reached network erc20 limit
-    if (limit.erc20 - Object.keys(networkLearnedTokens).length === TRESHOLD) {
```
Ivshti: can we use `>=` just in case

src/controllers/portfolio/portfolio.ts
```diff
@@ -664,6 +720,87 @@ export class PortfolioController extends EventEmitter {
...
+    const networkLearnedTokens = learnedTokens[networkId] || {}
+
+    const network = this.#settings.networks.find((x) => x.id === networkId)
+
+    if (!network) throw new Error('network not found')
+
+    const portfolioLib = this.initializePortfolioLibIfNeeded(accountId, networkId, network)
+
+    // Get the limit for the network
+    const limit = portfolioLib.deploylessTokens.isLimitedAt24kbData
```
Ivshti, line 739: this limit is not the same. This effectively means that we limit the learned tokens to one page of the deployless token mechanism, which makes sense to a big extent but at the same time not really because the RPC might change the limit, which is a bit chaotic

I recommend we just use a fixed number for this, eg 50

src/controllers/portfolio/portfolio.ts
```diff
@@ -664,6 +720,87 @@ export class PortfolioController extends EventEmitter {
...
+      ? LIMITS.deploylessProxyMode
+      : LIMITS.deploylessStateOverrideMode
+
+    // Reached network erc20 limit
+    if (limit.erc20 - Object.keys(networkLearnedTokens).length === THRESHOLD) {
+      await this.cleanLearnedTokens(networkId, limit.erc20, storagePreviousHints)
+    }
+    tokens.forEach((address) => {
```
Ivshti, line 747: we should use `for (..of..` instead of `forEach`

src/controllers/portfolio/portfolio.ts
```diff
@@ -664,6 +720,87 @@ export class PortfolioController extends EventEmitter {
...
+        networkLearnedTokens[address] = null
+      }
+    })
+    const updatedPreviousHintsStorage = { ...storagePreviousHints }
+    updatedPreviousHintsStorage.learnedTokens[networkId] = networkLearnedTokens
+
+    await this.#storage.set('previousHints', updatedPreviousHintsStorage)
+  }
+
+  #calculateTokensToDeleteCount(networkLearnedTokens: any, limit: number) {
```
Ivshti, line 759: this sounds excessive, why do we do this?

Ivshti: I think we should just always append tokens to this list, and AFTER that remove the oldest. That way we won't need such a complicated mechanism

Ivshti (changes requested):

src/libs/portfolio/helpers.ts
```diff
@@ -104,3 +104,46 @@ export const shouldGetAdditionalPortfolio = (account: Account) => {
...
+      collection.address,
+      result.hints.erc721s[collection.address]
+    ])
+  )
+  const previousHints = (hints.fromExternalAPI && hints.fromExternalAPI[key]) || {}
+
+  hints.fromExternalAPI = {
+    ...hints.fromExternalAPI,
+    [key]: { erc20s, erc721s }
+  }
```
PetromirDev, line 134: hints.fromExternalAPI[key] = { erc20s, erc721s }

PetromirDev (changes requested):

src/controllers/portfolio/portfolio.ts
```diff
@@ -664,6 +686,66 @@ export class PortfolioController extends EventEmitter {
     this.emitUpdate()
   }
 
+  /**
+   * Learn new tokens from humanizer and debug_traceCall
+   */
+  async learnTokens(tokens: string[], networkId: NetworkId) {
+    const storagePreviousHints = await this.#storage.get('previousHints', {})
```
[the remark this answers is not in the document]
PetromirDev: Can be combined with https://github.com/AmbireTech/ambire-common/pull/736#discussion_r1599744055

src/controllers/portfolio/portfolio.ts
```diff
@@ -618,33 +637,33 @@ export class PortfolioController extends EventEmitter {
...
-        if (
-          isSuccessfulLatestUpdate &&
-          !(accountState[network.id]!.result?.errors || []).find((err) => err.name === 'HintsError')
-        ) {
-          storagePreviousHints[key] = getHintsWithBalance(
-            accountState[network.id]!.result!
-            // !hasNonZeroTokens,
-            // this.#additionalHints
+        // Persist latest state in previousHints in the disk storage for further requests
+        if (isSuccessfulLatestUpdate && !areAccountOpsChanged && !forceUpdate) {
```
borislav-itskov, line 656: A small thing, but `forceUpdate` equals `opts?.forceUpdate || areAccountOpsChanged`, defined a bit above. So removing `!areAccountOpsChanged` seems safe as it's included in `forceUpdate`. No problem to keep it just in case, though.

And please tell me later why we don't call `getUpdatedHints` when forceUpdate is true

src/controllers/portfolio/portfolio.ts
```diff
@@ -109,6 +75,11 @@ export class PortfolioController extends EventEmitter {
 
   #additionalHints: GetOptions['additionalHints'] = []
```
borislav-itskov, line 76: We can delete these as they aren't used anymore.

Also, we can delete `resetAdditionalHints` and the places we use it

src/controllers/portfolio/portfolio.ts
```diff
@@ -596,6 +600,21 @@ export class PortfolioController extends EventEmitter {
...
         const forceUpdate = opts?.forceUpdate || areAccountOpsChanged
 
+        // Pass in learnedTokens as additionalHints only on areAccountOpsChanged
+        const fallbackHints = (storagePreviousHints?.fromExternalAPI &&
+          storagePreviousHints?.fromExternalAPI[key]) ?? {
+          erc20s: [],
+          erc721s: {}
+        }
+        const additionalHints =
+          ((areAccountOpsChanged || forceUpdate) &&
```
borislav-itskov, line 610: `areAccountOpsChanged` is included in `forceUpdate` but we can keep it if you think it's better just in case

src/controllers/portfolio/portfolio.ts
gergana95: We need to update only on latest state, so I think its safe to check only `!areAccountOpsChanged` and this is a mistake on my end.

Ivshti (changes requested):
need to be edited based on slack feedback

PetromirDev (approved):

src/controllers/portfolio/portfolio.ts
```diff
@@ -664,6 +652,46 @@ export class PortfolioController extends EventEmitter {
...
+    let networkLearnedTokens: { [key: string]: any } = learnedTokens[networkId] || {}
+
+    const tokensToLearn = tokenAddresses
+      .filter((address) => address !== ZeroAddress && !(address in networkLearnedTokens))
+      .reduce((acc: { [key: string]: null }, curr) => {
+        acc[curr] = null
+        return acc
+      }, {})
+
+    if (!tokensToLearn) return
```
PetromirDev, line 670: !tokensToLearn.length

src/libs/portfolio/helpers.ts
```diff
@@ -104,3 +106,79 @@ export const shouldGetAdditionalPortfolio = (account: Account) => {
...
+
+  if (Object.keys(previousHintsFromExternalAPI).length > 0) {
+    // eslint-disable-next-line no-restricted-syntax
+    for (const address of erc20s) {
+      const isPinned = PINNED_TOKENS.some(
+        (pinned) => pinned.address === address && pinned.networkId === networkId
+      )
+      const isTokenPreference = tokenPreferences.some(
+        (preference) => preference.address === address && preference.networkId === networkId
+      )
```
PetromirDev, line 144: Consider .toLowerCase()-ing address comparisons

src/controllers/portfolio/portfolio.ts
```diff
@@ -664,6 +652,46 @@ export class PortfolioController extends EventEmitter {
     this.emitUpdate()
   }
 
+  // Learn new tokens from humanizer and debug_traceCall
+  async learnTokens(tokenAddresses: string[] | undefined, networkId: NetworkId) {
+    if (!tokenAddresses) return
+    const storagePreviousHints = this.#previousHints
+    storagePreviousHints.learnedTokens = {}
+    const learnedTokens = storagePreviousHints.learnedTokens || {}
+    let networkLearnedTokens: { [key: string]: any } = learnedTokens[networkId] || {}
```
PetromirDev, line 661: replace any

PR #739, opened 2024-04-29
Refactor / Detect the private key import case in the AccountAdder via the `keyIterator` types

superKalo:
Improves a bit https://github.com/AmbireTech/ambire-common/pull/697. Instead of using the `pageSize` (workaround), detect the private key import case in the AccountAdder via the `keyIterator` types.


PetromirDev (approved):

Ivshti:
lgtm

PR #740, opened 2024-04-29
Refactor / Preselect private key accounts

superKalo:
Improves a bit https://github.com/AmbireTech/ambire-common/pull/697. Explicitly filter out the linked accounts from the `accountsOnPage` for the preselecting accounts logic in the private key import case when used accounts are found.


PetromirDev (approved):
Nice

PR #741, opened 2024-04-29
Fix / Typescript Errors caused by ambire-common on web wallet

gergana95:


superKalo:
Solves this problem: 

image

superKalo (approved):

PR #743, opened 2024-05-07
Add/ `removeUserRequests` public method in main

PetromirDev:
## More info in:
https://github.com/AmbireTech/ambire-app/pull/2177


PetromirDev:
Closing after https://github.com/AmbireTech/ambire-app/pull/2177#issuecomment-2097983762

PR #747, opened 2024-05-09
Transfer: Add checks for entered amount in transfer controller

JIOjosBG:
fix: We used to throw ethers error when the passed amount in the transfer screen was too small

### DIFF
--- a/src/controllers/transfer/transfer.ts
+++ b/src/controllers/transfer/transfer.ts
@@ -280,8 +280,16 @@ export class TransferController extends EventEmitter {
       ? FEE_COLLECTOR.toLowerCase()
       : this.recipientAddress.toLowerCase()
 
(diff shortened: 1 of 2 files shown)


src/controllers/transfer/transfer.ts
```diff
@@ -280,8 +280,16 @@ export class TransferController extends EventEmitter {
...
       : this.recipientAddress.toLowerCase()
 
+    if(parseFloat(this.amount)< 1/10**this.selectedToken.decimals)
+      return this.emitError({
+        level: 'major',
+        message:
+          'Token amount too low.',
+        error: new Error('transfer: user requested token transfer amount too low.')
+      })
+
```
PetromirDev, line 290: The idea is good, but this is not a good place to validate the amount. The amount should only be validated by `validateSendTransferAmount` so the user receives immediate feedback.

PetromirDev (changes requested):

PR #750, opened 2024-05-13
Notification Controller Refactoring

sonytooo:
Add ActionsController, a sub-controller part of main. The controller is responsible for storing all dapp requests and requests made from the transfer screen and visualizing them in an action-window.


src/controllers/actions/actions.ts
```diff
@@ -0,0 +1,154 @@
...
+
+export type UserRequestAction = {
+  id: UserRequest['id']
+  type: DappUserRequest['action']['kind']
+  userRequest: UserRequest
+}
+
+export type Action = AccountOpAction | SignMessageAction | BenzinAction | UserRequestAction
+
+export class ActionsController extends EventEmitter {
```
superKalo, line 35: Please add a few notes (as docs) for this controller.

superKalo (approved):

PR #753, opened 2024-05-14
Fix / Additional Portfolio In AccountPortfolio Lost in Pending State from Force Update

gergana95:
Currently on calculating account portfolio in the case of pending state we lose the additional portfolio networks, which erases them and doesnt show them on dashboard.

Reproduce:
1. Click on refresh portfolio button which causes forceUpdate
2. AdditionalPortfolio gets all data from pending state - but additional portfolio networks are not there.


src/libs/portfolio/portfolioView.ts
```diff
@@ -1,4 +1,7 @@
+import { Account } from 'interfaces/account'
```
PetromirDev, line 1: replace with a relative import

PR #761, opened 2024-05-22
Bug/ Infinite loading on selected account change in Transaction History/Signed Messages

PetromirDev:
## Before this PR:
Changing the `filters` param of the `activity` controller was resulting in two side effects:
1. Signed messages or signed account ops were filtered, depending on the history page
2. Transaction status banners on the dashboard were recomputed, based on the account 

This wasn't okay because activity banners on the dashboard should always be visualized for the currently selected account, while history pages can be filtered by either account. Because of the fact that activity banners should always be computed for the currently selected account `filters` had to be updated on selected account change.
```
-- main.ts:530
this.activity.init({ filters: { account: toAccountAddr } })
```
 As a consequence the existing filters were deleted, for example, if the user is on a txn history page.

## The bug:
A user could change the selected account from popup, which was updating `filters`, thus resulting in infinite loading because the filters in the UI don't match the filters in the BG. Adding a `useEffect` to match the two variables (tried here https://github.com/AmbireTech/ambire-app/pull/2234/files) isn't a viable option because the user may change the selected account while the history tab isn't focused and also results in unnecessary rerenders(because filters have to be updated twice)

## The fix:
The fix is quite simple- separate `filters` into two variables - `filters` and `#selectedAccount`. Use `filters` for history pages only and `selectedAccount` for current transactions/signed messages and other related things.

### DIFF
--- a/src/controllers/activity/activity.test.ts
+++ b/src/controllers/activity/activity.test.ts
@@ -5,6 +5,14 @@ import { AccountStates } from '../../interfaces/account'
 import { SettingsController } from '../settings/settings'
 import { ActivityController, SignedMessage, SubmittedAccountOp } from './activity'
 
+const INIT_PARAMS = {
+  selectedAccount: '0xB674F3fd5F43464dB0448a57529eAF37F04cceA5',
+  filters: {
+    account: '0xB674F3fd5F43464dB0448a57529eAF37F04cceA5',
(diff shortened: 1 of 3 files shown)


jordan-enev (approved):
Great findings and fixes! :rocket:

PR #768, opened 2024-05-27
Add pre-parsig for humanizer calls

JIOjosBG:
If `call.data` is `null` the humanizer breaks
test with `https://ambiretech.github.io/ambire-deploy-contracts/` wit hand EOA that is not used on ethereum

The fix is to parse such calls beforehand


Ivshti (commented):
lgtm

PR #769, opened 2024-05-27
Update NFT deployless return structure

JIOjosBG:
We used to fetch the nft uri from the deployless, which significantly limited the possible count of nfts per collection we can fetch because the the gas limit of `eth_call`. This PR removes the token uri from the deployless. 

The other part of the solution is to use the nft proxy to displlay the visuals from the nft address and token id.


Ivshti (approved):
tests still need to be updated probably

PetromirDev (approved):

Ivshti (approved):

Ivshti:
@borislav-itskov @jordan-enev tests need to be updated here, right? or no

Ivshti:
apparently tests are fine

PR #774, opened 2024-05-29
Config / Network ID to Coingecko platform ID (constant)

superKalo:
Adds one more config, a reversed structure of `coingeckoNets`.


PetromirDev:
Not needed as all network and platform ids are already included in `coingeckoNets`

PR #777, opened 2024-05-29
Fix/ Logic around portfolio hints refactoring

PetromirDev:
### Learned tokens were reset on every learnTokens call
```    
--- portfolio.learnTokens
    storagePreviousHints.learnedTokens = {} < ---
    const learnedTokens = storagePreviousHints.learnedTokens || {}
    ...
}
```
Together with that bugfix I simplified the function a bit

### lastSeenNonZeroAndNotInExternalAPIHints :see_no_evil:, or just the timestamp of learned tokens, wasn't updating
```
  const erc20s = result.tokens.filter((token) => token.amount > 0n).map((token) => token.address)
  
  const erc721s = Object.fromEntries(
    result.collections.map((collection) => [
      collection.address,
      result.hints.erc721s[collection.address]
    ])
  )
  const previousHintsFromExternalAPI =
    (hints.fromExternalAPI && hints.fromExternalAPI[key] && hints.fromExternalAPI[key]?.erc20s) ||
    []
 
  -----> We were putting the addresses of all tokens with balance, thus the condition !previousHintsFromExternalAPI.includes(address) was never satisfied
  hints.fromExternalAPI[key] = { erc20s, erc721s }
```
This one needed a bit more refactoring as code from the previous hints implementation was no longer needed and could remove the need of calculating hints from the external api. To be exact, `hints`, returned from the portfolio lib were no longer used because of the new structure. So, instead of returning all `hints` it was a better idea to return `hintsFromExternalAPI`, as all other hints were passed TO the lib and were already in the context of the controller. 
Only exposing `hintsFromExternalAPI` wasn't sufficient because the controller had to know when the external api call was successful and because the previous implementation returned `hintsFromExternalAPI` with empty arrays we had to do extra checks, so instead `hintsFromExternalAPI` is only truthy when its call is successful. If it isn't we don't update hints because:
- `learnedTokens` are already learned and their timestamps can't be updated because we don't have the response from the external api
- `fromExternalAPI` can't be updated because that would wipe "fallback" hints

## How to test the fixes:
1. Checkout the testing branch `test/fix-logic-around-portfolio-hints-refactoring`
2. Filter the output of the bg console by 'hints test:'
3. Use `https://app.uniswap.org/swap`, do transfers, block velcro requests, reload the extension as many times as you wish, delete previous hints from the extension storage and etc.

### DIFF
--- a/src/controllers/portfolio/portfolio.test.ts
+++ b/src/controllers/portfolio/portfolio.test.ts
@@ -421,7 +421,7 @@ describe('Portfolio Controller ', () => {
     })
   })
 
-  describe.only('Hints', () => {
+  describe('Hints', () => {
     test('Zero balance token is fetched after being learned', async () => {
       const BANANA_TOKEN_ADDR = '0x94e496474F1725f1c1824cB5BDb92d7691A4F03a'
       const { controller } = prepareTest()
--- a/src/controllers/portfolio/portfolio.test.ts
+++ b/src/controllers/portfolio/portfolio.test.ts
@@ -34,7 +34,8 @@ const prepareTest = () => {
   const controller = new PortfolioController(storage, settings, relayerUrl)
 
   return {
-    controller
+    controller,
+    storage
   }
 }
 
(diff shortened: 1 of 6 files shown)


gergana95 (approved):
Good! Will write more tests the following days!

PR #779, opened 2024-05-30
Config / Enhance Sign & Broadcast Error Handling

superKalo:
Tweaks a bit the error handling messaging so that:

- Fixed: Split the messages for the sign and the boardcast error cases, so that the messages are a bit more meaningful (previously Ambire was displaying a broadcast failed message when canceling a sign request by a hardware wallet for any reason).
  Prev:
  Screenshot 2024-05-31 at 17 35 53
  Now:
  Screenshot 2024-05-31 at 17 40 25
- Fixed: Encrease to 300 chars the limit for displaying long error messages, so that even the largest message gets displayed in full:
  Screenshot 2024-05-31 at 17 06 52
- Changed: normalize the most common not-so-human-readable Ledger and Trezor messages.


src/libs/trezor/trezor.ts
```diff
@@ -0,0 +1,22 @@
...
+  if (!errorCode && !errorMsg)
+    return 'Could not connect to your Trezor device. Please make sure it is connected.'
+
+  if (errorCode === 'Method_Interrupted')
+    return 'Closing the Trezor popup interrupted the connection.'
+
+  if (errorCode === 'Failure_ActionCancelled') return 'Rejected by your Trezor device.'
+
+  if (
+    errorMsg?.toLocaleLowerCase()?.includes('device disconnected during action') ||
```
PetromirDev, line 16: replace with toLowerCase just in case

PetromirDev (approved):

src/libs/trezor/trezor.ts
superKalo: ah, my Editor auto-completed this, fixed :)

PR #789, opened 2024-06-04
Simulation Improvements

sonytooo:
**fix: simulation when there are multiple pending accountOp actions**

* when there is an active signAccountOp we pass only the calls from that accountOp to be simulated
* when there is no active signAccountOp we pass all pending accountOps for the selectedAccount to be simulated and displayed on the Dashboard


borislav-itskov (approved):
:clap:

Ivshti (commented):
lgtm

PR #791, opened 2024-06-04
Fix: remove erc4337 settings from networkPreferences for predefined networks only

borislav-itskov:
Fix: remove erc4337 settings from networkPreferences for predefined networks only


Ivshti (commented):
lgtm

PR #801, opened 2024-06-07
 Update portfolio types and calculate total in the controller

PetromirDev:
## Changes:
- Made an attempt to simplify and improve portfolio types readability
- Moved `total` calculation from the lib to the controller


gergana95 (approved):

Ivshti (approved):

Ivshti (commented):
better than before for sure!

PR #808, opened 2024-06-11
Fix / PortfolioController race conditions

jordan-enev:
Closes https://github.com/AmbireTech/ambire-app/issues/1478.

-  **Fix: Implement a queue to prevent race conditions when calling `updateSelectedAccount`.** All calls are queued by network and account. Each time `updateSelectedAccount` is invoked to update the latest or pending state, the call is added to the queue. If a previous call is still running, the new call will be queued and executed only after the first one completes, regardless of whether it succeeds or fails. Before implementing this queue, multiple `updateSelectedAccount` calls made in a short period of time could cause the response of the latest call to be overwritten by a slower previous call.
- **Add a Jest test to ensure that the portfolio controller queue's logic chains the requests in order**.
- **Change: Pull out updatePortfolioState as a controller method, instead of an inner function of updateSelectedAccount, for better readability**.
- **Document how we should refactor portfolio force updates and apply account ops intent optimization once we implement [issue #2335](https://github.com/AmbireTech/ambire-app/issues/2335#issue-2345433944)**.
- **Ensure the queue doesn't result in a memory leak.** What's the difference between this article (https://cribl.io/blog/promise-chaining-memory-leak/) and our implementation? First of all, our portfolio method will always resolve, so we can't end up in a promise chain waiting for the first promise to resolve. Second, in the article, they are creating a recursive queue—if the promise doesn't resolve, they add the same method again to be invoked, and they do this for hundreds of calls. In our case, we don't have recursion. We simply chain the next portfolio update and don't retry the same method if it fails.


src/controllers/portfolio/portfolio.ts
```diff
@@ -404,7 +418,80 @@ export class PortfolioController extends EventEmitter {
     this.emitUpdate()
   }
 
-  // NOTE: we always pass in all `accounts` to ensure that the user of this
+  protected async updatePortfolioState(
+    accountId: string,
+    _accountState: AccountState,
+    network: Network,
```
PetromirDev, line 424: Isn't the naming convention `#updatePortfolioState` for private methods?

src/controllers/portfolio/portfolio.ts
```diff
@@ -52,6 +52,19 @@ export class PortfolioController extends EventEmitter {
...
   pending: PortfolioControllerState
 
+  // A queue to prevent race conditions when calling `updateSelectedAccount`.
+  // All calls are queued by network and account.
+  // Each time `updateSelectedAccount` is invoked to update the latest or pending state, the call is added to the queue.
+  // If a previous call is still running, the new call will be queued and executed only after the first one completes,
+  // regardless of whether it succeeds or fails.
+  // Before implementing this queue, multiple `updateSelectedAccount` calls made in a short period of time could cause
+  // the response of the latest call to be overwritten by a slower previous call.
+  queue: {
```
PetromirDev, line 62: rename to `#queue` because it will only be used internally by the ctrl

src/controllers/portfolio/portfolio.ts
```diff
@@ -52,6 +52,19 @@ export class PortfolioController extends EventEmitter {
...
+    [networkId: NetworkId]: {
+      [accountId: string]: Promise<void>
+    }
```
PetromirDev, line 65: why is the queue's structure not the same as AccountState's? (accountId:networkId)

src/controllers/portfolio/portfolio.ts
```diff
@@ -440,178 +526,138 @@ export class PortfolioController extends EventEmitter {
...
 
-          this.#previousHints = updatedStoragePreviousHints
-          await this.#storage.set('previousHints', updatedStoragePreviousHints)
+          // We cache the previously simulated AccountOps
+          // in order to compare them with the newly passed AccountOps before executing a new updatePortfolioState.
+          // This allows us to identify any differences between the two.
+          if (currentAccountOps) {
+            pendingState[network.id]!.accountOps = currentAccountOps
+          }
         }
```
PetromirDev, line 652: Perhaps this can be moved to a method as well?

src/controllers/portfolio/portfolio.ts
```diff
@@ -440,178 +526,138 @@ export class PortfolioController extends EventEmitter {
...
     await Promise.all(
       networks.map(async (network) => {
         const key = `${network.id}:${accountId}`
 
         const portfolioLib = this.initializePortfolioLibIfNeeded(accountId, network.id, network)
 
-        const currentAccountOps = accountOps?.[network.id]
+        const currentAccountOps = accountOps?.[network.id]?.filter(
+          (op) => op.accountAddr === accountId
+        )
```
PetromirDev, line 538: nice

PetromirDev (changes requested):

src/controllers/portfolio/portfolio.test.ts
```diff
@@ -88,48 +88,150 @@ describe('Portfolio Controller ', () => {
...
     })
+
+    expect(queueOrder).toEqual([
+      'updatePortfolioState - #1 call (latest state)',
+      'updatePortfolioState - #1 call (pending state)',
+      'updatePortfolioState - #2 call (latest state)',
+      'updatePortfolioState - #2 call (pending state)',
+      'updatePortfolioState - #3 call (latest state)',
+      'updatePortfolioState - #3 call (pending state)'
+    ])
```
PetromirDev, line 234: Can't this be simplified to something along the line of:
image

PetromirDev (changes requested):

src/controllers/portfolio/portfolio.test.ts
jordan-enev: unfortunately no, because of 2 reasons:

1. we can't delay one of the calls, as we did it with the mock
2. updateSelectedAccount resolves when all networks are resolved, but the queue is implemented per network and account, and we need to test a single queue.

src/controllers/portfolio/portfolio.ts
jordan-enev: it's a compromise, because a private method can't be mocked with Jest, will add a comment

src/controllers/portfolio/portfolio.ts
jordan-enev: I planned it in that way, but forgot to change its scope. Good point!

Ivshti (approved):

src/controllers/portfolio/portfolio.ts
jordan-enev: I considered the options and prefer to pull out only updatePortfolioState as a standalone method. Otherwise, it might be difficult to follow the implementation logic. Having updateSelectedAccount as the main method with an empty inner body that just calls other functions makes it harder to understand the order of execution. If the method becomes significantly larger, we will definitely refactor it accordingly.

PR #814, opened 2024-06-12
Fix: add custom network fee option

borislav-itskov:
Fix: when adding a custom network, add the `feeOption` to the correct place


Ivshti (commented):
lgtm

PR #817, opened 2024-06-13
Refactoring / Accounts

sonytooo:
* Added a required key **preferences** to the Account interface
* Introduced a new AccountsController that combines and encapsulates the common logic for accounts from the MainController and SettingsController
* tests refactoring
* Expose the forceEmitUpdate method publicly for each class, allowing it to be called externally

The new Account interface:
Screenshot 2024-06-18 at 11 57 36


src/controllers/main/main.ts
```diff
@@ -627,19 +461,26 @@ export class MainController extends EventEmitter {
...
             [this.actions.currentAction.accountOp.networkId]: [this.actions.currentAction.accountOp]
           }
-        : getAccountOpsByNetwork(selectedAccount, this.actions.visibleActionsQueue)
-    this.portfolio.updateSelectedAccount(this.accounts, selectedAccount, undefined, accountOps, {
-      forceUpdate
-    })
+        : getAccountOpsByNetwork(this.accounts.selectedAccount, this.actions.visibleActionsQueue)
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    this.portfolio.updateSelectedAccount(
+      this.accounts.accounts,
```
PetromirDev, line 476: We can pass the `accounts` controller to `portfolio`'s constructor so we don't have to pass `this.accounts.accounts` on every update(the same we did for networks)

src/interfaces/account.ts
```diff
@@ -2,6 +2,11 @@ import { Network } from './network'
 
 export type AccountId = string
 
+export type AccountPreferences = {
+  label: string
+  pfp: string // URL (https, ipfs or nft721://contractAddr/tokenId)
```
PetromirDev, line 7: I'm pretty sure that the comment isn't entirely correct as `pfp` currently stores the account address

PetromirDev (changes requested):

src/libs/account/account.ts
```diff
@@ -262,3 +278,34 @@ export const getAccountImportStatus = ({
...
+  if (existingAcc) return undefined
+
+  const number = prevAccounts.length + (i + 1)
+  return {
+    label: `Account ${number}`,
+    pfp: accountAddr // default pfp - a jazz icon generated from the addr
+  }
+}
+
+export function migrateAccountPreferencesToAccounts(
```
superKalo, line 299: Similarly to the `migrateNetworkPreferencesToNetworks`, I'd add a few notes since which version of the extension this migration is needed and some context why.

src/controllers/accountAdder/accountAdder.ts
```diff
@@ -573,14 +570,19 @@ export class AccountAdderController extends EventEmitter {
...
+      ...accounts.map(
+        (x, i) =>
+          ({
+            ...x.account,
+            preferences: getDefaultAccountPreferences(x.account.addr, this.#accounts.accounts, i),
+            newlyCreated: newlyCreatedAccounts.includes(x.account.addr)
+          } as Account & {
+            newlyCreated?: boolean | undefined
+          })
+      )
```
superKalo, line 582: Instead of `as Account & { newlyCreated?: boolean | undefined }`, this could be fixed if the `preferences` of the Account interface is optional, since it could be `undefined`, or not?

src/controllers/accountAdder/accountAdder.ts
```diff
@@ -864,8 +866,12 @@ export class AccountAdderController extends EventEmitter {
               factoryAddr,
               bytecode,
               salt
+            },
+            preferences: {
+              label: DEFAULT_ACCOUNT_LABEL,
+              pfp: addr
             }
-          },
+          } as Account,
```
superKalo, line 874: I see no need for casting to `Account`, the TypeScript on my end doesn't complain about this type. Please double-check on your end?

src/controllers/eventEmitter/eventEmitter.ts
```diff
@@ -57,7 +57,7 @@ export default class EventEmitter {
    * Without the `forceEmitUpdate` option, the application will only render the very first and last status updates,
    * batching the ones in between.
    */
-  protected async forceEmitUpdate() {
+  async forceEmitUpdate() {
```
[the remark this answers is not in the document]
sonytooo: Yes, while the getters are automatically updated in common, they are not updated on the FE. Therefore, all sub-controllers that use accounts should be force-updated to recalculate the getters on the FE

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable no-restricted-syntax */
+import { AccountsController } from 'controllers/accounts/accounts'
```
superKalo, line 2: Opsi, bad import.

src/controllers/addressBook/addressBook.test.ts
```diff
@@ -18,37 +23,55 @@ const MOCK_ACCOUNTS: Account[] = [
...
+)
 
 describe('AddressBookController', () => {
-  const addressBookController = new AddressBookController(storage, MOCK_ACCOUNTS, {
-    accountPreferences: MOCK_ACCOUNT_PREFERENCES
-  } as SettingsController)
+  let providersCtrl: ProvidersController
+  const networksCtrl = new NetworksController(
+    storage,
+    (net) => {
```
superKalo, line 64: Note: if https://github.com/AmbireTech/ambire-common/pull/815 gets merged first, need to pass the `fetch` as param after the `storage` here.

PR #821, opened 2024-06-17
Fix / Not Full Tokens List On Cached Data 

gergana95:
FIX: When we removed passing the previous tokens on this [last PR](https://github.com/AmbireTech/ambire-common/pull/794/files) this created an edge case which didnt populate the tokens from velcro ( but only from tokensList from jason.ambire.com ) on cached update from velcro. This is seen on the mobile app only since we refetch each time when app opens up.


superKalo (approved):
I tested against the prev tag (v1.0.15), the problem is not reproducible there. I wired up this fix and for my account (and the problematic accounts I found), the problem no longer exists. So looks good to me 👍

PR #824, opened 2024-06-18
Fix: none state-override rpcs

borislav-itskov:
Fix: if state override is not supported by the network, do not perform simulation; also, get the balances in a way that doesn't break the portfolio


Ivshti (changes requested):
simulations can work without state override too, using ``Detect` mode

borislav-itskov:
> simulations can work without state override too, using ``Detect` mode

Fixed

Ivshti:
Discussed with Bobi that this is a compromise solution, because in the case where Detect is used and we use normal mode, the internal call will not execute, but it will not fail either (because it will be to an EOA without code)

Ivshti (approved):

PR #825, opened 2024-06-19
Improvements / Simulation

sonytooo:
* Simulate only for the currently active accountOp for Basic Accounts
* Remove redundant FE code from common and move to ambire-app
* Fix: account in dApps does not change on accountSelect
* * Improve simulation calculations in portfolio


src/controllers/main/main.ts
```diff
@@ -471,19 +471,46 @@ export class MainController extends EventEmitter {
...
     if (!this.accounts.selectedAccount) return
 
-    // pass the accountOps if any so we could reflect the pending state
-    const accountOps =
-      this.actions.currentAction?.type === 'accountOp' && this.signAccountOp
-        ? {
-            [this.actions.currentAction.accountOp.networkId]: [this.actions.currentAction.accountOp]
+    const account = this.accounts.accounts.filter(
+      (a) => a.addr === this.accounts.selectedAccount
+    )[0]
```
borislav-itskov, line 476: We can use `find` instead of `filter` and we'll get the account directly, skipping the [0]

PR #827, opened 2024-06-19
Enhancement/ Improve UX when there is a slow or not working RPC

PetromirDev:
App PR and more info in: https://github.com/AmbireTech/ambire-app/pull/2374


src/controllers/main/main.ts
```diff
@@ -469,6 +470,15 @@ export class MainController extends EventEmitter {
     )
   }
 
+  async reloadSelectedAccount() {
+    if (!this.accounts.selectedAccount) return
+
+    await Promise.all([
+      this.accounts.updateAccountState(this.accounts.selectedAccount, 'latest'),
```
borislav-itskov, line 477: Let's change this to pending as I'm afraid this scenario could happen:
* the user broadcasts a transaction. Because of this, we start fetching the pending account state every x seconds until there are no more `BroadcastButNotConfirmed` transactions
* for some reason, he clicks the refresh button on his dashboard and instead of the pending, he gets the latest state
* get tries to broadcast a transaction really quickly after and a nonce discrepancy error appears as there's a conflict between our nonce (the latest one, which is behind) and the newest one, which is the correct one

Overall, we should fetch the latest state on initial load only. After, we should tend to fetch the pending state OR have a check to see if there are unconfirmed transactions on this network for this account (which I don't see a reason why to implement, better to just call the pending state)

borislav-itskov (changes requested):
Let's change `reloadSelectedAccount` to fetch the account state from the pending block

borislav-itskov (approved):

Ivshti (approved):

PR #829, opened 2024-06-24
Bug / Migrate Accounts Without Preferences

sonytooo:
Fix: handle cases where there are accounts without preferences in storage during accountsMigration


superKalo (approved):
For the archive: we had 11 failing E2E tests because of this missing fallback (that wasn’t handling the case where you might have an account without preferences).

Nonetheless, that’s not a real problem IMO, since importing an account with missing default preferences is not a valid use case since many versions (it was possible in the early versions of the extension, prev the beta one).

So I'll tweak the E2E .env variables too, but that's nice to have as a fallback, so approve ✅

PR #830, opened 2024-06-25
Fix: L2 broadcast no erc-4337

borislav-itskov:
Change log:
* EP authorization only for 0.7.0 accounts
* 0.6.0 accounts can only broadcast with EOA - fix the L2 broadcast so txns are successfull by including estimate_gas in the erc-4337 estimation if there's no erc 4337 estimation
* remove the custom Arbitrum estimation and rely on estimate_gas instead


Ivshti (commented):
lgtm

PR #836, opened 2024-06-27
Update/portfolio block number for portfolio view

gergana95:
Changelog: Update the portfolioView helper to compare latest and pending by blockNumber, in case we dont have signAccountOp active  - in this case we always use pending state. Otherwise - get the pending state in case is newer than  latest.


Ivshti (approved):
just look at Petyo's suggestion if still relevant

PetromirDev (approved):

PR #840, opened 2024-07-01
Feature/learn from trace call

borislav-itskov:
Change log:
* abstract `getAccountOpsForSimulation`
* return a boolean on `learnTokens` indicating whether they have learned something new or not
* implement debug_traceCall: 1) the original call for EOA 2) wrapping the calls into execute/deployAndExecute for SA 3) using state override where possible for giving funds to the broadcasting account so it doesn't break with insufficient funds 4) calling getBalances at the end to figure out which addresses are tokens
* pass the tokens from debug_traceCall to learnTokens. If something new has been learned, update the portfolio along with the pending state

Because we need the correct gas and gas price in order for this to work, we execute it once on the signAccountOp screen if there are no estimation errors. More info about this in the ambire-app PR


Ivshti (approved):

src/libs/main/main.ts
```diff
@@ -107,3 +110,20 @@ export const makeBasicAccountOpAction = ({
     accountOp
   }
 }
+
+export const getAccountOpsForSimulation = (
+  account: Account,
+  visibleActionsQueue: Action[],
+  // pass the account op along only in the case of a basic account
+  // with an open sing account op screen
```
Ivshti, line 118: typo

Ivshti (approved):

PR #845, opened 2024-07-02
Add errors on wrong signer requested by dapp for signing

JIOjosBG:
When the dapp requests a signature for a signer that is not the selected signer we should return an error


sonytooo (approved):

Ivshti (commented):
lgtm but we can work on the messages a bit futher. Will DM

PR #850, opened 2024-07-03
Feature / Claim Wallet Rewards

gergana95:
Create sign account op function in transfer helpers to claim wallet rewards.


Ivshti (changes requested):
I don't think mainCtrl  should get even more cluttered than it is and gain functionalities like this - let's think of a better solution

src/controllers/main/main.ts
```diff
@@ -681,6 +681,33 @@ export class MainController extends EventEmitter {
     await this.addUserRequest(userRequest)
   }
 
+  async claimWallet(
+    selectedToken: TokenResult,
+    claimableRewardsData: any,
```
PetromirDev, line 686: replace any

src/libs/portfolio/interfaces.ts
```diff
@@ -87,7 +87,9 @@ interface Total {
 type AdditionalPortfolioProperties = 'updateStarted' | 'tokens'
 
 // Create the final type with some properties optional
-export type AdditionalPortfolioNetworkResult = Partial<PortfolioLibGetResult> &
+export type AdditionalPortfolioNetworkResult = Partial<PortfolioLibGetResult> & {
+  claimableRewardsData: any
```
PetromirDev, line 91: just add it next to total. Also replace any

src/libs/transfer/userRequest.ts
```diff
@@ -14,6 +16,38 @@ interface BuildUserRequestParams {
...
 }
 
+const supplyControllerAddress = '0xA69B8074CE03A33B13057B1e9D37DCDE0024Aaff'
+const WALLET_STAKING_ADDR = '0x47Cd7E91C3CBaAF266369fe8518345fc4FC12935'
+
+function buildClaimWalletRequest({
+  selectedAccount,
+  selectedToken,
+  claimableRewardsData,
+}: { selectedAccount: string, selectedToken: TokenResult, claimableRewardsData: any }): SignUserRequest | null {
```
PetromirDev, line 26: replace any

src/libs/transfer/userRequest.ts
```diff
@@ -14,6 +16,38 @@ interface BuildUserRequestParams {
   recipientAddress: string
 }
 
+const supplyControllerAddress = '0xA69B8074CE03A33B13057B1e9D37DCDE0024Aaff'
+const WALLET_STAKING_ADDR = '0x47Cd7E91C3CBaAF266369fe8518345fc4FC12935'
```
PetromirDev, line 20: Can we add these constants to some common file?

PetromirDev (changes requested):
Fix tests

src/libs/transfer/userRequest.ts
```diff
@@ -1,11 +1,14 @@
 import erc20Abi from 'adex-protocol-eth/abi/ERC20.json'
 import { Interface, parseUnits } from 'ethers'
-import { SignUserRequest } from 'interfaces/userRequest'
+import { Call, SignUserRequest } from 'interfaces/userRequest'
```
PetromirDev, line 3: fix import

PetromirDev (changes requested):

PR #851, opened 2024-07-03
Refactor / Sign message flow improvements by moving logic to the Main controller (that gets executed in the background)

superKalo:
Refactors a bit the sign message flow, in order to:

- Changed: Improve the sign message flow by moving to the Main controller (background) the logic that signs and then "broadcasts" the signed message. Previously this logic was triggered by the extension front-end (on controller update), which we agreed is not a reliable pattern.
- Changed: Prioritize to resolve the successfully signed message quicker and then add it to the Activity controller later.
- Changed: Migrate the sign message method to the `withStatus` wrapper (for more consistent status and error handling).
- Fixed: Sign message flow was misleadingly re-using the `broadcastStatus` used for the Account Op broadcasting flow.


superKalo:
One test fails (`Portfolio › batching works`), but I was told it's fixed in this feature branch: https://github.com/AmbireTech/ambire-common/pull/850

PetromirDev (approved):

src/controllers/main/main.ts
```diff
@@ -1522,11 +1530,14 @@ export class MainController extends EventEmitter {
     this.emitUpdate()
   }
 
+  // TODO: Figure out better name for this method
+  // TODO: Re-use `withStatus` instead of the broadcastStatus flag
+  // TODO: The `broadcastStatus` flag gets re-used here and for signing an
+  // account op too, that could lead to troubles. It should be separated.
   async broadcastSignedMessage(signedMessage: SignedMessage) {
```
sonytooo, line 1537: I suggest removing that method and large comment and integrating the code directly into the signMessageSign function because we really do not broadcast anything

sonytooo (approved):

PR #855, opened 2024-07-05
Enhancement/ Improve portfolio related UX when an RPC is down

PetromirDev:
Fixes 1, 2 and 4 from https://github.com/AmbireTech/ambire-app/issues/2434


gergana95 (approved):
Better this way! Works great 🙌

PR #860, opened 2024-07-10
Fix / Select Default Account

sonytooo:
Fixed: select a default account after adding new accounts


superKalo (approved):

superKalo:
Note: we discussed offline that we will remove it from here and remove the `if (!this.selectedAccount)` in the Accounts controller -> addAccounts.

PR #861, opened 2024-07-10
Refactor / Sign (and broadcast) account op flow improvements

superKalo:
Refactors a bit the sign (and broadcast) account op flow, in order to:

- Added: Expose the fee payer key during the account op broadcast (that temporarily gets automatically selected to be the key with the same type as the signing key or the first key available).
- Changed: Migrate the Main controller `broadcastAccountOp` method to the `withStatus` wrapper (for more consistent status and error handling).
- Changed: Improve the broadcast signed account op flow by introducing `handleSignAccountOp` (similar to the `handleSignMessage`) instead of listening on controller update.
- Change: Improve a bit the broadcast error messages (shorten the account or key addresses, add a generic contact support to the ones that have not, tweak the wording a bit for some)


src/controllers/main/main.ts
```diff
@@ -1357,37 +1356,41 @@ export class MainController extends EventEmitter {
...
-    accountOp: AccountOp,
-    estimation: EstimateResult,
-    actionId: AccountOpAction['id']
-  ) {
-    this.broadcastStatus = 'LOADING'
-    this.emitUpdate()
+  async #broadcastSignedAccountOp() {
+    const accountOp = this.signAccountOp?.accountOp
+    const estimation = this.signAccountOp?.estimation
+    const actionId = this.signAccountOp?.fromActionId
```
PetromirDev, line 1362: const { accountOp, estimation, fromActionId: actionId } = this.signAccountOp || {}

src/controllers/main/main.ts
```diff
@@ -1409,14 +1412,18 @@ export class MainController extends EventEmitter {
...
-          return this.#throwAccountOpBroadcastError(
+          return await this.#throwBroadcastAccountOp(
             new Error(
-              `Key with address: ${accountOp.gasFeePayment!.paidBy} for account with address: ${
-                accountOp.accountAddr
-              } not found`
+              `Key with address ${shortenAddress(
+                accountOp.gasFeePayment!.paidBy,
+                13
+              )} for account with address ${shortenAddress(accountOp.accountAddr, 13)} not found.`
```
PetromirDev, line 1420: `#throwBroadcastAccountOp` is used ~13 times, 12 of which initialize a new Error(). Imo this can be simplified by making `#throwBroadcastAccountOp` accept an object as a param with 2 keys:
```
error?: Error
message?: string
```
, then the 12 calls can be simplified to `this.#throwBroadcastAccountOp({ message: 'something' })`

and then in the method:
```
let message = message || 'Unable to broadcast the transaction.'
...
return Promise.reject(new EmittableError({ level: 'major', message, error || new Error(message) }))
```

src/controllers/main/main.ts
```diff
@@ -1602,41 +1616,38 @@ export class MainController extends EventEmitter {
...
+        networkId: network.id,
+        isUserOp: !!accountOp?.asUserOperation
+      },
+      actionId
+    )
 
-    this.broadcastStatus = 'INITIAL'
-    this.emitUpdate()
+    this.onSignSuccess('account-op')
+    return Promise.resolve(submittedAccountOp)
```
PetromirDev, line 1650: Is this needed?

src/utils/shortenAddress.ts
```diff
@@ -0,0 +1,6 @@
+const shortenAddress = (address: string, maxLength: number = 30): string =>
+  address.length <= maxLength
+    ? address
```
PetromirDev, line 3: Nice

PetromirDev (approved):

borislav-itskov (approved):

sonytooo (approved):

src/controllers/main/main.ts
superKalo: Yes, I see your point. Refactored in https://github.com/AmbireTech/ambire-common/pull/861/commits/542b462ab6abb324aaefba3661b0c3dd99f509aa

src/controllers/main/main.ts
superKalo: Otherwise, TypeScript complains that not all code paths return a value. I could have returned void (`Promise.resolve()`), but I returned the submitted account op instead.

PR #867, opened 2024-07-12
Fix/4337 gas price updates

borislav-itskov:
Problem:
* gas prices were not updated every 12s on erc-4337 broadcast as they happened during estimation (which now happens once every 60s)

Change log:
* introduce `bundlerGasPrices` that are fetched along with normal gas prices if the network is a 4337 one
* if the estimation succeeds in estimating an ERC-4337 UserOp, update the gas prices by using the `bundlerGasPrices`
* if the estimation does NOT succeed in estimating an ERC-4337 UserOp, update the gas prices by using our standard gas price mechanism
* if supported, re-enable 4337 on a custom network. In the past, we used only 4337 and that was a problem as if the bundler fails, the user is stuck. Now we have EOA broadcast so re-enabling 4337 just gives more options to the user

We need both in order to be sure we're not dependant on the bundler. If the bundler fails, we fallback to our gas prices


src/controllers/main/main.ts
```diff
@@ -503,31 +506,35 @@ export class MainController extends EventEmitter {
...
+    const network = this.networks.networks.find((net) => net.id === accOp.networkId)
+    if (!network) return // shouldn't happen
+
+    const is4337 = isErc4337Broadcast(
+      network,
+      this.accounts.accountStates[accOp.accountAddr][accOp.networkId]
     )
+    const bundlerFetch = async () => {
+      if (!is4337) return null
+      return Bundler.fetchGasPrices(network).catch(() => null)
```
Ivshti, line 522: is swallowing the error here ok? maybe at least log it informatively?

src/controllers/main/main.ts
```diff
@@ -1171,7 +1178,13 @@ export class MainController extends EventEmitter {
     // the time we get here. To prevent issues, we check one more time
     if (!this.signAccountOp) return
 
-    this.signAccountOp.update({ gasPrices: this.gasPrices[networkId] })
+    this.signAccountOp.update({
+      gasPrices: this.gasPrices[networkId],
+      bundlerGasPrices:
+        this.bundlerGasPrices && this.bundlerGasPrices[networkId]
```
Ivshti, line 1184: why? can't we just alwaysa initialize `this.bundlerGasPrices` and just use `this.bundlerGasPrices[networkId]`. Or simply `this.bundlerGasPrices?[networkId]`? but I vote for always initialize

src/controllers/main/main.ts
```diff
@@ -1708,12 +1721,20 @@ export class MainController extends EventEmitter {
...
-    // Enhance the error incoming for this corner case
-    if (message.includes('insufficient funds'))
-      message = 'Insufficient funds to cover the fee for broadcasting the transaction.'
-
-    // Trip the error message, errors coming from the RPC can be huuuuuge
-    message = message.length > 300 ? `${message.substring(0, 300)}...` : message
+    if (message) {
+      if (message.includes('insufficient funds')) {
+        // TODO: Better message?
+        message = 'Insufficient funds for intristic transaction cost'
```
Ivshti, line 1727: yea, we need a better message. who doesn't have enough funds? you don't have enough native?

Ivshti (changes requested):

PR #868, opened 2024-07-12
Do not show nonce discrepancy error

borislav-itskov:
Fix:
* do not show the nonce discrepancy error, instead just wait for the re-estimation to finish


PetromirDev (approved):

Ivshti (commented):
lgtm

PR #872, opened 2024-07-12
Bug / Agressively reloading Dashboard race condition (resulting a "Please wait for the completion of the previous action..." warn)

superKalo:
The reload should either 1) cancel the previous update account state or 2) wait until the previous one has been completed before starting a new one.

I applied the second option since it's easier with the current code state. But TBD with @gergana95 and @PetromirDev if that's the best approach. Probably no 🙄


superKalo:
Closed in favor of https://github.com/AmbireTech/ambire-common/pull/873 , until we have a broader discussion what's the best approach for this.

PR #877, opened 2024-07-15
Update/ Allow concurrent actions for some methods in `accounts` ctrl

PetromirDev:
## Changes:
- Allowed concurrent actions for select account and preference updates because the actions may happen in very close succession during import
- Fixed a few ts/eslint errors


superKalo (approved):
Sounds good to me 👍

PR #878, opened 2024-07-15
Fix / Connect Dapp on an Uninitialized Extension

sonytooo:
Fixed: When a dApp requested to connect and the extension was not yet initialized, the request would hang in the wallet without providing a proper response to the dApp.

After the fix, the expected behavior of the extension should be as follows: If the extension is not initialized and receives a "Connect" request from a dApp, it should open the onboarding flow in an action window. Once the user sets up the wallet, the "Connect" request should be prompted in the same action window.

TODO: we should think of how to scale down the content in an action window on small screens
Screenshot 2024-07-15 at 12 54 21


PetromirDev (approved):
Good job

Ivshti (commented):
lgtm

PR #879, opened 2024-07-15
Fix / Learned tokens: Remove non-ERC20s

jordan-enev:
Closes https://github.com/AmbireTech/ambire-app/issues/2417.

---

**Introduce a Self-cleaning mechanism for removing non-ERC20 items from the learned tokens.**

There's a possibility that the discovered tokens (from debug_traceCall or mostly Humanizer) include items that are not ERC20 tokens. For instance, a SmartContract address can be passed as a learned token.

Thanks to BalanceGetter, we know which tokens encounter an error when we try to update the portfolio.
All the errors are collected in `tokenErrors`, and if we cannot retrieve its balance, the contract returns `bytes('unkn')`, which is equal to `0x756e6b6e`.

**Note:**
When we extract tokens from `debug_traceCall`, we are already filtering the tokens the same way as here (relying on BalanceGetter).
However, for the Humanizer tokens, we skipped that check because the Humanizer is invoked more frequently on the Sign screen, and this validation may slow down the performance of the page. Because of this, we perform the check here, where we are calling BalanceGetter anyway.

### DIFF
--- a/src/controllers/portfolio/portfolio.test.ts
+++ b/src/controllers/portfolio/portfolio.test.ts
@@ -525,6 +525,23 @@ describe('Portfolio Controller ', () => {
 
       expect(token).toBeTruthy()
     })
+
+    test('Learned tokens to avoid persisting non-ERC20 tokens', async () => {
+      const BANANA_TOKEN_ADDR = '0x94e496474F1725f1c1824cB5BDb92d7691A4F03a'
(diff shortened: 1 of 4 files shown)


borislav-itskov (approved):

borislav-itskov (approved):

PetromirDev (commented):
lgtm

PR #880, opened 2024-07-15
Refactor / Require all prerequisites needed in the Sign Message controller and improve the controller tests

superKalo:
Covers corner cases for the Sing Message controller:

- Fixed: Require all prerequisites when initiating the controller (Accounts controller to have been loaded) and when signing (to have pulled at least once an account state for the network on which a sign message request gets received).

Also, improves a bit the Sign Message tests by:

- Change: Initiate a real `AccountsController` (instead of a dummy one)
- Change: Migrate the pattern for writing tests to the one proposed here: https://github.com/AmbireTech/ambire-app/issues/2454

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -444,6 +444,15 @@ export class MainController extends EventEmitter {
   }
 
   async handleSignMessage() {
+    const accountAddr = this.signMessage.messageToSign?.accountAddr
+    const networkId = this.signMessage.messageToSign?.networkId
+
+    // Could (rarely) happen if not even a single account state is fetched yet
+    const shouldForceUpdateAndWaitForAccountState =
+      accountAddr && networkId && !this.accounts.accountStates[accountAddr]?.[networkId]
+    if (shouldForceUpdateAndWaitForAccountState)
+      await this.accounts.updateAccountState(accountAddr, 'latest', [networkId])
+
     await this.signMessage.sign()
 
     const signedMessage = this.signMessage.signedMessage
--- a/src/controllers/signMessage/signMessage.test.ts
+++ b/src/controllers/signMessage/signMessage.test.ts
@@ -48,6 +48,7 @@ const messageToSign: Message = {
 describe('SignMessageController', () => {
   let signMessageController: SignMessageController
   let keystore: KeystoreController
+  let accountsCtrl: AccountsController
 
   beforeEach(async () => {
     const storage = produceMemoryStore()
@@ -67,7 +68,7 @@ describe('SignMessageController', () => {
     providersCtrl = new ProvidersController(networksCtrl)
     providersCtrl.providers = providers
 
-    const accountsCtrl = new AccountsController(
+    accountsCtrl = new AccountsController(
       storage,
       providersCtrl,
       networksCtrl,
@@ -130,6 +131,7 @@ describe('SignMessageController', () => {
     expect(signMessageController.signingKeyType).toBe('internal')
   })
 
+  // TODO: Would be better to test the signing via the Main controller -> handleSignMessage instead
   test('should sign a message', async () => {
     const signingKeyAddr = account.addr
     const dummySignature =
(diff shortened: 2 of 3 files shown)


src/controllers/signMessage/signMessage.test.ts
```diff
@@ -37,106 +37,74 @@ const account: Account = {
...
-    } as AccountsController
+    const accountsCtrl = new AccountsController(
+      storage,
+      providersCtrl,
+      networksCtrl,
+      () => {},
+      () => {}
+    )
+    // Internally, it waits Networks and Providers to be loaded
+    await accountsCtrl.initialLoadPromise
```
sonytooo, line 78: The SignMessageController should wait for accounts to be initialized and ready in most of its methods. You should not wait for that in the test. Otherwise, you test only the happy case when everything is loaded

src/controllers/signMessage/signMessage.test.ts
```diff
@@ -37,106 +37,74 @@ const account: Account = {
...
+
+    await accountsCtrl.updateAccountState(account.addr, 'latest')
```
sonytooo, line 80: same here: do not "await"

src/controllers/signMessage/signMessage.test.ts
superKalo: Yes, I agree. Changed in https://github.com/AmbireTech/ambire-common/pull/880/commits/e24c57df77b0662b9d4d7e45a0348c251917f47a and https://github.com/AmbireTech/ambire-common/pull/880/commits/b754f3c4ec1440482639b57858004979e40f11a1

src/controllers/signMessage/signMessage.test.ts
superKalo: Yes, no longer needed after the changes I mentioned in https://github.com/AmbireTech/ambire-common/pull/880#discussion_r1679079120

src/controllers/signMessage/signMessage.ts
```diff
@@ -180,7 +188,13 @@ export class SignMessageController extends EventEmitter {
         throw new Error('Network not supported on Ambire. Please contract support.')
       }
 
-      const accountState = this.#accounts.accountStates[account.addr][network!.id]
+      // Could (rarely) happen if not even a single account state is fetched yet
+      const shouldForceUpdateAndWaitForAccountState =
+        !this.#accounts.accountStates[account.addr]?.[network.id]
+      if (shouldForceUpdateAndWaitForAccountState)
+        await this.#accounts.updateAccountState(account.addr, 'latest', [network.id])
```
sonytooo, line 195: anti-pattern

sonytooo: only the main ctrl should call methods of another sub-ctrl to update the state

superKalo:
Simo said in an offline discussion, approved! :)

src/controllers/main/main.ts
```diff
@@ -444,6 +444,15 @@ export class MainController extends EventEmitter {
...
 
   async handleSignMessage() {
+    const accountAddr = this.signMessage.messageToSign?.accountAddr
+    const networkId = this.signMessage.messageToSign?.networkId
+
+    // Could (rarely) happen if not even a single account state is fetched yet
+    const shouldForceUpdateAndWaitForAccountState =
+      accountAddr && networkId && !this.accounts.accountStates[accountAddr]?.[networkId]
+    if (shouldForceUpdateAndWaitForAccountState)
+      await this.accounts.updateAccountState(accountAddr, 'latest', [networkId])
```
Ivshti, line 454: should we blow up if even after this, the account state is not there?

Ivshti (commented):
question

src/controllers/main/main.ts
superKalo: Yes, I think so. Improved in https://github.com/AmbireTech/ambire-common/pull/885

PR #881, opened 2024-07-16
Fix: react to incorrect user op nonce

borislav-itskov:
If the estimation returns an AA25 wrong account nonce error, update the account state and re-estimate but show the available EOA options before that


Ivshti (approved):
a bit hacky but if it works, it's fine

Ivshti (commented):
sounds reasonable

PR #893, opened 2024-07-20
Add basic nft simulation diffs

JIOjosBG:
Used for displayng the simulated nft balance change on the FE

Can be tested with the branch from this PR:
 - https://github.com/AmbireTech/ambire-app/pull/2481


src/libs/portfolio/getOnchainBalances.ts
```diff
@@ -176,14 +176,23 @@ export async function getNFTs(
       : null
 
     const token = mapToken(beforeToken)
+    const [receiving, sending]: [bigint[], bigint[]] = [[], []]
+    beforeToken[2].forEach((oldCollectible: bigint) => {
```
jordan-enev, line 180: here, it would be better to use named properties from mapToken (i.e. `token.collectibles`), instead of indices, as it will improve the readability

src/libs/portfolio/getOnchainBalances.ts
```diff
@@ -176,14 +176,23 @@ export async function getNFTs(
...
+    })
 
     return [
       beforeToken.error,
       {
         ...token,
         // Please refer to getTokens() for more info regarding `amountBeforeSimulation` calc
         simulationAmount: simulationToken ? simulationToken.amount - token.amount : undefined,
-        amountPostSimulation: simulationToken ? simulationToken.amount : token.amount
+        amountPostSimulation: simulationToken ? simulationToken.amount : token.amount,
```
jordan-enev, line 194: simulationAmount and amountPostSimulation are now redundant

src/libs/portfolio/getOnchainBalances.ts
```diff
@@ -176,14 +176,23 @@ export async function getNFTs(
       : null
 
     const token = mapToken(beforeToken)
+    const [receiving, sending]: [bigint[], bigint[]] = [[], []]
```
jordan-enev, line 179: why not simply: const receiving: bigint[] = []; const sending: bigint[] = []

jordan-enev (approved):
some minor comments, but generally lgtm

src/libs/portfolio/getOnchainBalances.ts
JIOjosBG: we should only rely only on `sending` and `receinving` for any changes to the balance in regards to NFTs in the context of executing an account op.

if we want to get the static *`'amount'`* for the collection we could refer to `.collectibles.length` anyway

if we take this approach the collectibles array should be from before the simulation

JIOjosBG: wyt @Ivshti @jordan-enev

src/libs/portfolio/getOnchainBalances.ts
JIOjosBG: wanted to minimize non-logic bearing lines, but ok

src/libs/portfolio/getOnchainBalances.ts
jordan-enev: Yes, the simulationAmount and amountPostSimulation fields represent a numeric/static change in the collectibles' length. The properties that Yosif introduced, sending and received, are much more informative. In terms of NFTs, both simulationAmount and amountPostSimulation seem irrelevant to me.

PR #896, opened 2024-07-24
Fix / Pending Actions Order

sonytooo:
Fixes: https://github.com/AmbireTech/ambire-app/issues/2496 

Add an action to the queue only if there's already one of the same type from the same dApp. If not, prioritize the new action and put it at the start of the queue so it's displayed first when the action window opens. 
Basically, every new action gets top priority unless there's already an action of the same type from the same dApp in the queue.


Ivshti (approved):

PetromirDev (approved):
Good job

PR #900, opened 2024-07-24
WIP: Enhancement/ Safety level ranking flag to tokens

PetromirDev:


src/libs/portfolio/helpers.ts
```diff
@@ -347,6 +348,32 @@ export const tokenFilter = (
...
 }
 
+const getTokenSafetyLevel = (token: TokenResult) => {
+  let safetyLevel = 'unknown' as 'unknown' | 'trusted' | 'spoof'
+
+  if (token.address === ZeroAddress) {
+    return 'trusted'
+  }
+
+  Object.values(humanizerInfo.knownAddresses).forEach((value) => {
```
Ivshti, line 358: kolega, this code is quote hard to read as-is, because you have to follow iteration

can't you do
```
const matchingEntry = humanizerInfo.knownAddresses.find(...)
if (matchingEntry) {
   // found in humanizer
   const isTrusted  = ...
}
```

but also I don't get the logic in general, why would the same address token have a different symbol? I think it would be better to search the humanizer entries for ones that have the same exact symbol but diff address, and in that case mark it as spoof

Ivshti (changes requested):
I don't understand the logic and I think the code can be rewritten to be better

src/libs/portfolio/helpers.ts
PetromirDev: I've written it this way because the humanizer isn't divided by networks.

JIOjosBG (changes requested):

src/libs/portfolio/helpers.ts
```diff
@@ -347,6 +350,44 @@ export const tokenFilter = (
...
+
+  if (token.address === ZeroAddress || hasFlags) return 'trusted'
+
+  const canTokenBeChecked = networks.some((network) => network.id === token.networkId)
+
+  // If the token is on a custom network we can't check if it's spoof
+  // thus we can't trust it
+  if (!canTokenBeChecked) return undefined
+
+  Object.values(humanizerInfo.knownAddresses as unknown as HumanizerMeta).some((value) => {
```
JIOjosBG, line 366: add comment about `.some()` doing a `break` on `return true` and `continue` on `return false`

JIOjosBG (approved):
lgtm, really liked how you optimized for long arrays of tokens

PR #904, opened 2024-07-25
Add assetInfo service

JIOjosBG:
Service for token/nft info batch fetching

can be tested with this for benzin
https://github.com/AmbireTech/ambire-app/pull/2548


src/services/assetInfo/assetInfo.ts
```diff
@@ -0,0 +1,75 @@
...
+import { GetOptions, Portfolio } from '../../libs/portfolio'
+
+const scheduledActions: {
+  [network in NetworkId]?: {
+    promise: Promise<any>
+    data: { callback: Function; address: string }[]
+  }
+} = {}
+
+export async function executeBatchedFetch(network: Network): Promise<any> {
```
PetromirDev, line 13: Replace any with a strict type. This applies for both PRs

src/services/assetInfo/assetInfo.test.ts
```diff
@@ -0,0 +1,43 @@
...
+const UNISWAP_ROUTER = '0xE592427A0AEce92De3Edee1F18E0157C05861564'
+
+global.fetch = fetch as any
+
+describe('Asset info service', () => {
+  test.only('Fetches all tokens and NFTS correctly', async () => {
+    jest.spyOn(assetInfo, 'executeBatchedFetch')
+
+    const wethCallback = jest.fn(({ tokenInfo }) =>
+      expect(tokenInfo).toMatchObject({ symbol: 'WETH', decimals: 18 })
```
PetromirDev, line 20: Write a batch test. Similar to the one written for portfolio lib

src/libs/portfolio/portfolio.ts
```diff
@@ -306,7 +306,7 @@ export class Portfolio {
       tokenErrors: tokensWithErrResult
         .filter(([error, result]: [string, TokenResult]) => error !== '0x' || result.symbol === '')
         .map(([error, result]: [string, TokenResult]) => ({ error, address: result.address })),
-      collections: collections.filter((x) => x.collectibles?.length)
+      collections
```
PetromirDev, line 309: Why?

src/services/assetInfo/assetInfo.ts
```diff
@@ -0,0 +1,75 @@
...
+    disableAutoDiscovery: true,
+    previousHints: {
+      erc20s: allAddresses,
+      erc721s: Object.fromEntries(
+        allAddresses.map((i) => [
+          i,
+          {
+            tokens: ['1'],
+            isKnown: false
+          }
```
PetromirDev, line 28: Can you explain this part?

PetromirDev (changes requested):

src/libs/portfolio/portfolio.ts
JIOjosBG: So we get collection info regardless if there is any changes to nft ownership. 
This check is moved to the dashboard component so as we need it only there

src/services/assetInfo/assetInfo.ts
JIOjosBG: sentinel value, used for single nft fetch from the portfolio

PetromirDev (approved):

PR #905, opened 2024-07-25
Refactor / Flow for adding accounts

superKalo:
Refactors the flow for adding accounts, so that:

1. It is moved from the background process to the Main controller and
2. Improves the flow by introducing `handleAddAccounts` (similar to the `handleSignAccountOp` and `handleSignMessage`) instead of listening to controller updates.


superKalo:
No longer relevant.

PR #906, opened 2024-07-25
Feature / Import Default Seed Phrase

sonytooo:
* Implements: https://github.com/AmbireTech/ambire-app/issues/2262

* Moved: getDefaultKeyLabel from ambire-app to common
* Added: addSeed and getSeed functions to the keystore
* Added: newlyAdded flag to the latest accounts. This change is necessary for the Personalization Screen on the FE. With each account having a newlyAdded property, it becomes easier to identify the newly added accounts for all account types on the Personalization Screen
* Added: importSmartAccountFromDefaultSeed function in the main ctrl that combines logic for multiple nested controllers used for creating a new seed phrase wallet or adding a new account from an existing seed phrase wallet


src/controllers/keystore/keystore.ts
```diff
@@ -76,6 +86,8 @@ export class KeystoreController extends EventEmitter {
 
   #storage: Storage
 
+  #keystoreDefaultSeed: string | null = null
+
```
PetromirDev, line 90: Imo we should store the seed in an array in the same way we do for `#keystoreKeys` to be future-proof. Then in `addSeed` we will write:
```
if (#keystoreSeeds.length !== 0)... error state
```

src/controllers/main/main.ts
```diff
@@ -333,6 +345,94 @@ export class MainController extends EventEmitter {
     this.emitUpdate()
   }
 
+  async importSmartAccountFromDefaultSeed(seed?: string) {
+    await this.withStatus('importSmartAccountFromDefaultSeed', async () => {
+      if (this.accountAdder.isInitialized) this.accountAdder.reset()
```
PetromirDev, line 350: `allowConcurrentActions` should be true in main

src/controllers/main/main.ts
```diff
@@ -333,6 +345,94 @@ export class MainController extends EventEmitter {
...
+      if (!nextSmartAccount) {
+        throw new EmittableError({
+          message:
+            'Internal error while looking for account to add. Please start the process all over again and if the issue persists contact Ambire support.',
+          level: 'major',
+          error: new Error('Internal error: Failed to find a smart account to add')
+        })
+      }
+
+      await this.accountAdder.selectAccount(nextSmartAccount)
```
PetromirDev, line 412: not a promise

PetromirDev (changes requested):

src/controllers/keystore/keystore.ts
```diff
@@ -337,6 +350,49 @@ export class KeystoreController extends EventEmitter {
...
+          'keystore: seed phase already added. Storing multiple seed phrases not supported yet'
+        )
+      })
+    }
+
+    // Set up the cipher
+    const counter = new aes.Counter(this.#mainKey!.iv) // TS compiler fails to detect we check for null above
+    const aesCtr = new aes.ModeOfOperation.ctr(this.#mainKey!.key, counter) // TS compiler fails to detect we check for null above
+    this.#keystoreSeeds.push(hexlify(aesCtr.encrypt(new TextEncoder().encode(seed))))
+    await this.#storage.set('keystoreSeeds', this.#keystoreSeeds)
```
superKalo, line 387: @Ivshti , please take a look to see if this looks good to you too ✌️

src/controllers/main/main.ts
```diff
@@ -333,6 +344,98 @@ export class MainController extends EventEmitter {
     this.emitUpdate()
   }
 
+  async importSmartAccountFromDefaultSeed(seed?: string) {
+    await this.withStatus(
+      'importSmartAccountFromDefaultSeed',
```
superKalo, line 349: Minor: you could you use our pattern with `#importSmartAccountFromDefaultSeed`, to avoid 1 level of nesting, but it's fine as is also.

src/controllers/main/main.ts
```diff
@@ -333,6 +344,98 @@ export class MainController extends EventEmitter {
...
+            nextSmartAccount = this.accountAdder.accountsOnPage.find(
+              ({ isLinked, account }) => !isLinked && isSmartAccount(account)
+            )?.account
+
+            if (!nextSmartAccount) break
+
+            isAccountAlreadyAdded = !!this.accounts.accounts.find(
+              // eslint-disable-next-line @typescript-eslint/no-loop-func
+              (a) => a.addr === nextSmartAccount!.addr
+            )
```
superKalo, line 396: Corner case: not checking if this account has keys. The account could have been added as a view only.

Can this happen? It might. If the user copies smart account addresses and adds them in a new instance as a view-only. Might happen I guess.

src/controllers/main/main.ts
```diff
@@ -333,6 +344,98 @@ export class MainController extends EventEmitter {
...
+            error: new Error('failed to retrieve default seed phrase from keystore')
+          })
+        }
+
+        const keyIterator = new KeyIterator(defaultSeed)
+        this.accountAdder.init({
+          keyIterator,
+          hdPathTemplate: BIP44_STANDARD_DERIVATION_TEMPLATE,
+          pageSize: 1
+        })
```
superKalo, line 372: Works just fine, but I had 2 scenarios with importing seeds that caused this error:

Screenshot 2024-08-05 at 17 23 51

and an error that we failed to check if the acc is used on Gnosis network.

Best would be to introduce flags if the AccountAdder should: 1) Search for linked accounts and 2) Should check if account is used on the networks. Both should be false for this use-case. Can be moved in another PR, I'm opening an issue.

superKalo: Ref: https://github.com/AmbireTech/ambire-app/issues/2615

superKalo: Fixed in https://github.com/AmbireTech/ambire-common/pull/922

src/controllers/main/main.ts
superKalo: I'm moving the discussion in https://github.com/AmbireTech/ambire-app/issues/2619

superKalo (approved):

PR #909, opened 2024-07-26
Enhancement/ USD field in transfer

PetromirDev:
Implements the changes required for https://github.com/AmbireTech/ambire-app/pull/2561

### DIFF
--- a/src/controllers/transfer/transfer.ts
+++ b/src/controllers/transfer/transfer.ts
@@ -14,6 +14,12 @@ import { validateSendTransferAddress, validateSendTransferAmount } from '../../s
 import { Contacts } from '../addressBook/addressBook'
 import EventEmitter from '../eventEmitter/eventEmitter'
 
+const CONVERSION_PRECISION = 16
+const CONVERSION_PRECISION_POW = BigInt(10 ** CONVERSION_PRECISION)
+
+/**
+ * Converts floating point token price to big int
+ */
 const convertTokenPriceToBigInt = (
   tokenPrice: number
 ): {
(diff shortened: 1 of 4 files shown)


src/controllers/transfer/transfer.ts
```diff
@@ -141,7 +172,12 @@ export class TransferController extends EventEmitter {
 
     // Validate the amount
     if (this.selectedToken) {
-      validationFormMsgsNew.amount = validateSendTransferAmount(this.amount, this.selectedToken)
+      validationFormMsgsNew.amount = validateSendTransferAmount(
+        this.amount,
+        Number(this.maxAmount),
```
jordan-enev, line 177: Relying on the Number type is risky, as it uses floating-point representation, which can lead to precision issues with numbers having more than 16 decimal places.

src/controllers/transfer/transfer.ts
```diff
@@ -289,6 +366,62 @@ export class TransferController extends EventEmitter {
...
+
+    if (this.amountFieldMode === 'fiat' && this.selectedToken?.decimals) {
+      this.amountInFiat = fieldValue
+
+      // Get the number of decimals
+      const amountInFiatDecimals = fieldValue.split('.')[1]?.length || 0
+      const { tokenPriceBigInt, tokenPriceDecimals } = convertTokenPriceToBigInt(tokenPrice)
+
+      // Convert the numbers to big int
+      const amountInFiatBigInt = parseUnits(fieldValue, amountInFiatDecimals)
```
jordan-enev, line 394: Kolega, that's a really good approach using Strings.

Relying on Strings means fieldValue is a String. Otherwise, if it were a number, we would face floating-point issues.

src/controllers/transfer/transfer.ts
```diff
@@ -289,6 +366,62 @@ export class TransferController extends EventEmitter {
...
+      const precision = 16
+      const pow = BigInt(10 ** precision)
+
+      this.amount = formatUnits(
+        (amountInFiatBigInt * pow) / tokenPriceBigInt,
+        // Don't lose precision
+        amountInFiatDecimals + precision - tokenPriceDecimals
```
jordan-enev, line 401: Actually, it's not very likely, but it's possible to lose precision if the division results in more than 16 decimals, right?

But as I said, it's not something that will happen in 99% of cases.

jordan-enev (approved):
Bravo, Kolega!

Just one thing—please include documentation about the exchange between USD and Token, including information about the decimals. You can also include an example because it's a bit complex, but at the same time, it's very logical.

PR #910, opened 2024-07-29
Fix failing portolio tests

JIOjosBG:
One of the hardcoded accounts we used for testing the portfolio library had to have at least ~5 USDT for the test to pass. 

The onchain USDT balance of this account dropped under 5 USDT, which caused the tests to fail without being updated.

The solution for this is to update the tests to use 1 USDT instead of fueling the account on ethereum.

To prevent confusion in such future occurrences this PR adds `getSafeSendUSDTTransaction`, which checks the balance of the sender before returning the full `{to, value,data}` for USDT transfer

also linting


jordan-enev (approved):
Good job!

PR #913, opened 2024-07-30
Add debug tracecall for nfts

JIOjosBG:
Adds debug_traceCall discovery for nft transfers
image


src/libs/tracer/debugTraceCall.ts
```diff
@@ -71,8 +73,29 @@ export async function debugTraceCall(
...
       {
-        tracer:
-          "{data: [], fault: function (log) {}, step: function (log) { if (log.op.toString() === 'LOG3') { this.data.push([ toHex(log.contract.getAddress()), '0x' + ('0000000000000000000000000000000000000000' + log.stack.peek(4).toString(16)).slice(-40)])}}, result: function () { return this.data }}",
+        tracer: `{
+          data: [], 
+          fault: function (log) {}, 
+          step: function (log) { 
+            if (log.op.toString() === 'LOG3') { 
+              this.data.push(toHex({
+                type:"ERC-20",
```
Ivshti, line 82: we mix quote styles here, i would probably do `erc: 20` instead of `type: "ERC-20"` though. because all types would be ERCs

src/libs/tracer/debugTraceCall.ts
```diff
@@ -85,23 +108,52 @@ export async function debugTraceCall(
...
   }
   const [tokensWithErr] = await deploylessTokens.call(
     'getBalances',
-    [op.accountAddr, foundAddresses],
+    [op.accountAddr, foundTokens],
     opts
   )
 
-  return foundAddresses.filter((addr, i) => tokensWithErr[i].error === '0x')
+  const [nftsWithErr] = await deploylessNfts.call(
```
Ivshti, line 145: we should call these two in parallel, not sequentially

Ivshti (changes requested):

src/controllers/portfolio/portfolio.ts
```diff
@@ -725,6 +732,31 @@ export class PortfolioController extends EventEmitter {
     return true
   }
 
+  async learnNfts(
+    nftsData: [string, bigint[]][] | undefined,
+    networkId: NetworkId
+  ): Promise<boolean> {
+    if (!nftsData || !nftsData?.length) return false
```
jordan-enev, line 739: `!nftsData?.length` should be sufficient

src/libs/tracer/debugTraceCall.ts
```diff
@@ -85,23 +108,52 @@ export async function debugTraceCall(
...
+  ]
+  const foundNftTransfersObject = results
+    .filter((i) => i?.type === 'ERC-721')
+    .reduce((res: { [address: string]: Set<bigint> }, i: any) => {
+      if (!res[i?.address]) res[i?.address] = new Set()
+      res[i.address].add(i.tokenId)
+      return res
+    }, {})
+  const foundNftTransfers: [string, bigint[]][] = Object.entries(foundNftTransfersObject).map(
+    ([k, v]) => [k, Array.from(v).map((i) => BigInt(i))]
```
jordan-enev, line 127: address, id, instead of k, v

src/controllers/main/main.ts
```diff
@@ -426,7 +427,7 @@ export class MainController extends EventEmitter {
     const state = this.accounts.accountStates[accountOp.accountAddr][accountOp.networkId]
     const provider = this.providers.providers[network.id]
     const gasPrice = this.gasPrices[network.id]
-    const addresses = await debugTraceCall(
+    const { tokens, nfts } = await debugTraceCall(
```
jordan-enev, line 430: we don't have any tests for `debugTraceCall` lib nor for the`main->traceCall`, would be great if we write tests for the discovery mechanism

Ivshti (approved):

src/libs/tracer/debugTraceCall.ts
```diff
@@ -51,57 +53,105 @@ export async function debugTraceCall(
...
+                  balance: '0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff'
+                }
+              }
+            : {}
+        }
+      ])
+      .catch((e) => {
+        console.log(e)
+        return [ZeroAddress]
+      })
```
borislav-itskov, line 115: The return statement in `catch` is the old one. Maybe we should make it `return []` and check right after the promise if results are empty:
```
if (!results.length) return {tokens: [], nfts: []}
```

borislav-itskov (changes requested):

borislav-itskov (approved):

jordan-enev (approved):

PR #914, opened 2024-07-31
[DONT MERGE] AmbireAccount7702 added: accepts signatures from address(this)

Ivshti:
## Summary
DOn't panic, we won't make contract changes. This is just an example of how it should be built.


Ivshti:
this is now essentially merged :D

PR #915, opened 2024-08-01
Bug / Prevent triggering to sign an account op more than once

superKalo:
Improves the sign and broadcast flow by:

- Fixed: Prevention for triggering to sign an account op more than once by wrapping `withStatus` the two parts of `handleSignAndBroadcastAccountOp` - 1) Signing and 2) Broadcasting. So that in addition to the loading statuses, this solves a race that may happen when the user clicks (really fast) twice the Sign button (because the button gets disabled on event emit, which might get delayed in some scenarios).
- Fixed: Improve the sign error messaging, so that they sound less asparagus (although, ideally, these situations should never happen).

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -507,7 +507,7 @@ export class MainController extends EventEmitter {
 
         return this.signAccountOp.sign()
       },
-      false
+      true
     )
 
     // Error handling on the prev step will notify the user, it's fine to return here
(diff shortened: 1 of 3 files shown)


src/controllers/main/main.ts
```diff
@@ -490,19 +491,33 @@ export class MainController extends EventEmitter {
...
+            'The signing process was not initialized as expected. Please try again later or contact Ambire support if the issue persists.'
+          const error = new Error('SignAccountOp is not initialized')
+          this.emitError({ level: 'major', message, error })
+          return Promise.reject(error)
+        }
 
-    await this.signAccountOp.sign()
+        return this.signAccountOp.sign()
+      },
+      false
```
superKalo, line 510: Note: these probably should be `true`, please double-check.

sonytooo (approved):

PR #916, opened 2024-08-02
Feature / Notify for Queued Transactions

sonytooo:
Resolves: https://github.com/AmbireTech/ambire-app/issues/2100

* Introduced notificationManager
* removed onSignSuccess callback from main because it was used mainly for creating notifications in the background but now the creation of notifications is available in the main controller too
* Added queued transaction notification on force window close from the system (X) button

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -18,6 +18,7 @@ import {
   TxnRequest
 } from '../../interfaces/keystore'
 import { AddNetworkRequestParams, Network, NetworkId } from '../../interfaces/network'
+import { NotificationManager } from '../../interfaces/notification'
 import { Storage } from '../../interfaces/storage'
 import { Calls, DappUserRequest, SignUserRequest, UserRequest } from '../../interfaces/userRequest'
 import { WindowManager } from '../../interfaces/window'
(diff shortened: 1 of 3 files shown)


src/controllers/main/main.ts
```diff
@@ -1798,8 +1801,10 @@ export class MainController extends EventEmitter {
       },
       actionId
     )
-
-    this.onSignSuccess('account-op')
+    await this.#notificationManager.create({
+      title: 'Successfully signed',
+      message: 'Your transaction was successfully signed and broadcasted to the network'
+    })
```
superKalo, line 1807: We can take the change to update the text (since the title and message repeat themselves). How about:

- title: Done!
- message: Transaction was signed and broadcasted to the network.

src/controllers/main/main.ts
```diff
@@ -493,8 +491,13 @@ export class MainController extends EventEmitter {
...
     await this.resolveUserRequest({ hash: signedMessage.signature }, signedMessage.fromActionId)
-
-    this.onSignSuccess(signedMessage.content.kind === 'typedMessage' ? 'typed-data' : 'message')
+    await this.#notificationManager.create({
+      title: 'Successfully signed',
+      message:
+        signedMessage.content.kind === 'typedMessage'
+          ? 'TypedData was successfully signed'
+          : 'Message was successfully signed'
+    })
```
superKalo, line 500: We can take the chance to update the text (since the title and message repeat themselves). How about simply:

title: Done!
message: The message was signed.

superKalo (approved):

Ivshti (commented):
at whic hpoint to we show this? When you close the signAccountOp window?

sonytooo:
Yes, immediately after the action window closes

PR #921, opened 2024-08-06
Feature/erc 5792

borislav-itskov:
Change log:
* add ERC-5792 handlers

The main change here is that `calls` from dapps can now come in a batch instead of always coming one by one


sonytooo (approved):

Ivshti (approved):

src/controllers/main/main.ts
```diff
@@ -951,18 +954,15 @@ export class MainController extends EventEmitter {
...
         action: {
           kind,
-          calls: [
-            {
-              to: transaction.to,
-              value: transaction.value ? getBigInt(transaction.value) : 0n,
-              data: transaction.data || '0x'
-            }
-          ]
+          calls: calls.map((call) => ({
```
Ivshti, line 961: i suggest implementing a helper for this conversion that will be used when we want to implement deploying smart contracts

borislav-itskov: A similar logic has already started in https://github.com/AmbireTech/ambire-common/pull/935/files so I'll leave it for there

PR #923, opened 2024-08-06
Change/ Benzin action meta swap networkId with chainId

PetromirDev:
Needed for https://github.com/AmbireTech/ambire-app/pull/2620

### DIFF
--- a/src/interfaces/userRequest.ts
+++ b/src/interfaces/userRequest.ts
@@ -1,7 +1,6 @@
 import { SignMessageAction } from 'controllers/actions/actions'
 import { TypedDataDomain, TypedDataField } from 'ethers'
 
-import { Call as AccountOpCall } from '../libs/accountOp/types'
 import { AccountId } from './account'
 import { DappProviderRequest } from './dapp'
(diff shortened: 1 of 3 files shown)


src/interfaces/userRequest.ts
```diff
@@ -47,7 +46,8 @@ export interface SignUserRequest {
   meta: {
     isSignAction: true
     accountAddr: AccountId
-    networkId: NetworkId
+    chainId?: bigint
+    networkId?: NetworkId
```
sonytooo, line 50: do not make the networkId optional. Pass it to the benzin and use only the chainId

src/interfaces/userRequest.ts
```diff
@@ -47,7 +46,8 @@ export interface SignUserRequest {
   meta: {
     isSignAction: true
     accountAddr: AccountId
-    networkId: NetworkId
+    chainId?: bigint
```
sonytooo, line 49: no need to pass chainId?: bigint because [key: string]: any covers it

borislav-itskov (approved):

PR #930, opened 2024-08-07
Config / Migrate the logic for initiating the Account Adder with Lattice1 in the Main controller

superKalo:
Self-explanatory. Brings loading states and race condition prevention, both needed for the extension front-end.

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -686,9 +686,9 @@ export class MainController extends EventEmitter {
 
       await latticeCtrl.unlock(undefined, undefined, true)
 
-      const { sdkSession } = latticeCtrl
+      const { walletSDK } = latticeCtrl
       this.accountAdder.init({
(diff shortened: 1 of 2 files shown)


src/controllers/main/main.ts
```diff
@@ -655,6 +656,46 @@ export class MainController extends EventEmitter {
...
+
+    try {
+      const latticeCtrl = this.#externalSignerControllers.lattice
+      if (!latticeCtrl) {
+        const message =
+          'Could not initialize connection with your Lattice1 device. Please try again later or contact Ambire support.'
+        throw new EmittableError({ message, level: 'major', error: new Error(message) })
+      }
+
+      await latticeCtrl.unlock(undefined, undefined, true)
```
superKalo, line 672: Optional params look a bit ugly like that, would happily refactor sometime in future :v:

PR #932, opened 2024-08-08
Bug/ Account state not updated before estimation

PetromirDev:
## How to test:
- Open two extension instances
- Add the same account in both instances
- Make a transaction from instance (A)
- Immediately try to make a transaction from instance (B)
- Both transactions should be signed with no problems


Ivshti (approved):

PetromirDev:
There may not be an issue and even if there is, this isn't the proper way to fix it.

PR #933, opened 2024-08-08
Add networkId to gasFeePayment in accountOp

JIOjosBG:
We need this info to corretly display the fee token in history

https://github.com/AmbireTech/ambire-app/issues/2628


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -814,6 +814,7 @@ export class SignAccountOpController extends EventEmitter {
         isErc4337Broadcast(this.#network, accountState),
       isGasTank: this.feeTokenResult.flags.onGasTank,
       inToken: this.feeTokenResult.address,
+      networkId: this.feeTokenResult.networkId,
```
borislav-itskov, line 817: `networkId` is not the best name as it's not clear what it indicates. Especially in this case when it's about the fee token. Let's rename it to `feeTokenNetworkId`

borislav-itskov (changes requested):

PR #937, opened 2024-08-09
Enhancement/ Don't ask for entry point authorization if already signed

PetromirDev:
If entry point authorization is needed first check in history if one is already signed. If it is pass the signature to the account op.

Closes https://github.com/AmbireTech/ambire-app/issues/2426

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -1181,8 +1181,6 @@ export class MainController extends EventEmitter {
             a.accountOp.networkId === network.id
         ) as AccountOpAction | undefined
 
-        const hasAuthorized = !!currentAccountOpAction?.accountOp?.meta?.entryPointAuthorization
-
         this.activity.setFilters({
           account: account.addr,
           network: network.id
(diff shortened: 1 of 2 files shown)


Ivshti (approved):

src/controllers/main/main.ts
```diff
@@ -1182,7 +1182,20 @@ export class MainController extends EventEmitter {
...
+          network: network.id
+        })
+        const entryPointAuthorizationMessageFromHistory = this.activity.signedMessages?.items.find(
+          (message) =>
+            message.fromActionId === ENTRY_POINT_AUTHORIZATION_REQUEST_ID &&
+            message.networkId === network.id
+        )
+        if (
+          shouldAskForEntryPointAuthorization(network, account, accountState, hasAuthorized) &&
+          !entryPointAuthorizationMessageFromHistory
```
borislav-itskov, line 1197: We can mark the `hasAuthorized` as true if `entryPointAuthorizationMessageFromHistory` is present and therefore skip the extra `!entryPointAuthorizationMessageFromHistory` check

borislav-itskov (changes requested):
Let's make the small adjustment and merge

PR #939, opened 2024-08-13
Feature: better maxPriorityFeePerGas estimation

borislav-itskov:
Change log:
* make the min for maxPriorityFeePerGas 200n
* make the maxPriorityFeePerGas for each speed at least 12% bigger than the previous one
* when doing RBF, increase the maxPriorityFeePerGas instead of the total gas price (if eip-1559)


Ivshti (approved):

Ivshti (commented):
lgtm i think

PR #944, opened 2024-08-14
Config / Enchance Ledger error handling for corner cases

superKalo:
Tweaks a few things around the Ledger error handling:

- Add: When Ledger throws unexpected error, include a fallback note what user can do to resolve it (because the errors incoming from the device are very vague like: "invalid channel")
- Change: LedgerKeyIterator now uses the LedgerController (instead of directly calling Ledger SDK methods)
- Change: Account Adder error emit now contains the actual error message (instead of a generic one)

### DIFF
--- a/src/interfaces/keyIterator.ts
+++ b/src/interfaces/keyIterator.ts
@@ -1,14 +1,14 @@
 import { HD_PATH_TEMPLATE_TYPE } from '../consts/derivation'
 import { SelectedAccountForImport } from './account'
-import { Key } from './keystore'
+import { ExternalSignerController, Key } from './keystore'
 
 export interface KeyIterator {
   type: Key['type']
   subType?: 'seed' | 'private-key'
   /** The wallet native SDK instance, if any exists */
   walletSDK?: any
-  /** The wallet controller, for the hardware wallets only */
-  controller?: any
+  /** Needed for the hardware wallets only */
+  controller?: ExternalSignerController
   /** Retrieves the the public addresses (accounts) from specific indexes */
   retrieve: (
     fromToArr: { from: number; to: number }[],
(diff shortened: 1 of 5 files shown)


Ivshti (approved):

src/interfaces/keyIterator.ts
```diff
@@ -7,6 +7,8 @@ export interface KeyIterator {
   subType?: 'seed' | 'private-key'
   /** The wallet native SDK instance, if any exists */
   walletSDK?: any
+  /** The wallet controller, for the hardware wallets only */
+  controller?: any
```
PetromirDev, line 11: Why any?

src/libs/ledger/ledger.ts
```diff
@@ -27,5 +27,8 @@ export const normalizeLedgerMessage = (error?: string): string => {
     return 'Rejected by your Ledger device.'
   }
 
-  return error
+  // Indicates a custom timeout error, no need to normalize
+  if (error.includes('Could not connect to your Ledger device for an extended period')) return error
+
+  return `Could not connect to your Ledger device. Please close any other apps that may be accessing your Ledger device (including wallet apps on your computer and web apps). Ensure your Ledger is responsive. Unplug and plug it again. Device error: ${error}`
```
PetromirDev, line 33: What if the device error is very long?

PetromirDev (changes requested):

src/interfaces/keyIterator.ts
superKalo: You're connect, there is a type for this, fixed in https://github.com/AmbireTech/ambire-common/pull/944/commits/d343850ac2ccafd5a05aa3e6ddb6e5bcdc6b1c70

src/libs/ledger/ledger.ts
superKalo: Based on what I've seen, the device errors are relatively short, not descriptive and vague like "Invalid channel" and "Ledger Device is busy". I didn't manage to reproduce a lot longer error, so let's take the risk :v:

PR #946, opened 2024-08-15
no default token to cover fee display error

JIOjosBG:
If the account does not have a token that can cover the fee we need to display an error

### DIFF
--- a/src/controllers/signAccountOp/signAccountOp.ts
+++ b/src/controllers/signAccountOp/signAccountOp.ts
@@ -289,9 +289,9 @@ export class SignAccountOpController extends EventEmitter {
     }
 
     if (
-      this.selectedOption &&
-      this.accountOp.gasFeePayment &&
-      this.selectedOption.availableAmount < this.accountOp.gasFeePayment.amount
+      !this.selectedOption ||
+      (this.accountOp.gasFeePayment &&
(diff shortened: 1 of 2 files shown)


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -258,9 +258,6 @@ export class SignAccountOpController extends EventEmitter {
       errors.push(this.estimation.error.message)
     }
 
-    const availableFeeOptions = this.availableFeeOptions
-    if (!availableFeeOptions.length) errors.push(CRITICAL_ERRORS.eoaInsufficientFunds)
```
Ivshti, line 262: i don't understad why this is removed?

if we went into this condition why wasn't this error displayed? maybe a separate bug?

JIOjosBG: There are two conditionals
- Does the account have any eligible tokens (this one)
- Is a default token present and do we haveenough of it

If you have fee tokens, but not enough to pay the fee, both conditionals will be skipped. The second conditional makes more sense, but after fixing the second one, sometimes both can be triggered. 
we dont really need the first one since the second should always push the error

JIOjosBG: + this is a quick fix and not the whole solution

PR #947, opened 2024-08-16
Add uniReducer

JIOjosBG:
Reduces the elements in the humanization of uniswaps

Wrap swap -> swap
swap unwrap -> swap
swap swap -> swap

resolves https://github.com/AmbireTech/ambire-app/issues/2706


src/libs/humanizer/modules/Uniswap/utils.ts
```diff
@@ -29,3 +29,59 @@ export const joinWithAndLabel = (
...
+
+export const uniReduce = (calls: HumanizerVisualization[][]): HumanizerVisualization[] => {
+  const reduced: HumanizerVisualization[][] = []
+  for (let i = 0; i < calls.length; i++) {
+    if (i === calls.length - 1) {
+      reduced.push(calls[i])
+      // eslint-disable-next-line no-continue
+      continue
+    }
+
```
Ivshti, line 41: all of this is incomprehensible, can you think of a better way to do it? I also think that it can be kind of more generalized

also consider that in some cases the wrap won't be RIGHT before the swap (I can't think of such cases but it's theoretically possible)

Ivshti (changes requested):
consider a more generalized approach, do not assume things are always one after the other

Ivshti (approved):

PR #948, opened 2024-08-16
Fix/ Improve errors state, fix `readyToSign` and more in `signAccountOp`

PetromirDev:
## Changes:
- Add: New errror
- Change: Moved `isAmbireV1AndNetworkNotSupported` from the extension to an error in `errors`
- Fix: `readyToSign` flag

More info in: https://github.com/AmbireTech/ambire-app/pull/2674


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -351,7 +382,13 @@ export class SignAccountOpController extends EventEmitter {
   }
 
   get readyToSign() {
-    return !!this.status && this.status?.type === SigningStatus.ReadyToSign
+    const criticalErrors = this.errors.filter(
+      (error) => !Object.values(NON_CRITICAL_ERRORS).includes(error)
+    )
+
+    return (
+      !!this.status && this.status?.type === SigningStatus.ReadyToSign && !criticalErrors.length
```
Ivshti, line 390: why do we have to check for critical errors here? shouldn't the rest of the code already ensure that `this.status` can't be changed to ReadyToSign if we encounter a critical error?

borislav-itskov: agreed

borislav-itskov (approved):

PR #949, opened 2024-08-16
Feature / Do not derive Ambire smart accounts from private keys

superKalo:
Do not derive Ambire smart accounts from private keys. That's because we can't use the common derivation offset (`SMART_ACCOUNT_SIGNER_KEY_DERIVATION_OFFSET`), and deriving smart accounts out of the private key (with another approach - salt and extra entropy) was creating confusion among the users.

Note: Anyone who has Ambire smart account that was derived from importing a private key can continue to use the account, but is encouraged to migrate to another Ambire smart account. Because re-importing this account on a new extension instance won't be possible.

Closes https://github.com/AmbireTech/ambire-app/issues/2491


src/controllers/accountAdder/accountAdder.ts
```diff
@@ -675,11 +675,20 @@ export class AccountAdderController extends EventEmitter {
...
+        // Since v4.31.0, do not retrieve smart accounts for the private key
+        // type. That's because we can't use the common derivation offset
+        // (SMART_ACCOUNT_SIGNER_KEY_DERIVATION_OFFSET), and deriving smart
+        // accounts out of the private key (with another approach - salt and
+        // extra entropy) was creating confusion.
+        ...(this.#keyIterator.type === 'private-key'
+          ? []
+          : [
+              {
+                from: startIdx + SMART_ACCOUNT_SIGNER_KEY_DERIVATION_OFFSET,
```
PetromirDev, line 688: You can make the whole array a variable and push to it 
```
if (this.#keyIterator.type !== 'private-key')
```

PetromirDev:
An app PR is required because:
"Hide empty basic accounts" should not be displayed
image

Actually, we can skip the account adder step like we do for view-only accounts. It's kind of pointless, isn't?

PetromirDev (changes requested):

superKalo:
Great catch! No, the Account Adder screen is still needed for the linked accounts.

src/controllers/accountAdder/accountAdder.ts
superKalo: Yes, refactored in https://github.com/AmbireTech/ambire-common/pull/949/commits/fc063eb06168d270ac61011c497eeddfd60f8d2b

superKalo:
Great catch! No, the Account Adder screen is still needed for the linked accounts. Changed in https://github.com/AmbireTech/ambire-app/pull/2678

Ivshti (approved):

Ivshti (approved):

Ivshti (approved):

PR #953, opened 2024-08-19
Remove / Benzin Action when a new Action Arrives

sonytooo:
resolves: https://github.com/AmbireTech/ambire-app/issues/2677

* Remove benzin action when a new action that is with priority arrives while the action window is still opened


superKalo (approved):

Ivshti (commented):
lgtm, simple fix

PR #956, opened 2024-08-20
Fix / Potential Issue With Duplicate Accounts

sonytooo:
* Some users are encountering duplicate accounts in their storage, leading to duplicates on the FE. While the cause of these duplicates in storage is still unknown, filtering for unique accounts on AccountsController init ensures this issue no longer occurs.
* When a user adds, edits, or removes an account, the issue of duplicate accounts in storage will be automatically resolved (if such), as we filter for unique addresses with every storage.set

### DIFF
--- a/src/controllers/accounts/accounts.ts
+++ b/src/controllers/accounts/accounts.ts
@@ -212,7 +212,7 @@ export class AccountsController extends EventEmitter {
   }
 
   async removeAccountData(address: Account['addr']) {
-    this.accounts = this.accounts.filter((acc) => acc.addr !== address)
+    this.accounts = getUniqueAccountsArray(this.accounts.filter((acc) => acc.addr !== address))
     if (this.selectedAccount === address) await this.#selectAccount(this.accounts[0]?.addr)
     delete this.accountStates[address]
     this.#storage.set('accounts', this.accounts)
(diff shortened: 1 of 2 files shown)


Ivshti (approved):

src/controllers/accounts/accounts.ts
```diff
@@ -209,14 +212,9 @@ export class AccountsController extends EventEmitter {
   }
 
   async removeAccountData(address: Account['addr']) {
-    this.accounts = this.accounts.filter((acc) => acc.addr !== address)
-
-    if (this.selectedAccount === address) {
-      await this.#selectAccount(this.accounts[0]?.addr)
-    }
-
+    this.accounts = getUniqueAccountsArray(this.accounts.filter((acc) => acc.addr !== address))
```
superKalo, line 215: Upon controller load, `this.accounts` are wrapped in `getUniqueAccountsArray`, therefore are guaranteed to be unique. So probably there is no need to wrap `this.accounts` again here?

src/controllers/accounts/accounts.ts
```diff
@@ -230,14 +228,16 @@ export class AccountsController extends EventEmitter {
...
   async #updateAccountPreferences(accounts: { addr: string; preferences: AccountPreferences }[]) {
-    this.accounts = this.accounts.map((acc) => {
-      const account = accounts.find((a) => a.addr === acc.addr)
-      if (!account) return acc
-      if (isAddress(account.preferences.pfp)) {
-        account.preferences.pfp = getAddress(account.preferences.pfp)
-      }
-      return { ...acc, preferences: account.preferences, newlyAdded: false }
-    })
+    this.accounts = getUniqueAccountsArray(
```
superKalo, line 231: Same as https://github.com/AmbireTech/ambire-common/pull/956/files?diff=split&w=1#r1723843436

superKalo:
Idea: can we ask a user with duplicated accounts if the duplicate accounts have different casing in their addresses (maybe one of them is checksummed and one not)?

PetromirDev:
Same casing. Source: Ivan's local storage

PR #959, opened 2024-08-21
ERC-2470 deployment check

borislav-itskov:
Check if the contract is not already deployed upon singleon interaction.

Screenshot from 2024-08-21 15-33-47

Closes [1794](https://github.com/AmbireTech/ambire-app/issues/1794)


Ivshti (changes requested):
i am more in favor of closing this after a slack discussion on how the problem should actually be solved

borislav-itskov:
Let's not merge this as there is a better way of solving

PR #962, opened 2024-08-21
Fix: relayer networks nonce discrepancy

borislav-itskov:
Fix: nonce discrepancy on relayer networks and SA by making the estimation re-estimation until it finds the correct nonce from the RPC

This is the flow:
* if the estimation finds a higher nonce than the currently set one - reestimate
* if there's a nonce discrepancy (found nonce in the estimation is LOWER than the one in account op) - reestimate. This also is a valid case when an RPC is not up-to-date and it was a bug that caused showing "nonce discrepancy..."
* if there's no nonce discrepancy but the nonce from the estimation and the account op are the same as the last broadcast txn on a relayer network with a smart account, re-estimate again until the next nonce is found

Closes [2705](https://github.com/AmbireTech/ambire-app/issues/2705)

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -1623,7 +1623,13 @@ export class MainController extends EventEmitter {
               network,
               this.accounts.accountStates[localAccountOp.accountAddr][localAccountOp.networkId]
             )
-          }
+          },
+          '0x0000000000000000000000000000000000000001',
+          this.activity.broadcastedButNotConfirmed.filter(
+            (op) => op.networkId === localAccountOp.networkId
+          ).length
+            ? 'pending'
+            : 'latest'
         ).catch((e) => {
           this.emitError({
             level: 'major',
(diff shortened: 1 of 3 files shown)


src/controllers/main/main.ts
```diff
@@ -1623,7 +1623,13 @@ export class MainController extends EventEmitter {
               network,
               this.accounts.accountStates[localAccountOp.accountAddr][localAccountOp.networkId]
             )
-          }
+          },
+          '0x0000000000000000000000000000000000000001',
+          this.activity.broadcastedButNotConfirmed.filter(
+            (op) => op.networkId === localAccountOp.networkId
+          ).length
+            ? 'pending'
```
Ivshti, line 1631: why not just always use pending?

src/controllers/main/main.ts
```diff
@@ -1638,23 +1644,42 @@ export class MainController extends EventEmitter {
...
+        const hasNonceDiscrepancy = estimation.error?.cause === 'NONCE_FAILURE'
+        const lastTxn = this.activity.getLastTxn(localAccountOp.networkId)
+        const SAHasOldNonceOnARelayerNetwork =
+          isSmartAccount(account) &&
+          !network.erc4337.enabled &&
+          lastTxn &&
+          localAccountOp.nonce === lastTxn.nonce
+
+        if (hasNonceDiscrepancy || SAHasOldNonceOnARelayerNetwork) {
+          // wait 1s before re-estimating
```
Ivshti, line 1672: are we waiting for accountState to be updated? can't we await that?

Ivshti (changes requested):

PR #964, opened 2024-08-22
Fix / Dashboard Simulation Cleanup After Deleting the Last Call from an AccountOp

sonytooo:
Resolves: https://github.com/AmbireTech/ambire-app/issues/2682

Fixed: Dashboard simulation cleanup after deleting the last call from an active accountOp of a SA

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -909,6 +909,7 @@ export class MainController extends EventEmitter {
       networkData,
       this.signAccountOp?.accountOp
     )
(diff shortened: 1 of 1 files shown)


PetromirDev (approved):

PetromirDev:
I don't really like the amount of if-else statements. Imo we should try to refactor `removeUserRequest` at some point

jordan-enev (approved):

Ivshti:
@PetromirDev @jordan-enev if I remember correctly the initial version of the mainCtrl had a very clear place after updating the pending to be signed accountOps, where it was supposed to re-trigger portfolio update. Since then we've made it a bit complicated unfortunately.

PR #967, opened 2024-08-26
Fix: identified by tx

borislav-itskov:
Add `identifiedBy` to `SubmittedAccountOp` and refactor the logic around it to make more sense.

What does `identifiedBy` mean?
The `txnId` may not neceseraly be final on the moment of broadcast. It is final when the type is Transaction. This is the case when we do a regular EOA broadcast (including SA EOA broadcast) The relayer and bundler work differently, though. The relayer may sometimes not to return a txnId at all if it decides the current gas prices are too high for the transaction. Also, it may return a `txnId` only to replace it with another if the conditions meet.  
Here is an example:
- you broadcast a transaction on slow and the relayer returns a txnId
- at the same time another person broadcasts via the relayer, fast speed
- the relayer sees that the second txn's chances of getting confirmed sooner are higher and replaces the current one with the RBF logic
- at a later stage, the relayer re-broadcasts the first txn but since it's a different nonce and signature, the txnId also differs

That's why we cannot rely on `txnId` for smart accounts in the relayer broadcast case to fetch information about the transaction. Instead, the relayer will return a database ID record of the transaction and we will be refetching the txnId from the relayer until the transaction gets mined.  
The same logic applies for `userOps` and bundler broadcast. In the case of `userOps`, the only difference is that we get a `userOpHash` instead of a database ID record.

Along with this, here are some noticeable changes:
* `userOpHash` gets deleted from `submittedAccountOp` and instead it exists in `identifiedBy` if the account op is an user op
* `txnId` in `submittedAccountOp` may be undefined
* `submittedAccountOp` gets moved to it's own file and logic around `identifiedBy` is written in that file
* `explorerId` gets deteted as a network option as we don't use it anymore (it was used to help jiffyscan but instead it only made it worse)
* `callRelayer` and `fetch` were made public in main.ts as the `ProviderController` needs them. This sounds like an anti-pattern but the `ProviderController` already breaks the patter "subcontroller shoudn't call other controllers". The ProviderController takes the main controller as a constructor argument and then decides what to do by calling a lot of methods from the main controller. Having already broken this rule, I decided it's OK for fetch and callRelayer to be taken from the main controller as well


borislav-itskov:
@PetromirDev suggested changing the identifiedBy structure to:
```
export type AccountOpIdentifiedBy = {
  type: 'txnId' | 'userOpHash' | 'relayerId'
  identifier: string
}
```
instead of the now that we now currently have:
```
type Transaction = 'txnId'
type UserOpHash = {
  userOpHash: string
}
type Relayer = {
  id: 'string'
}
export type AccountOpIdentifiedBy = Transaction | UserOpHash | Relayer
```

In doing so, we will always have an `identifier` property to use and we do stuff like `${type}=${identifier}`.  
This is more convenient than having a bunch of functions around saying isThisUserOp / isThis RelayerId but I'll highlight the readability benefit of the initial approach compared to this one.  
Let's discuss

sonytooo (approved):

jordan-enev (approved):

PR #975, opened 2024-08-28
Enhancement / toBeLearned tokens before LearnedTokens

gergana95:
Changelog:
- introduce toBeLearned tokens. ToBeLearned tokens is a new temporary mechanism to store tokens and pass to portfolio lib. Populate address with timestamp in learnedTokens once they are in toBeLearned with balance
- always pass learnedTokens to portfolio lib as additionalHints
- Cleaning learnedTokens mechanism changes - remove the ones with timestamp lastly.


Ivshti (changes requested):
I think hintsFromExternalAPI should not be re-assigned, instead there should be a new constant allHints that is also the one being returned, and everyone using the portfolio lib output should use the allHints output.

Otherwise we introduce a naming issue right here, because everyone using hintsFromExternalAPI will actually be using an aggregate of all hints

Also, didn't we purposefully not store anything that is being returned by velcro? What happens with this decision? BEcause not having this means that we'll litter the storage quickly and have to remove tokens when we reach the limit

gergana95:
@Ivshti 
Currently the structure in this PR is wrong and returns all hints (including preferences, pinned and etc.), the idea is to return hints from additionalHints with balance only, in order to store them in previousHints - from velcro and from deployless, so I will update it, good catch.
Do we need to return allHints structure and where would we use it? The lib returns tokens which is all hints as tokens with prices. It seems we will duplicate it this way, but maybe i am missing the point.

Regarding the velcro tokens caching, currently we ARE storing them in previousHints - fromExternalAPI structure. We discussed we will store them there while implementing this [issue](https://github.com/AmbireTech/ambire-app/issues/2109). This way we have a fallback if velcro is not available.

src/controllers/portfolio/portfolio.ts
```diff
@@ -692,6 +700,34 @@ export class PortfolioController extends EventEmitter {
...
+
+    const alreadyLearned = networkToBeLearnedTokens.map((addr) => getAddress(addr))
+    const alreadyLearnedTokens = Object.keys(networkLearnedTokens).map((addr) => getAddress(addr))
+
+    const tokensToLearn = tokenAddresses.filter((address) => {
+      const normalizedAddress = getAddress(address)
+      return (
+        !alreadyLearned.includes(normalizedAddress) &&
+        !alreadyLearnedTokens.includes(normalizedAddress) &&
+        !address === ZeroAddress
```
PetromirDev, line 719: !address is a boolean. `!address === ZeroAddress` compares boolean and string

src/controllers/portfolio/portfolio.ts
```diff
@@ -719,7 +755,7 @@ export class PortfolioController extends EventEmitter {
     // Reached limit
     if (LEARNED_TOKENS_NETWORK_LIMIT - Object.keys(networkLearnedTokens).length < 0) {
       const learnedTokensArray = Object.entries(networkLearnedTokens).sort(
-        (a, b) => Number(a[1]) - Number(b[1])
+        (a, b) => Number(b[1]) - Number(a[1])
```
PetromirDev, line 758: Please add a comment because this `.sort` is hard to read

src/controllers/portfolio/portfolio.ts
```diff
@@ -692,6 +700,34 @@ export class PortfolioController extends EventEmitter {
     this.emitUpdate()
   }
 
+  addTokensToBeLearned(tokenAddresses: string[], networkId: NetworkId) {
+    if (!tokenAddresses) return false
+    if (!this.#toBeLearnedTokens) this.#toBeLearnedTokens = {}
```
PetromirDev, line 705: is this even possible?

src/controllers/portfolio/portfolio.ts
```diff
@@ -692,6 +700,34 @@ export class PortfolioController extends EventEmitter {
     this.emitUpdate()
   }
 
+  addTokensToBeLearned(tokenAddresses: string[], networkId: NetworkId) {
+    if (!tokenAddresses) return false
```
PetromirDev, line 704: `![] === true`. The correct check is `!tokenAddresses.length`

src/controllers/portfolio/portfolio.ts
```diff
@@ -692,6 +700,34 @@ export class PortfolioController extends EventEmitter {
...
+    if (!this.#toBeLearnedTokens) this.#toBeLearnedTokens = {}
+    if (!this.#toBeLearnedTokens[networkId]) this.#toBeLearnedTokens[networkId] = []
+
+    let networkToBeLearnedTokens = this.#toBeLearnedTokens[networkId] || []
```
PetromirDev, line 708: || [] isn't needed because of:
```
if (!this.#toBeLearnedTokens[networkId]) this.#toBeLearnedTokens[networkId] = []
```

src/controllers/portfolio/portfolio.ts
```diff
@@ -692,6 +700,34 @@ export class PortfolioController extends EventEmitter {
...
+        !alreadyLearnedTokens.includes(normalizedAddress) &&
+        !address === ZeroAddress
+      )
+    })
+
+    if (!tokensToLearn.length) return false
+
+    networkToBeLearnedTokens = [...tokensToLearn, ...networkToBeLearnedTokens]
+
+    this.#toBeLearnedTokens[networkId] = networkToBeLearnedTokens
```
PetromirDev, line 727: Imo this logic is mostly unnecessary. I don't see a problem in adding all passed `tokenAddresses` to `toBeLearnedTokens`. After all, hints are deduped in portfolio lib and learnTokens doesn't allow duplicates.

src/libs/portfolio/helpers.ts
```diff
@@ -289,17 +290,23 @@ export function getUpdatedHints(
       const isTokenPreference = lowercaseNetworkPreferenceTokenAddresses.includes(lowercaseAddress)
       const isTokenInExternalAPIHints =
         lowercaseERC20HintsFromExternalAPI.includes(lowercaseAddress)
+      const isTokenInToBeLearnedTokens = tobeLearnedTokens.includes(lowercaseAddress)
```
PetromirDev, line 293: I don't think this check is ok. addresses in `toBeLearnedTokens` are checksummed and `lowercaseAddress` is lowercase

src/libs/portfolio/helpers.ts
```diff
@@ -289,17 +290,23 @@ export function getUpdatedHints(
...
       const hasBalance = lowercaseNetworkTokenAddressesWithBalance.includes(lowercaseAddress)
 
-      if (!isTokenInExternalAPIHints && !isPinned && !isTokenPreference && hasBalance) {
+      if (
+        !isTokenInExternalAPIHints &&
+        isTokenInToBeLearnedTokens &&
+        !isPinned &&
+        !isTokenPreference &&
+        hasBalance
```
PetromirDev, line 301: This condition is only true the first time based on the logic in the controller. We should update the timestamp every time the portfolio is updated and the token has balance.

PetromirDev (changes requested):
Please write a unit test for this logic.

PR #977, opened 2024-08-29
Feature / Add Key Timestamp

sonytooo:
* Store the key timestamp in the meta prop for each key

### DIFF
--- a/src/interfaces/keyIterator.ts
+++ b/src/interfaces/keyIterator.ts
@@ -25,6 +25,8 @@ export interface KeyIterator {
     label: string
     privateKey: string
     dedicatedToOneSA: boolean
-    meta: null
+    meta: {
+      timestamp: number
+    }
   }[]
 }
--- a/src/interfaces/keyIterator.ts
+++ b/src/interfaces/keyIterator.ts
@@ -26,7 +26,7 @@ export interface KeyIterator {
     privateKey: string
     dedicatedToOneSA: boolean
     meta: {
-      timestamp: number
+      createdAt: number
     }
   }[]
 }
--- a/src/interfaces/keystore.ts
+++ b/src/interfaces/keystore.ts
@@ -105,7 +105,6 @@ export type InternalKey = {
   dedicatedToOneSA: boolean
   meta: {
     createdAt: number | null
-    [key: string]: any
   }
 }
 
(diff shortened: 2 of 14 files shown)


src/interfaces/keystore.ts
```diff
@@ -116,6 +119,8 @@ export type ExternalKey = {
     deviceModel: string
     hdPathTemplate: HD_PATH_TEMPLATE_TYPE
     index: number
+    timestamp: number
```
superKalo, line 122: I'd call this `createdAt` (we use `createdAt` and `updatedAt` for the Address book, inspired by the NodeJS conventions).

src/interfaces/keyIterator.ts
```diff
@@ -25,6 +25,8 @@ export interface KeyIterator {
     label: string
     privateKey: string
     dedicatedToOneSA: boolean
-    meta: null
+    meta: {
+      timestamp: number
+    }
```
superKalo, line 30: To be consistent with the old data, that would not have a timestamp, the type should be `number` or `null` maybe?

superKalo, line 30: Also we should discuss a migration that will change meta: null to an object for the current data?

src/interfaces/keystore.ts
```diff
@@ -103,7 +103,10 @@ export type InternalKey = {
   type: 'internal'
   label: string
   dedicatedToOneSA: boolean
-  meta: null
+  meta: {
+    createdAt: number | null
+    [key: string]: any
```
superKalo, line 108: For internal keys, there won't be any other attr than `createdAt`, no?

superKalo: Synced up with Simo and we agreed to remove it https://github.com/AmbireTech/ambire-common/pull/977/commits/3c84c61bbcdecd87df6f31c9df32fa501d6ef5e3

src/interfaces/keystore.ts
```diff
@@ -116,6 +119,8 @@ export type ExternalKey = {
     deviceModel: string
     hdPathTemplate: HD_PATH_TEMPLATE_TYPE
     index: number
+    createdAt: number | null
+    [key: string]: any
```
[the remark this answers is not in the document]
superKalo: Same as https://github.com/AmbireTech/ambire-common/pull/977/files#r1743967187

superKalo (approved):

superKalo (approved):

superKalo:
I'm merging this with one failed test:

FAIL src/libs/portfolio/portfolio.test.ts (17.351 s)
  ● Portfolio › token simulation fails if there are two account ops but the last one has a higher nonce than expected

That's not caused by the changes in here.

PR #979, opened 2024-08-29
Remove async ops from humanizer

JIOjosBG:
## internal optimisization and simplifying

breaking changes:
- remove humanReadable: IrCall[] = [] from controller/signAccountOp.ts
- same for signMessage.ts
- remove   humanizerMetaFragments?: HumanizerFragment[] from libs/accountOp.ts

- remove function humanizeCalls & humanizeTypedMessage
- changed args for humanizeAccountOp
- removed callsHumanizer & messageHumanizer
- removed getTokenInfo
- remove integrateFragments
- remove HumanizerMeta property from Call HumnanizerVisualization
- signAccountOp constructor parameters


src/controllers/signMessage/signMessage.ts
```diff
@@ -105,22 +101,6 @@ export class SignMessageController extends EventEmitter {
         this.dapp = dapp
       }
       this.messageToSign = messageToSign
-      const network = this.#networks.networks.find(
-        (n: Network) => n.id === this.messageToSign?.networkId
-      )
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises
-      messageHumanizer(
-        messageToSign,
-        this.#storage,
```
superKalo, line 114: After these changes, `storage` and `fetch` are no longer needed for the SignMessage controller. Remove them?

src/libs/humanizer/modules/Uniswap/uniUniversalRouter.ts
```diff
@@ -31,26 +31,17 @@ const extractParams = (inputsDetails: any, input: any) => {
...
-    return res as string[]
-  } catch (e) {
-    emitError &&
-      emitError({
-        level: 'minor',
-        message: 'Unexpected error in parsing uniswap commands',
-        error: e
-      })
-    return null
+function parseCommands(commands: string): string[] | null {
```
gergana95, line 34: I am not sure i understand if this works properly, can you leave a comment what it does?

src/libs/humanizer/modules/Uniswap/uniUniversalRouter.ts
```diff
@@ -31,26 +31,17 @@ const extractParams = (inputsDetails: any, input: any) => {
...
-function parseCommands(commands: string, emitError?: Function): string[] | null {
-  try {
-    if (!commands.startsWith('0x') || commands.length % 2 !== 0) return null
-    const hex = commands.slice(2)
-    const hexRegex = /[0-9A-Fa-f]/g
-    if (!hexRegex.test(hex)) return null
-    const res = hex.match(/.{2}/g)?.map((p: string) => `0x${p}`)
-    return res as string[]
-  } catch (e) {
-    emitError &&
```
gergana95, line 43: Wont we have an error anymore?

src/libs/humanizer/utils.ts
```diff
@@ -121,50 +112,8 @@ export async function getNativePrice(network: Network, fetch: Fetch): Promise<nu
...
-  }
-}
-
-export function checkIfUnknownAction(v: Array<HumanizerVisualization>): boolean {
-  try {
-    return !!(v[0].type === 'action' && v?.[0]?.content?.startsWith('Unknown action'))
-  } catch (e) {
-    return false
-  }
+export function checkIfUnknownAction(v: HumanizerVisualization[] | undefined): boolean {
```
gergana95, line 115: isnt it better with the try, catch, like before? Just in case

gergana95 (changes requested):

src/libs/humanizer/humanizerModules.test.ts
```diff
@@ -370,12 +361,7 @@ describe('module tests', () => {
...
       .flat()
     accountOp.calls = allCalls
-    const [irCalls, asyncOps] = humanizeCalls(
-      accountOp,
-      humanizerModules,
-      humanizerInfo,
-      standartOptions
-    )
+    const irCalls = humanizeAccountOp(accountOp, standartOptions)
     // irCalls.forEach((c: IrCall, i) => {
```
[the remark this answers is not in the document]
JIOjosBG: This test is skipped because we do not have the textification functionality anymore. It is not deleted as we will need such test for collision detection for the modules. We should rewrite this in separate PR

superKalo (approved):
I left a few notes, but overall, I tested and based on my discussion with Yossif offline - sounds good to me 👍

I don't feel extremely confident to approve, so I'd love someone to take a look too, because I may not have the full context on all changes.

src/libs/humanizer/modules/Uniswap/uniUniversalRouter.ts
JIOjosBG: the argument we are gonna pass to this function come from, presumably, uniswap txns' calldata. This is the happy case. If the arg is invalid that means the calldata of the original txn is not uniswaps and this module (Uniswap module calls parseCommands) is not the one for this call. 

Happy case = no error
error case = it is supposed to return null anyway

JIOjosBG: since the humanizer is now pure we wont emit errors

src/libs/humanizer/utils.ts
JIOjosBG: not really. the `?` is sufficient to cover all potential edge cases here.
The only questionable thing is if
`v[0].content.startsWith` is not a function, but this is not really possible, because HumanizerVisualization `action` type should always have a `.content: string` property

gergana95 (approved):

PR #980, opened 2024-08-29
Refactor / AccountAdder controller tweaks

superKalo:
Refactors a bit the AccountAdder controller so that:

- Changed: Better derivation labels, now called 1) "BIP44 Standard (MetaMask, Trezor, Grid+)"; 2) "BIP44 Ledger Live" and 3) "Ledger Legacy (MyEtherWallet, MyCrypto)" that I find most sensible as names.
- Changed: The Networks and Providers controllers are now dependencies of the AccountAdder controller, which implies accessing them in the AccountAdder methods.


src/controllers/accountAdder/accountAdder.ts
```diff
@@ -272,21 +284,13 @@ export class AccountAdderController extends EventEmitter {
...
-  }): void {
-    this.hdPathTemplate = path
+  setHDPathTemplate({ hdPathTemplate }: { hdPathTemplate: HD_PATH_TEMPLATE_TYPE }): void {
+    this.hdPathTemplate = hdPathTemplate
     this.page = DEFAULT_PAGE
     this.emitUpdate()
     // get the first page with the new hdPathTemplate (derivation)
-    this.setPage({ page: DEFAULT_PAGE, networks, providers })
+    this.setPage({ page: DEFAULT_PAGE })
+    // TODO: Reset the selected accounts when the HD path is changed?
```
superKalo, line 293: TBD. Maybe the safest option is to reset them.

superKalo: Yes.

PR #984, opened 2024-09-02
signAccount op error handling and warnings improvements

PetromirDev:
## Changes:
- Delete: `NON_CRITICAL_ERRORS` concept in `signAccountOp`
- Change: `overridePendingResults` doesn't have to be async
- Change: Calculate signAccountOp warnings in `update`, instead of `estimateSignAccountOp`. Calculate them in `estimateSignAccountOp` only if there is an error
- Fix: "Unable to estimate the transaction fee in USD" wasn't visualized
- Fix: `emitUpdate` isn't called in `calculateWarnings`
## Screenshots:
image


borislav-itskov (approved):
:clap:

PR #991, opened 2024-09-04
Fix: limit 500kk estimation rule to predefined networks

borislav-itskov:
Fix: limit the 500kk stop estimation rule to predefined networks only as on custom networks like Mantle, it seems it regularly goes above 500kk


Ivshti (commented):
lgtm

PR #994, opened 2024-09-04
Add/metamask contract metadata

JIOjosBG:
follows this issue https://github.com/AmbireTech/ambire-app/issues/2684


scripts/addMetamaskContractMapToHumanizerConsts.js
```diff
@@ -0,0 +1,35 @@
...
+  'humanizerInfo.json'
+)
+fetch('https://raw.githubusercontent.com/MetaMask/contract-metadata/master/contract-map.json')
+  .then((r) => r.json())
+  .then((mmContractMap) => {
+    Object.entries(mmContractMap).forEach(([address, info]) => {
+      const alreadyPresentData = humanizerInfo.knownAddresses[address.toLowerCase()]
+      humanizerInfo.knownAddresses[address.toLowerCase()] = {
+        ...alreadyPresentData,
+        address: address.toLowerCase(),
```
superKalo, line 20: Lowercasing the addresses is intentional, right? A while ago we discussed when to use checksummed addresses vs lower casing them.

scripts/addMetamaskContractMapToHumanizerConsts.js
```diff
@@ -0,0 +1,35 @@
...
+            : alreadyPresentData?.token,
+        isSC: {},
+        logo: info.logo
+          ? `https://raw.githubusercontent.com/MetaMask/contract-metadata/master/images/${info.logo}`
+          : undefined
+      }
+    })
+    return fsPromises.writeFile(humanizerInfoPath, JSON.stringify(humanizerInfo, null, 4), 'utf8')
+  })
+  .then(() => console.log('ready'))
```
superKalo, line 34: Minor, but there might be a case when reading this log is needed. Might not. I'd anyways make it more informative like "Successfully updated the Humanizer (info) JSON with the checksummed Ethereum contract addresses pulled from the @metamask/contract-metadata repo on <timestamp>".

scripts/addMetamaskContractMapToHumanizerConsts.js
```diff
@@ -0,0 +1,35 @@
...
+const humanizerInfoPath = path.join(
+  __dirname,
+  '..',
+  'src',
+  'consts',
+  'humanizer',
+  'humanizerInfo.json'
+)
+fetch('https://raw.githubusercontent.com/MetaMask/contract-metadata/master/contract-map.json')
+  .then((r) => r.json())
```
superKalo, line 14: You might want to have some basic error handling here like:

```
  .then((r) => {
    if (!r.ok) throw new Error(`Failed to fetch. Status: ${r.statusText}`);
    return r.json();
  })
  ```

File: scripts/addMetamaskContractMapToHumanizerConsts.js [hunk 4]
```diff
@@ -0,0 +1,35 @@
...
+  .then((mmContractMap) => {
+    Object.entries(mmContractMap).forEach(([address, info]) => {
```
superKalo, line 16: Might be overkill, sharing it as just an idea. Consider validating the structure of the `mmContractMap` and info fields (such as checking if name, symbol, decimals, etc. exist) to prevent any runtime errors if the data format changes unexpectedly.

package.json
```diff
@@ -3,7 +3,7 @@
   "name": "ambire-common",
   "description": "Common ground for the Ambire apps",
   "scripts": {
-    "generate:humanizerSelectorsAndNames": "node scripts/compileContractSelectorsAndNames.js",
+    "humanizer:addMMList": "node scripts/addMetamaskContractMapToHumanizerConsts.js",
```
superKalo, line 6: Minor, but that's too generic. The file name is also a bit confusing. Since this script is quite specific, I'd add a quite specific name here too like: `updateHumanizerInfoWithMetaMaskContactMetadata.js` and the script name could be `humanizer:update:metamaskContactMetadata` or something similar.

If you want to introduce `humanizer:` prefix to all related scripts, that's fine, but also update all others.

superKalo (changes requested):
Similar scripts, such as the ones that generate or compile resources, are typically run locally during development. Because of this, we tend not to invest too much effort into refining them. However, since this script will be a part of the codebase and will likely be maintained and updated over time, I believe it would benefit from implementing the improvements I mentioned.

scripts/addMetamaskContractMapToHumanizerConsts.js
JIOjosBG: yes, it is intentional as the whole internally works with lowercase addresses. The migration should be in separate PR

scripts/updateHumanizerInfoWithMetaMaskContactMetadata.js
```diff
@@ -0,0 +1,55 @@
...
+  'src',
+  'consts',
+  'humanizer',
+  'humanizerInfo.json'
+)
+
+// fixed for specific release
+const githubSourceBaseUrl = 'https://raw.githubusercontent.com/MetaMask/contract-metadata/v2.5.0'
+
+// Updates Ambire's Humanizer (info) JSON with the Ethereum contract addresses metadata (decimals, symbol, name, logo) pulled from the @metamask/gstacontract-metadata repo.
```
superKalo, line 18: Minor: typo in `@metamask/contract-metadata`

PR #1001, opened 2024-09-10
Refactor: entry point sign message

borislav-itskov:
Move the entry point signature logic to `adjustEntryPointAuthorization`


JIOjosBG:
discussed my comment offline. Since we have only one edge case this is ok

PR #1004, opened 2024-09-13
Enhancement/ `info2` Banner type

PetromirDev:
`info2` is the blue banner

### DIFF
--- a/src/interfaces/banner.ts
+++ b/src/interfaces/banner.ts
@@ -2,7 +2,7 @@ import { AccountOpAction } from 'controllers/actions/actions'
 
 import { Network } from './network'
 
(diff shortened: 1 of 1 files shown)


src/interfaces/banner.ts
```diff
@@ -2,7 +2,7 @@ import { AccountOpAction } from 'controllers/actions/actions'
 
 import { Network } from './network'
 
-export type BannerType = 'error' | 'warning' | 'info' | 'success'
+export type BannerType = 'error' | 'warning' | 'info' | 'info2' | 'success'
```
gergana95, line 5: I would name it infoSecondary or something like this, but its not necessary to change it

PR #1011, opened 2024-09-27
Bug / Transaction (call) can't be deleted from a batch

superKalo:
The `update` method of the SingAccountOp controller was not handling the case when an account op with updated (deleted) calls was passed.

Moreover, this PR refactors the other use-case where the full `accountIOp` got passed (which should not be the case), because this would reset other props that could cause potential bugs).

Resolves https://github.com/AmbireTech/ambire-app/issues/2875

### DIFF
--- a/src/controllers/signAccountOp/signAccountOp.ts
+++ b/src/controllers/signAccountOp/signAccountOp.ts
@@ -397,13 +397,12 @@ export class SignAccountOpController extends EventEmitter {
     speed,
     signingKeyAddr,
     signingKeyType,
-    accountOp,
+    calls,
     gasUsedTooHighAgreed,
     rbfAccountOps,
     bundlerGasPrices,
     blockGasLimit
   }: {
-    accountOp?: AccountOp
     gasPrices?: GasRecommendation[]
     estimation?: EstimateResult | null
     feeToken?: TokenResult
@@ -411,6 +410,7 @@ export class SignAccountOpController extends EventEmitter {
     speed?: FeeSpeed
     signingKeyAddr?: Key['addr']
     signingKeyType?: Key['type']
+    calls?: AccountOp['calls']
     gasUsedTooHighAgreed?: boolean
     rbfAccountOps?: { [key: string]: SubmittedAccountOp | null }
     bundlerGasPrices?: BundlerGasPrice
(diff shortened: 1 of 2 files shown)


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -480,6 +480,12 @@ export class SignAccountOpController extends EventEmitter {
       this.estimation.erc4337GasLimits.gasPrice = bundlerGasPrices
     }
 
+    const shouldUpdateAccountOpCalls =
+      accountOp && accountOp.calls.length !== this.accountOp.calls.length
+    if (shouldUpdateAccountOpCalls) {
```
sonytooo, line 485: The condition should check if (accountOp) as we may want to update other props of the object without affecting the number of calls

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -480,6 +480,12 @@ export class SignAccountOpController extends EventEmitter {
...
+      this.accountOp.calls = accountOp.calls
```
sonytooo, line 486: this.accountOp = accountOp

src/controllers/signAccountOp/signAccountOp.ts
superKalo: Yes, makes sense. I did this in https://github.com/AmbireTech/ambire-common/pull/1011/commits/7ef3abd1a640557fbfe11e8cd3c328c911e3f091

src/controllers/signAccountOp/signAccountOp.ts
superKalo: I used the spread operator, just in case the incoming one has a prop missing in the current one, in https://github.com/AmbireTech/ambire-common/pull/1011/commits/7ef3abd1a640557fbfe11e8cd3c328c911e3f091

borislav-itskov (changes requested):
I don't think allowing updates on the whole accountOp is a good idea. Even if we make it work now, it opens a door for potential bugs in the future.
Currently, doing this sets to null gasFeePayment . Also, it set signingKeyType and signingKeyAddr to whatever the values in accountOp are (could be null). Thirdly, it resets the nonce. Exploring the code carefully:
gasFeePayment  is covered as it's recalculated on each update
signingKeyType  and signingKeyAddr are questionable. They have defaults (on init) but on update they get set again only if they are passed along. If they are not passed along but the accountOp is, they will get the values from the passed accountOp. Which should be either wrong or a bad design as it opens doors for strange bugs to appear
nonce should be okay but only because we've made the sync between accountState, estimation and signAccountOp. I wouldn't be suprised if there is a bug, though
Instead, let's try to pass the properties that are changing and should get reflected in signAccountOp. In this case, the calls have changed. So I propose updating the calls only

superKalo:
I agree and I applied the changes we discussed with Bobby offline ✅ 

Ready for another round of review.

borislav-itskov (approved):

PR #1014, opened 2024-10-01
Bug / Remove duplicates from the Velcro discovery queue

superKalo:
Requesting portfolio latest and pending results duplicates in the Velcro discovery queue. This PR strips the duplicates out before doing the actual Velcro request.


superKalo:
I'm moving this back to draft, because these changes fuck up pending to be confirmed flags:

Screenshot 2024-10-01 at 15 44 01

superKalo:
That's prob no longer an issue or relevant, right @gergana95? I'll just close it, because it's already far off the latest implementation.

PR #1020, opened 2024-10-04
Feature/account export import improvements

borislav-itskov:
Change log:
* rename everywhere "default" seed to "saved" seed
* pass the private key to the FE through a channel, not the state


src/controllers/keystore/keystore.ts
```diff
@@ -605,15 +617,12 @@ export class KeystoreController extends EventEmitter {
...
-    keyAddress: string - the address of the key you want to export
-    publicKey: string - the public key, with which to asymmetrically encypt it (used for key sync with other device's keystoreId)
-  */
-  async exportKeyWithPublicKeyEncryption(
-    keyAddress: string,
-    publicKey: string
-  ): Promise<Encrypted> {
+  async sendPrivateKeyToUi(keyAddress: string) {
+    const decryptedPrivateKey = await this.#getPrivateKey(keyAddress)
+    this.#windowManager.sendWindowUiMessage({ privateKey: `0x${decryptedPrivateKey}` })
```
PetromirDev, line 622: @sonytooo please take a look at this mechanism

src/controllers/keystore/keystore.ts
```diff
@@ -628,7 +637,19 @@ export class KeystoreController extends EventEmitter {
...
     const decryptedBytes = aesCtr.decrypt(encryptedBytes)
-    const decryptedPrivateKey = aes.utils.hex.fromBytes(decryptedBytes)
+    return aes.utils.hex.fromBytes(decryptedBytes)
+  }
+
+  /*
+    DOCS
+    keyAddress: string - the address of the key you want to export
+    publicKey: string - the public key, with which to asymmetrically encypt it (used for key sync with other device's keystoreId)
+  */
```
PetromirDev, line 647: write as jsdoc and fix typo `encypt`

PetromirDev (changes requested):

src/controllers/keystore/keystore.ts
sonytooo: yes, we synced with Bobi and this is exactly how we want it to be. I can't think of a more direct way of sending something from the controllers to the UI

PR #1045, opened 2024-10-28
Refactoring / dApps Session

sonytooo:
* Enhanced the Session class and DappsController to resolve the issue where the extension couldn’t reconnect to open dApps after the service worker reactivates.


src/controllers/dapps/dapps.ts
```diff
@@ -1,3 +1,5 @@
+import { Messenger } from 'interfaces/messenger'
```
superKalo, line 1: Bad import ref.

superKalo (approved):

PR #1046, opened 2024-10-28
Enhancement/ Switch account action

PetromirDev:
Some whitelisted dApps should be able to send requests to all accounts in the extension, unlike most, which only have access to the selected account. This change is needed for https://github.com/AmbireTech/ambire-app/pull/3056


Ivshti (commented):
looks great

PR #1049, opened 2024-10-28
Refactor/ Reduce get account state calls

PetromirDev:
Fetch only the state of newly imported accounts, instead of fetching the account state of all accounts when new accounts are imported. 

## Other changes:
- Fetch account state on account select
- Allow concurrent actions

Closes https://github.com/AmbireTech/ambire-app/issues/2952

### DIFF
--- a/src/controllers/accounts/accounts.ts
+++ b/src/controllers/accounts/accounts.ts
@@ -210,7 +210,11 @@ export class AccountsController extends EventEmitter {
     const defaultSelectedAccount = getDefaultSelectedAccount(accounts)
     if (defaultSelectedAccount) {
       await this.#selectAccount(defaultSelectedAccount.addr)
-      await this.updateAccountState(defaultSelectedAccount.addr)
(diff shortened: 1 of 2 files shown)


borislav-itskov (approved):

src/controllers/accounts/accounts.ts
```diff
@@ -203,10 +210,9 @@ export class AccountsController extends EventEmitter {
     const defaultSelectedAccount = getDefaultSelectedAccount(accounts)
     if (defaultSelectedAccount) {
       await this.#selectAccount(defaultSelectedAccount.addr)
+      await this.updateAccountState(defaultSelectedAccount.addr)
```
superKalo, line 213: This is much better compared to updating all account states, I agree. However, what we discussed in https://github.com/AmbireTech/ambire-app/issues/2952 is that fetching the initial account state should not block the account import process.

Can we enhance (speed up) this even further by triggering, but not awaiting the selected account state update here?

PR #1051, opened 2024-10-29
Fix: relayer error messages

borislav-itskov:
When a relayer error has been encountered, explain it better to the user


Ivshti (commented):
lgtm

PR #1056, opened 2024-10-29
add: methods to store a seed temporarily

borislav-itskov:
Here's what we do:
* add a `#tempSeed` in the keystore
* add a method `addSeedToTemp` in the keystore that will add the seed to temp **encrypted**
* add `deleteTempSeed` - if the user declines to save the seed, we delete it from temp
* add `moveTempSeed` - if the user wants the seed stored, we move it to keystoreSeeds

The flow from account adder is as follows:
* users doesn't have a saved seed
* user imports one, we save it to temp **encryptedly**
* he adds his accounts, personalizes them, goes to final step - do you want to save the imported seed
* if he clicks yes, we store it and remove it from temp
* if he clicks no, we delete it from temp

If the user doesn't complete the login process (rage quits as account personalize), the temp seed remains in the keystore until the user locks his extension or the background process is restarted/reloaded. If the user locks his extension, we delete the temp seed. This is visible in the ambire-app PR (yet to be published)

### DIFF
--- a/src/controllers/keystore/keystore.ts
+++ b/src/controllers/keystore/keystore.ts
@@ -57,6 +57,7 @@ const STATUS_WRAPPED_METHODS = {
   unlockWithSecret: 'INITIAL',
   addSecret: 'INITIAL',
   addSeed: 'INITIAL',
+  moveTempSeedToKeystoreSeeds: 'INITIAL',
   deleteSavedSeed: 'INITIAL',
   removeSecret: 'INITIAL',
   addKeys: 'INITIAL',
@@ -474,7 +475,7 @@ export class KeystoreController extends EventEmitter {
     if (shouldUpdate) this.emitUpdate()
   }
 
-  async addTempSeedToKeystoreSeeds() {
+  async #moveTempSeedToKeystoreSeeds() {
     if (this.#mainKey === null)
       throw new EmittableError({
         message: KEYSTORE_UNEXPECTED_ERROR_MESSAGE,
@@ -507,6 +508,10 @@ export class KeystoreController extends EventEmitter {
     await this.#storage.set('keystoreSeeds', this.#keystoreSeeds)
     this.#tempSeed = null
     this.emitUpdate()
+  }
+
+  async moveTempSeedToKeystoreSeeds() {
+    await this.withStatus('moveTempSeedToKeystoreSeeds', () => this.#moveTempSeedToKeystoreSeeds())
   }
 
   async #addSeed({ seed, hdPathTemplate }: KeystoreSeed) {
--- a/src/controllers/keystore/keystore.ts
+++ b/src/controllers/keystore/keystore.ts
@@ -22,7 +22,7 @@ import {
 import scrypt from 'scrypt-js'
 
 import EmittableError from '../../classes/EmittableError'
-import { HD_PATH_TEMPLATE_TYPE } from '../../consts/derivation'
+import { DERIVATION_OPTIONS, HD_PATH_TEMPLATE_TYPE } from '../../consts/derivation'
 import { Banner } from '../../interfaces/banner'
 import {
   ExternalKey,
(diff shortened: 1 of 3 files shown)


src/controllers/keystore/keystore.ts
```diff
@@ -104,6 +105,12 @@ export class KeystoreController extends EventEmitter {
 
   #keystoreSeeds: KeystoreSeed[] = []
 
+  // when importing a seed, save it temporary here before deciding
+  // whether to place it in #keystoreSeeds or delete it
+  //
+  // this should be done only if there isn't a saved seed already
+  #tempSeed: KeystoreSeed | undefined = undefined
```
superKalo, line 112: Minor: I usually set the empty state to `null` when set on purpose. Undefined is the default (if you access the a prop of the object that doesn't exist). Explicitly setting an empty object property to `null` is generally better because it conveys intentionality.

src/controllers/keystore/keystore.ts
```diff
@@ -439,9 +446,61 @@ export class KeystoreController extends EventEmitter {
 
     // Set up the cipher
     const counter = new aes.Counter(this.#mainKey!.iv) // TS compiler fails to detect we check for null above
-    const aesCtr = new aes.ModeOfOperation.ctr(this.#mainKey!.key, counter) // TS compiler fails to detect we check for null above
+    const aesCtr = new aes.ModeOfOperation.ctr(this.#mainKey!.key, counter) // TS compiler fails to detect we check for null above\
+    return hexlify(aesCtr.encrypt(new TextEncoder().encode(seed)))
+  }
+
+  async addSeedToTemp({ seed, hdPathTemplate }: KeystoreSeed) {
```
superKalo, line 453: I'd try to be a bit more protective here (veggly validating if `seed` and `hdPathTemplate` incoming are defined or are in the expected format), because if for some reason this method gets invoked by a broken front-end, this might brick the KeyStore.

src/controllers/keystore/keystore.ts
```diff
@@ -439,9 +446,61 @@ export class KeystoreController extends EventEmitter {
...
+
+    this.emitUpdate()
+  }
+
+  async deleteTempSeed() {
+    this.#tempSeed = undefined
+    this.emitUpdate()
+  }
+
+  async moveTempSeed() {
```
superKalo, line 467: Minor, but "move" sounds bizarre. I'd name this in a more explicit manner like `addTempSeedToKeystoreSeeds`. But feel free to leave it as it, if that sounds better to you.

src/controllers/keystore/keystore.ts
borislav-itskov: Hey, Kalo, thanks :)  
For the seed we have validation - I moved it to `#getEncryptedSeed`, which is called in this method. It does:
```
if (!Mnemonic.isValidMnemonic(seed)) {
      throw new EmittableError({
        message: 'You are trying to store an invalid seed phrase.',
        level: 'major',
        error: new Error('keystore: trying to add an invalid seed phrase')
      })
    }
```

i only don't know how to validate `hdPathTemplate`. Suggestions?

superKalo: Yep, `DERIVATION_OPTIONS.some((o) => o.value === hdPathTemplate)`

src/controllers/keystore/keystore.ts
```diff
@@ -439,9 +447,71 @@ export class KeystoreController extends EventEmitter {
...
+
+    this.emitUpdate()
+  }
+
+  deleteTempSeed(shouldUpdate = true) {
+    this.#tempSeed = null
+    if (shouldUpdate) this.emitUpdate()
+  }
+
+  async addTempSeedToKeystoreSeeds() {
```
superKalo, line 477: Please add `await this.#initialLoadPromise`, just in case.

superKalo (approved):

PetromirDev (approved):

PR #1057, opened 2024-10-30
Improve / Swap & Bridge Error Handling on the SignAccountOpScreen

sonytooo:
* When swapping or bridging with a BA that requires approval, two transactions are queued: one for the approval and another for the actual swap or bridge. Previously, if the user attempted to select the second transaction before approving the first, the estimation error message was poorly presented. This PR enhances the error message display on the SignAccountOpScreen to improve the UX
Screenshot 2024-10-29 at 18 02 39
Screenshot 2024-10-29 at 18 02 21

* Remove redundant swap or bridge userRequest if the user deletes the approval request while both requests are still pending


borislav-itskov (approved):

superKalo:
👍

PR #1059, opened 2024-10-30
Fixes / Swap & Bridge

sonytooo:
Fixes: https://github.com/AmbireTech/ambire-app/issues/3096

### DIFF
--- a/src/controllers/swapAndBridge/swapAndBridge.ts
+++ b/src/controllers/swapAndBridge/swapAndBridge.ts
@@ -569,7 +569,6 @@ export class SwapAndBridgeController extends EventEmitter {
           isSmartAccount: isSmartAccount(selectedAccount),
           sort: this.routePriority
         })
-
         if (
           this.#getIsFormValidToFetchQuote() &&
           quoteResult &&
@@ -582,12 +581,17 @@ export class SwapAndBridgeController extends EventEmitter {
           let routeToSelectSteps
 
           const selectedRouteInQuoteRes = this.quote
-            ? quoteResult.routes.find(
-                (r: SocketAPIRoute) =>
-                  r.usedBridgeNames[0] === this.quote!.selectedRoute.usedBridgeNames[0] // because we have only routes with unique bridges
-              )
-            : null
+            ? quoteResult.routes.find((r: SocketAPIRoute) => {
+                if (this.quote!.selectedRoute?.usedBridgeNames) {
+                  return r?.usedBridgeNames?.[0] === this.quote!.selectedRoute.usedBridgeNames[0] // because we have only routes with unique bridges
+                }
+                if (this.quote!.selectedRoute?.usedDexName) {
+                  return r?.usedDexName === this.quote!.selectedRoute.usedDexName
+                }
 
+                return false
+              })
+            : null
           if (selectedRouteInQuoteRes) {
             routeToSelect = selectedRouteInQuoteRes
             routeToSelectSteps = getQuoteRouteSteps(selectedRouteInQuoteRes.userTxs)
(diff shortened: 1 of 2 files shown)


superKalo:
Looks good to me, but @borislav-itskov please take a look 🙌

src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -84,15 +85,58 @@ const getActiveRoutesUpdateInterval = (minServiceTime?: number) => {
   return 15000
 }
 
-const buildSwapAndBridgeUserRequests = (
+const buildRevokeApprovalIfNeeded = async (
+  userTx: SocketAPISendTransactionRequest,
+  account: Account,
+  provider: RPCProvider
+): Promise<Call | undefined> => {
+  if (!userTx.approvalData) return
```
borislav-itskov, line 93: This check here might not be correct in the case of basic accounts as the flow is written like this:

1. Check if we should do approve
2. Check if we should do revoke after checking if we should do approve
3. Do the swap

So, we first check `userTx.approvalData` and decide whether we should do an approval. Imagine the token is USDT on Ethereum and the user has already done an infinite approval prior to this request. We get `shouldApprove = false`.  
But then we enter `buildRevokeApprovalIfNeeded` which checks  `userTx.approvalData` which is true, so it continues. It tries to change the approval amount of USDT (which is currently infinite) and the function reverts. We mark `shouldApprove = true`

Imo, we should call `buildRevokeApprovalIfNeeded` only if `shouldApprove` is true as calculated before entering `buildRevokeApprovalIfNeeded`.

src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -130,30 +174,65 @@ const buildSwapAndBridgeUserRequests = (
...
+        userTx.approvalData.approvalTokenAddress,
+        ERC20.abi,
+        provider
+      )
+      const allowance = await erc20Contract.allowance(
+        userTx.approvalData.owner,
+        userTx.approvalData.allowanceTarget
+      )
+      // check if an approval already exists
+      if (BigInt(allowance) === MaxUint256) shouldApprove = false
```
borislav-itskov, line 189: This is good, but it can be better. Let's decide should we pursue such a case.  
Currently, if the amount is infinite, it won't ask for an approval, which is great. But what if I have approved 2000 DAI prior to coming to our swap and decide to swap&bridge 100? It will ask me for an infinite approval.  
We can change this to: 
```
if (BigInt(allowance) >== userTx.approvalData.allowanceTarget) shouldApprove = false
```
This won't force an approval until I put an amount bigger than the allowance

src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -84,15 +85,58 @@ const getActiveRoutesUpdateInterval = (minServiceTime?: number) => {
...
+}
+
+const buildSwapAndBridgeUserRequests = async (
   userTx: SocketAPISendTransactionRequest,
   networkId: string,
-  account: Account
+  account: Account,
+  provider: RPCProvider
 ) => {
   if (isSmartAccount(account)) {
```
borislav-itskov, line 132: Not a bug.  
In the smart account case, I see we don't' check allowance. I think this is better as we're going to force an exact approval, which means the allowance will get reduced to 0 after the transaction succeeds. From a security perspective, this is better.

But I also imagine getting complains "why do I have to approve again, I already have an infinite approval for this token".

Imo, as smart accounts we should force allowances to go to 0. Maybe it will be a good idea to communicate this and see what other people think as well?

borislav-itskov (approved):

PR #1060, opened 2024-10-30
Add: hiddenActivityBanners

borislav-itskov:
Add: hiddenActivityBanners to activity.ts to allow for manually hiding the activity banner

This is the main change:
```
// do not show a banner for forcefully hidden banners
this.broadcastedButNotConfirmed
.filter(
  (op) =>
    !this.#hiddenActivityBanners.includes(
      `${op.accountAddr}-${op.networkId}-${op.timestamp}`
    )
)
...
```


Ivshti (approved):

src/interfaces/banner.ts
```diff
@@ -79,6 +79,11 @@ export type Action =
       actionName: 'close-swap-and-bridge'
       meta: { activeRouteId: number }
     }
+  | {
+      label: 'Close'
+      actionName: 'hide-activity-banner'
+      meta: { timestamp: number; addr: string; network: string; isReject: boolean }
```
PetromirDev, line 85: Three terms are used for the same action:
```
close- in the label ('Close')
reject- in meta (isReject: boolean)
hide- in the action (ACTIVITY_CONTROLLER_HIDE_BANNER)
```
please choose one term and stick to it for clarity or one term for internal use and one for the UI. For example `hide` in the code and `Close` in the UI

PetromirDev (changes requested):

src/controllers/activity/activity.ts
```diff
@@ -506,6 +510,15 @@ export class ActivityController extends EventEmitter {
     this.emitUpdate()
   }
 
+  hideBanner({ addr, network, timestamp }: { addr: string; network: string; timestamp: number }) {
+    this.#hiddenActivityBanners.push(`${addr}-${network}-${timestamp}`)
+
+    /* eslint-disable @typescript-eslint/no-floating-promises */
+    this.#storage.set('hiddenActivityBanners', this.#hiddenActivityBanners)
```
PetromirDev, line 517: How long should these be stored for? Shouldn't there be a cleaning mechanism?

PetromirDev (changes requested):

src/controllers/activity/activity.ts
```diff
@@ -506,6 +510,15 @@ export class ActivityController extends EventEmitter {
...
+
+    this.emitUpdate()
```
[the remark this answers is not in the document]
borislav-itskov: Okay, but I don't understand why should we do that? For example, in `removeAccountData`, we do the same as here: no awaits, first setting the storage and only after that updating

src/controllers/activity/activity.ts
borislav-itskov: Isn't it a bit overkill to implement a cleaning mechanism for this? It's not like you're going to rage hit "close" on every single banner, not to mention that most of the time you will not be able to see it them at all as the txn will finish before actually seeing the activity

borislav-itskov: Decided offline: to change the logic completely

PR #1064, opened 2024-10-31
Bug / Duplicated ETH token in the Swap & Bridge receive list

superKalo:
Caused by a bug in the Socket API (duplicate ETH tokens are incoming, one is with the `ZERO_ADDRESS` and one with `NULL_ADDRESS`, both for ETH). Strip out the one with the `ZERO_ADDRESS` to be consistent with the rest.

PS: Bug reported to the Socket team ✅


superKalo:
Merging with 2 failed tests, but they are unrelated to the changes here.

PR #1065, opened 2024-11-01
Feature/paymaster step

borislav-itskov:
Add:
* a timeout of 8s before declaring the paymaster request a failure (ask the user to retry)
* an `isSignRequestStillActive` to query the application state from in-memory and understand whether the user has rejected the request while waiting for the paymaster response
* another sign status: `WaitingForPaymaster`


src/controllers/main/main.ts
```diff
@@ -1345,6 +1348,20 @@ export class MainController extends EventEmitter {
     this.emitUpdate()
   }
 
+  // a method in the main controller to understand whether
+  // signAccountOp is still initialized
+  isSignRequestStillActive(op: AccountOp): boolean {
+    const accountOpIndex = this.actions.actionsQueue.findIndex(
+      (a) => a.type === 'accountOp' && a.id === `${op.accountAddr}-${op.networkId}`
```
sonytooo, line 1355: this function will work only for SA accountOps. The BA actions of type 'calls' have the same id as their userRequest which in most cases is a number

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -1189,18 +1203,32 @@ export class SignAccountOpController extends EventEmitter {
 
         if (usesPaymaster) {
           try {
-            const response = await this.#callRelayer(
-              `/v2/paymaster/${this.accountOp.networkId}/sign`,
-              'POST',
-              {
+            const santinelTimeoutErr = {}
```
PetromirDev, line 1206: What does this mean `santinelTimeoutErr`

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -1189,18 +1203,32 @@ export class SignAccountOpController extends EventEmitter {
...
+            const paymasterTimeout = new Promise((resolve) => {
+              setTimeout(() => {
+                resolve(santinelTimeoutErr)
+              }, 8000)
```
PetromirDev, line 1210: Wouldn't it be better if the promise is rejected with `Ambire relayer error` instead? This way:
```
   // the ugly error message below gets replaced after
            if (response === santinelTimeoutErr) throw new Error('Ambire relayer error')

            // go back to in progress after paymaster has been confirmed
            this.status = { type: SigningStatus.InProgress }
            this.emitUpdate()
```
will no longer be needed

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -1223,6 +1251,11 @@ export class SignAccountOpController extends EventEmitter {
           }
         }
 
+        // query the application state from memory to understand if the user
+        // hasn't actually rejected the request while waiting for the
+        // paymaster to respond
+        if (!this.#isSignRequestStillActive(this.accountOp)) return
+
```
PetromirDev, line 1258: Note: This is personal opinion:
I don't really like this approach of passing functions down to sub-controllers which check for data of other controllers. Imo this breaks our one-way flow, although it is used in many controllers. Instead, I think you can add a boolean attribute to the controller `isSignRequestStillActive` and update it from main when the account op is rejected- main updates a sub-controller instead of sub-controller calls a function of main.

PetromirDev (changes requested):

PetromirDev (changes requested):

src/controllers/signAccountOp/signAccountOp.ts
borislav-itskov: I don't think it works this way. I tried changing the sign account op state while it was waiting for the paymaster response and it didn't work because on reject, we make signAccountOp null. But signAccountOp remains in memory and it just continues on with its state. That's why I tried querying the application state in this scenario

src/controllers/signAccountOp/signAccountOp.ts
borislav-itskov: This is a timeout error. I don't think it will work as a promise reject

src/controllers/signAccountOp/signAccountOp.ts
borislav-itskov: This is how it was proposed by Ivo in estimate.ts and I just reused it. I agree it can be better

src/controllers/signAccountOp/signAccountOp.ts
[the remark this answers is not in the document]
borislav-itskov: This is not legit as this will only happen if the sign account op screen is no longer visible as the user has rejected the transaction

src/controllers/signAccountOp/signAccountOp.ts
PetromirDev: Check this out https://github.com/AmbireTech/ambire-common/commit/365d334892d45c07bcd95e16d06af347da8e52ad

src/controllers/signAccountOp/signAccountOp.ts
PetromirDev: well, change the state first and then destroy the account op

src/controllers/signAccountOp/signAccountOp.ts
borislav-itskov: Thanks, I will change it

src/controllers/signAccountOp/signAccountOp.ts
borislav-itskov: Refactor however possible

PR #1067, opened 2024-11-01
Higher L2 fees

borislav-itskov:
An attempt to better the UX by increasing the fees a bit on L2s and to overcome the transaction underpriced problem

### DIFF
--- a/src/consts/networks.ts
+++ b/src/consts/networks.ts
@@ -66,8 +66,14 @@ const networks: Network[] = [
       is1559: true,
       elasticityMultiplier: 6n,
       baseFeeMaxChangeDenominator: 50n,
+      // increase the optimism base fee by 50% as it's too low atm
+      // this is to make sure we're getting a bit more than paying
+      // for the transaction
       feeIncrease: 50n,
       userOpFees: {
+        // increase the preVerificationGas by 30% on optimism as more
+        // often than not, the calculated preVerificationGas by the bundler
+        // is not good enough (too small) and the bundler ends up losing money
         preVerificationGas: 30n
       }
     },
--- a/src/consts/networks.ts
+++ b/src/consts/networks.ts
@@ -65,11 +65,7 @@ const networks: Network[] = [
     feeOptions: {
       is1559: true,
       elasticityMultiplier: 6n,
-      baseFeeMaxChangeDenominator: 50n,
-      // increase the optimism base fee by 50% as it's too low atm
-      // this is to make sure we're getting a bit more than paying
-      // for the transaction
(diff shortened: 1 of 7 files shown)


src/services/bundlers/bundler.ts
```diff
@@ -192,20 +192,33 @@ export class Bundler {
...
+    // set bigger gas prices
+    // slow: increase by 5%
+    // medium: increase by 6.67%
+    // fast: increase by 10%
+    // ape: increase by 20%
+    const gasPrices: any = {}
+    Object.keys(results).forEach((key: any, index: number) => {
+      const up = BigInt((4 - index) * 5)
+
+      const finalKey = key !== 'standard' ? key : 'medium'
```
Ivshti, line 209: wouldn't it be a lot more readable if you retain the original code but just use a wrapper function

eg
```
return {
  slow: toBeHex(addExtra(results.max))
  ...
}
```

src/consts/networks.ts
```diff
@@ -65,7 +65,11 @@ const networks: Network[] = [
     feeOptions: {
       is1559: true,
       elasticityMultiplier: 6n,
-      baseFeeMaxChangeDenominator: 50n
+      baseFeeMaxChangeDenominator: 50n,
+      feeIncrease: 50n,
```
Ivshti, line 69: what unit is this in?

src/consts/networks.ts
```diff
@@ -117,7 +121,8 @@ const networks: Network[] = [
     features: [],
     feeOptions: {
       is1559: true,
-      minBaseFee: 100000000n // 1 gwei
+      minBaseFee: 100000000n, // 1 gwei
```
Ivshti, line 124: can we add comments to explain why those increases are required

src/interfaces/network.ts
```diff
@@ -14,6 +14,10 @@ interface FeeOptions {
   // transactions on Base get stuck on slow as we lower the baseFee a lot
   // so we make the minBaseFee the same as the last block one
   minBaseFeeEqualToLastBlock?: boolean
+  userOpFees?: {
+    // all the fees are in percentages
```
Ivshti, line 18: why is this required in FeeOptions? is it different for each fee option or fixed?

also, using promilles or basis points is more appropriate with BigInts since it allows for more precision

Ivshti (changes requested):

src/interfaces/network.ts
borislav-itskov: This is removed as discussed offline

Ivshti (approved):

PR #1073, opened 2024-11-05
min 1 gwei for gasPrice

borislav-itskov:
Closes https://github.com/AmbireTech/ambire-app/issues/3179

The problem with SEI network was that sometimes there were no transactions in the last block, making our calculation 0 for gas price


Ivshti (changes requested):
discussing on slack but this looks to be redundant

Ivshti (approved):

PR #1074, opened 2024-11-06
use eth_gasPrice

borislav-itskov:
Call eth_gasPrice to check for a minimum to compare for gas prices on non-1559 networks

### DIFF
--- a/src/libs/gasPrice/gasPrice.ts
+++ b/src/libs/gasPrice/gasPrice.ts
@@ -137,7 +137,11 @@ export async function getGasPriceRecommendations(
 ): Promise<{ gasPrice: GasRecommendation[]; blockGasLimit: bigint }> {
   const [lastBlock, ethGasPrice] = await Promise.all([
     refetchBlock(provider, blockTag),
-    (provider as JsonRpcProvider).send('eth_gasPrice', []).catch(() => '0x00')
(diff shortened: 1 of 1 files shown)


src/libs/gasPrice/gasPrice.ts
```diff
@@ -198,11 +201,16 @@ export async function getGasPriceRecommendations(
     return { gasPrice: fee, blockGasLimit: lastBlock.gasLimit }
   }
   const prices = filterOutliers(txns.map((x) => x.gasPrice!).filter((x) => x > 0))
+
+  // use th fetched price as a min if not 0 as it could be actually lower
+  // than the hardcoded MIN.
+  const minOrFetchedGasPrice = BigInt(ethGasPrice) !== 0n ? BigInt(ethGasPrice) : MIN_GAS_PRICE
```
Ivshti, line 207: what if eth_gasPrice returns 0, shouldn't we actually respect it? 

Also, is eth_gasPrice returning errors on networks that use 1559? If not, then let's not swallow the error and always log it, because it could lead to a network mysteriously not working without us knowing

if it returns 0 and we want to respect it, we need to distinct between a zero (0x00...000) and an empty response (0x)

Ivshti (changes requested):

Ivshti (approved):

PR #1075, opened 2024-11-06
Bug / Continuous "Failed to fetch routes for this pair" error thrown when updating the quote for swapping (not bridging) tokens

superKalo:
Right after the swap quote gets requested, in 1 minute, the continuous update throws JS error, causing "Failed to fetch routes for this pair" throw. That's because the swap quote response lacks `usedBridgeNames` (which is bridge-only return param).

Closes https://github.com/AmbireTech/ambire-app/issues/3177 (technically, https://github.com/AmbireTech/ambire-common/pull/1059 got merged before this PR that closed this beforehand).


src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -580,21 +585,23 @@ export class SwapAndBridgeController extends EventEmitter {
...
+            if (!this.quote) return false
+
+            // Because we have only routes with unique bridges (bridging case)
+            const selectedRouteUsedBridge = this.quote.selectedRoute.usedBridgeNames?.[0]
+            if (selectedRouteUsedBridge)
+              return nextRoute.usedBridgeNames?.[0] === selectedRouteUsedBridge
+
+            // Assuming to have routes with unique DEXes ONLY (swapping case)
+            const selectedRouteUsedDex = this.quote.selectedRoute.usedDexName
+            if (selectedRouteUsedDex) return nextRoute.usedDexName === selectedRouteUsedDex
```
superKalo, line 598: TODO: I'll double check this with Socket, because that's not always the case, see https://github.com/AmbireTech/ambire-app/issues/3173

superKalo: Yep, all good. Should be unique per quote. Example on Arbitrum:

Screenshot 2024-11-07 at 11 53 13

sonytooo (approved):

PR #1076, opened 2024-11-06
Bug / Update Swap & Bridge from token with the latest from portfolio tokens

superKalo:
Update Swap & Bridge from token with the latest from portfolio tokens to handle:

- Fixed: Updates from token when the selected account gets changed (closes https://github.com/AmbireTech/ambire-app/issues/3025)
- Fixed: Updates from token amount if the currently selected account receives or sends the token in the meantime.

### DIFF
--- a/src/controllers/swapAndBridge/swapAndBridge.ts
+++ b/src/controllers/swapAndBridge/swapAndBridge.ts
@@ -411,11 +411,23 @@ export class SwapAndBridgeController extends EventEmitter {
     if (shouldEmit) this.emitUpdate()
   }
 
-  updatePortfolioTokenList(portfolioTokenList: TokenResult[]) {
-    this.portfolioTokenList = portfolioTokenList
+  updatePortfolioTokenList(nextPortfolioTokenList: TokenResult[]) {
+    this.portfolioTokenList = nextPortfolioTokenList
 
-    if (!this.fromSelectedToken) {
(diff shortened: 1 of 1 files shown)


src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -411,11 +411,23 @@ export class SwapAndBridgeController extends EventEmitter {
...
 
-  updatePortfolioTokenList(portfolioTokenList: TokenResult[]) {
-    this.portfolioTokenList = portfolioTokenList
+  updatePortfolioTokenList(nextPortfolioTokenList: TokenResult[]) {
+    this.portfolioTokenList = nextPortfolioTokenList
 
-    if (!this.fromSelectedToken) {
-      this.updateForm({ fromSelectedToken: this.portfolioTokenList[0] || null })
+    const fromSelectedTokenInNextPortfolio = nextPortfolioTokenList.find(
+      (t) => t.address === this.fromSelectedToken?.address
```
sonytooo, line 418: Search by address and network combined

superKalo: Thanks, I just spot that. Fixed in https://github.com/AmbireTech/ambire-common/pull/1076/commits/1bf584a2f2dbfc168943d30ba9f7b085dd8bca07

PR #1077, opened 2024-11-06
Proposal: change state override detection

Oxbobby:
Dedicate a special function for checking state override that tries to do an `estimateEOA` - if it fails, we mark the chain as unsupported for state override as it doesn't do it properly.

Problem: Moonbeam doesn't work.  
Moonbeam doesn't work as it supports state override but not fully. Simple state override like putting code to an EVM address works, but putting the Ambire code on an EOA and calling execute() on it doesn't. So both simulations + estimation doesn't work. And the worst of all is that the EOA estimation gets bricked and the user cannot broadcast.


Oxbobby:
this one just got too old, we will have to reevaluate its purpose

PR #1081, opened 2024-11-08
Feature / SelectedAccountController (PART 1)

sonytooo:
* Introduced a new ctrl called: SelectedAccountController 
* Init SelectedAccountController into the MainController
* selectAccount logic moved from the AccountsController to the MainController

**Instead of accessing the account address via mainCtrl.accounts.selectedAccount, you can now directly access it through mainCtrl.selectedAccount.account.addr and other respective properties within mainCtrl.selectedAccount.account**


src/controllers/accounts/accounts.ts
```diff
@@ -207,22 +183,15 @@ export class AccountsController extends EventEmitter {
...
-    }
+    this.#onAddAccounts(accounts)
 
     this.emitUpdate()
   }
 
   async removeAccountData(address: Account['addr']) {
     this.accounts = this.accounts.filter((acc) => acc.addr !== address)
-    if (this.selectedAccount === address) await this.#selectAccount(this.accounts[0]?.addr)
+    this.#onRemoveAccount(address)
```
PetromirDev, line 193: This is not okay. Move the logic to `main.removeAccount`

PetromirDev (changes requested):

PR #1083, opened 2024-11-08
Fix: infinite estimation

borislav-itskov:
Fix: if the last txn on a network fails and the nonce does not move, we enter infinite estimation; this is fixed by making sure we're comparing the nonce to a previusly successful transaction

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -1829,7 +1829,9 @@ export class MainController extends EventEmitter {
           isSmartAccount(account) &&
           !network.erc4337.enabled &&
           lastTxn &&
-          localAccountOp.nonce === lastTxn.nonce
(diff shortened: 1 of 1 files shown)


lattaai4:
Hello, I tried to fix the issue

This is what I did:

Modify the estimate function to compare the current nonce with the last successful transaction nonce to prevent infinite estimation loops.

I also created pull request: https://github.com/AmbireTech/ambire-common/pull/1084


> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

PR #1089, opened 2024-11-12
Extend / Selected Account Ctrl

sonytooo:
* Implemented selectedAccount.portfolio
* Implemented selectedAccount.defiPositions
* swapAndBridge - get portfolio tokens from selectedAccount instead of relying on the FE to update them
* removed portfolioView and moved the logic to libs/selectedAccount
* SelectedAccountPortfolio interface imrovements
* Throttled the selectedAccount portfolio updates for a better performance


src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -69,10 +135,130 @@ export class SelectedAccountController extends EventEmitter {
...
+
+    this.#updatePortfolioThrottle.time = now
+
+    const defiPositionsAccountState = this.#defiPositions.state[this.account.addr]
+
+    const updatedPortfolioState = getSelectedAccountPortfolio(
+      structuredClone({
+        latest: this.#portfolio.latest,
+        pending: this.#portfolio.pending
+      }),
```
PetromirDev, line 161: Why do this here? Isn't it better to pass the original object to the function and clone it in the function body?
So, instead of spreading large objects:
```
  const updatedPortfolioState = {
    latest: {
      ...portfolioState.latest,
      [account.addr]: portfolioLatestAccountState
    },
    pending: {
      ...portfolioState.pending,
      [account.addr]: portfolioPendingAccountState
    }
  }
```
we can do:
```
  portfolioState.latest[account.addr] = portfolioLatestAccountState
  portfolioState.pending[account.addr] = portfolioPendingAccountState

return portfolioState
```

src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -69,10 +135,130 @@ export class SelectedAccountController extends EventEmitter {
...
+      this.portfolioStartedLoadingAtTimestamp = Date.now()
+    }
+
+    if (
+      newSelectedAccountPortfolio.isAllReady ||
+      (!this.portfolio?.tokens?.length && newSelectedAccountPortfolio.tokens.length)
+    ) {
+      this.portfolio = newSelectedAccountPortfolio
+    } else {
+      this.portfolio.isAllReady = newSelectedAccountPortfolio.isAllReady
```
PetromirDev, line 189: ```this.portfolio.isAllReady = false```
to simplify it

src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -69,10 +135,130 @@ export class SelectedAccountController extends EventEmitter {
...
+    if (!this.#portfolio || !this.#defiPositions || !this.account) return
+
+    const now = Date.now()
+    const timeSinceLastCall = now - this.#updatePortfolioThrottle.time
+    if (timeSinceLastCall <= 150) {
+      if (!this.#updatePortfolioThrottle.throttled) {
+        this.#updatePortfolioThrottle.throttled = true
+        await wait(150 - timeSinceLastCall)
+        this.#updatePortfolioThrottle.throttled = false
+        await this.#updateSelectedAccountPortfolio()
```
PetromirDev, line 148: There is no use-case for this yet but why not pass skipUpdate to `await this.#updateSelectedAccountPortfolio()`. It'd get lost otherwise

src/libs/selectedAccount/selectedAccount.ts
```diff
@@ -1,65 +1,144 @@
...
+          networkBalance += posByProv.positionInUSD || 0
+        })
+      })
+
+      // eslint-disable-next-line no-param-reassign
+      portfolioAccountState[networkId]!.result!.total.usd = networkBalance
+      // eslint-disable-next-line no-param-reassign
+      portfolioAccountState[networkId]!.result!.tokens = tokens
+    }
+  })
```
PetromirDev, line 73: Nesting can be reduced with 2 guard clauses:
```
 if (!networkState?.result) return
    let tokens = networkState.result.tokens || []
    let networkBalance = networkState.result.total?.usd || 0

    const positions = defiPositionsAccountState[networkId] || {}

    positions.positionsByProvider?.forEach((posByProv: PositionsByProvider) => {
      posByProv.positions.forEach((pos) => {
        pos.assets.forEach((a) => {
          const tokenInPortfolioIndex = tokens.findIndex((t) => {
            return getAddress(t.address) === getAddress(a.address) && t.networkId === networkId
          })

          if (tokenInPortfolioIndex === -1) return

          const tokenInPortfolio = tokens[tokenInPortfolioIndex]
          const priceUSD = tokenInPortfolio.priceIn.find(
            ({ baseCurrency }: { baseCurrency: string }) => baseCurrency.toLowerCase() === 'usd'
          )?.price
          const tokenBalanceUSD = priceUSD
            ? Number(
                safeTokenAmountAndNumberMultiplication(
                  BigInt(tokenInPortfolio.amount),
                  tokenInPortfolio.decimals,
                  priceUSD
                )
              )
            : undefined

          networkBalance -= tokenBalanceUSD || 0 // deduct portfolio token balance

          // eslint-disable-next-line no-param-reassign
          tokens = tokens.filter((_, index) => index !== tokenInPortfolioIndex)
        })

        networkBalance += posByProv.positionInUSD || 0
      })
    })
```

src/libs/selectedAccount/selectedAccount.ts
```diff
@@ -1,65 +1,144 @@
...
+                      BigInt(tokenInPortfolio.amount),
+                      tokenInPortfolio.decimals,
+                      priceUSD
+                    )
+                  )
+                : undefined
+
+              networkBalance -= tokenBalanceUSD || 0 // deduct portfolio token balance
+
+              // eslint-disable-next-line no-param-reassign
```
PetromirDev, line 59: this comment isn't needed

PetromirDev (changes requested):

src/libs/selectedAccount/selectedAccount.ts
PetromirDev: I mean // eslint-disable-next-line no-param-reassign

src/controllers/selectedAccount/selectedAccount.ts
sonytooo: we should not modify the original state. I tried this pattern portfolioState.latest[account.addr] but for some reason assigning it this way to a structuredClone object was not working in the tests and this is why I ended up spreading it

PR #1095, opened 2024-11-15
Optimization / Make DefiPositionsController State Private in MainController

sonytooo:
* made defiPositions private in the MainController
* impl defiPositionsBanners in selectedAccount
* removed defiPositions.banners because these banners were specific for the selectedAccount only

**By keeping the defiPositions state private and not sending it to the FE, we ensure optimal performance even with multiple heavy accounts added to the wallet**

### DIFF
--- a/src/controllers/selectedAccount/selectedAccount.ts
+++ b/src/controllers/selectedAccount/selectedAccount.ts
@@ -1,6 +1,9 @@
 import { Account } from '../../interfaces/account'
+import { Banner } from '../../interfaces/banner'
 import { SelectedAccountPortfolio } from '../../interfaces/selectedAccount'
 import { Storage } from '../../interfaces/storage'
+// eslint-disable-next-line import/no-cycle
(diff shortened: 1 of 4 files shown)


src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -110,6 +128,9 @@ export class SelectedAccountController extends EventEmitter {
       this.#debounceFunctionCallsOnSameTick('updateSelectedAccountDefiPositions', () =>
         this.#updateSelectedAccountDefiPositions()
       )
+      this.#debounceFunctionCallsOnSameTick('updateDefiPositionsBanners', () =>
+        this.#updateDefiPositionsBanners()
+      )
```
PetromirDev, line 133: Call it in `#updateSelectedAccountDefiPositions` as it only depends on them

PetromirDev (changes requested):

PR #1096, opened 2024-11-15
Fixes / Portfolio Update Loading State

sonytooo:
* prevent the Dashboard from going into a loading state when the portfolio is automatically updated in the background


src/controllers/main/main.ts
```diff
@@ -902,6 +905,7 @@ export class MainController extends EventEmitter {
 
     const isUpdatingAccount = this.accounts.statuses.updateAccountState !== 'INITIAL'
 
+    this.selectedAccount.portfolio = DEFAULT_SELECTED_ACCOUNT_PORTFOLIO
```
PetromirDev, line 908: It'd be much better if the controller had a method called `resetSelectedAccountPortfolio`

PetromirDev (changes requested):

PR #1098, opened 2024-11-18
Optimize / Portfolio

sonytooo:
*

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -1,7 +1,6 @@
 import { ethErrors } from 'eth-rpc-errors'
 /* eslint-disable @typescript-eslint/brace-style */
 import { getAddress, getBigInt, Interface, isAddress } from 'ethers'
-import { SocketAPISendTransactionRequest } from 'interfaces/swapAndBridge'
 
 import AmbireAccount from '../../../contracts/compiled/AmbireAccount.json'
 import AmbireFactory from '../../../contracts/compiled/AmbireFactory.json'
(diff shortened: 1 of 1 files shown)


src/controllers/portfolio/portfolio.ts
```diff
@@ -178,14 +178,19 @@ export class PortfolioController extends EventEmitter {
...
   }
 
   #prepareState(selectedAccount: Account, key: 'latest' | 'pending') {
-    const state = this[key]
+    const stateMap = {
+      latest: this.#latest,
+      pending: this.#pending
+    }
+    const state = stateMap[key]
+
```
PetromirDev, line 193: What's the idea behind this change?

src/controllers/main/main.ts
```diff
@@ -1234,6 +1240,38 @@ export class MainController extends EventEmitter {
...
+
+    if (!addrVestingData) return
+    const userRequest: UserRequest = buildMintVestingRequest({
+      selectedAccount: this.selectedAccount.account.addr,
+      selectedToken: token,
+      addrVestingData
+    })
+
+    this.addUserRequest(userRequest)
+  }
```
PetromirDev, line 1273: nice

PetromirDev (changes requested):

src/controllers/portfolio/portfolio.ts
```diff
@@ -71,13 +71,7 @@ export class PortfolioController extends EventEmitter {
 
   validTokens: any = { erc20: {}, erc721: {} }
```
superKalo, line 72: What's the purpose of "any" here? This could get extracted in a separate type that can be reused for the extension too?

```
type ValidTokens = { erc20: { [key: string]: boolean }; erc721: { [key: string]: boolean } }
```

and then here:

```
validTokens?: ValidTokens = { erc20: {}, erc721: {} }
```

superKalo: PS: I know that's not part of your changes, but since it's near I mentioned it :)

PR #1109, opened 2024-11-21
Bug / Swap & Bridge controller update portfolio token list mechanism glitch

superKalo:
TBD.


superKalo:
This is already optimized in https://github.com/AmbireTech/ambire-common/pull/1103

E2E Test fails because of a trivial reason - the CI doesn't have a valid Socket API key and the Swap & Bridge controller fires an error toast that prevents the CI from being able to directly click on the "Send" button in the Dashboard.

PR #1112, opened 2024-11-22
Fix / Resolving of Batched Routes

sonytooo:
* Fixed: https://github.com/AmbireTech/ambire-app/issues/3272
* Added network name to the bridge banners on the Dashboard

Screenshot 2024-11-22 at 16 02 56

Screenshot 2024-11-25 at 10 56 04

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -1626,13 +1626,13 @@ export class MainController extends EventEmitter {
     )
 
     // Update route status immediately, so that the UI quickly reflects the change
-    await Promise.all(
-      swapAndBridgeUserRequests.map(async (r) => {
-        await this.swapAndBridge.updateActiveRoute(r.meta.activeRouteId, {
-          routeStatus: 'in-progress'
-        })
+    // eslint-disable-next-line no-restricted-syntax
+    for (const r of swapAndBridgeUserRequests) {
+      // eslint-disable-next-line no-await-in-loop
+      await this.swapAndBridge.updateActiveRoute(r.meta.activeRouteId, {
+        routeStatus: 'in-progress'
       })
-    )
+    }
 
     // Note: this may take a while!
     const txnId = await pollTxnId(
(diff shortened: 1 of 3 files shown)


src/controllers/main/main.ts
```diff
@@ -1626,13 +1626,13 @@ export class MainController extends EventEmitter {
...
-          routeStatus: 'in-progress'
-        })
+    // eslint-disable-next-line no-restricted-syntax
+    for (const r of swapAndBridgeUserRequests) {
+      // eslint-disable-next-line no-await-in-loop
+      await this.swapAndBridge.updateActiveRoute(r.meta.activeRouteId, {
+        routeStatus: 'in-progress'
       })
-    )
+    }
```
superKalo, line 1635: What was the problem with Promise.all-ing these?

src/libs/banners/banners.ts
```diff
@@ -32,13 +32,24 @@ const getBridgeActionText = (routeStatus: ActiveRoute['routeStatus'], isBridgeTx
...
+  let assetsText = `${fromAssetSymbol} to ${toAssetSymbol}`
+
+  if (networks) {
+    const fromAssetNetwork = networks.find((n) => Number(n.chainId) === steps[0].fromAsset.chainId)
+    const toAssetNetwork = networks.find(
+      (n) => Number(n.chainId) === steps[steps.length - 1].toAsset.chainId
+    )
+    if (fromAssetNetwork && toAssetNetwork) {
+      assetsText = `${fromAssetSymbol} on (${fromAssetNetwork.name}) to ${toAssetSymbol} on (${toAssetNetwork.name})`
+    }
```
superKalo, line 50: Maybe better to either: 1) move "on" inside the network name in parentheses? or 2) Remove the parentheses.

> Bridged USDC on (Optimism) to CBBTC on (Base) (step 2 of 2)

superKalo (approved):

PetromirDev (changes requested):
:heavy_plus_sign: Same as Kalo's feedback

src/controllers/main/main.ts
sonytooo: The issue occurs because updateActiveRoute writes to this.activeRoutes. When multiple routes are updated in parallel, it causes a race condition, resulting in some routes failing to update properly

PR #1121, opened 2024-11-25
Feature/Activity Filtration by session

jordan-enev:
- Change: Updated ActivityController to filter and paginate AccountsOps and SignedMessages by sessionId. Now, different pages can apply independent filters, with their activity tied to the specific page context. This was necessary because AccountsOps will be displayed and filtered on both Settings -> Transaction History and Dashboard -> Activity tabs. Previously, applying a filter affected both places, leading to poor UX.
- Change: Streamlined the way we search for `entryPointAuthorizationMessageFromHistory` in MainController. Previously, initializing the filters was cumbersome; now, there's a simple helper function in ActivityController. @sonytooo, we need to discuss the hasAuthorized logic. The current implementation keeps the last 1000 signed messages in history, and I'm not sure if this is adequately accounted for.
- Change: Documentation updated, regarding how ActivityController is working.


src/controllers/activity/activity.ts
```diff
@@ -171,95 +178,108 @@ export class ActivityController extends EventEmitter {
...
   }
 
-  reset() {
-    this.filters = null
-    this.isInitialized = false
-    this.emitUpdate()
+  // Reset filtered AccountsOps session.
+  // Example: when a FE component is being unmounted, we don't need anymore the filtered accounts ops and we
+  // free the memory calling this method.
+  resetAccountsOpsFilters(sessionId: string) {
```
PetromirDev, line 220: make method documentation accessible on hover /** */ instead of //

PetromirDev (changes requested):

PR #1128, opened 2024-11-27
An old account banner warning

borislav-itskov:
Screenshot from 2024-11-27 09-59-44


src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -317,10 +321,40 @@ export class SelectedAccountController extends EventEmitter {
     }
   }
 
+  get oldAccountBanner(): Banner[] {
```
superKalo, line 324: Minor: I'd call this `deprecatedSmartAccountAccountBanner` or something, because "old" is too generic (v1 accs are also old).

PetromirDev (approved):
Same as Kalo's feedback. Nice work otherwise :raised_hands:

superKalo (approved):

PR #1130, opened 2024-11-27
Fix/ Pinned Tokens Not Passed to Portfolio lib on failed RPC

gergana95:
Closes [3294](https://github.com/AmbireTech/ambire-app/issues/3294)

Update the structure of networksWithAssetsByAccounts with an object structure and boolean for each network.
Display rpc еrror banner on 1. network is with asset 2. network is not yet added in the networksWithAssetsByAccounts


src/libs/banners/banners.ts
```diff
@@ -317,17 +317,13 @@ export const getKeySyncBanner = (addr: string, email: string, keys: string[]) =>
...
+  networks
 }: {
   providers: RPCProviders
   networks: Network[]
-  networksWithAssets: NetworkId[]
 }): Banner[] => {
   const banners: Banner[] = []
-  const networkIds = getNetworksWithFailedRPC({ providers }).filter((networkId) =>
-    networksWithAssets.includes(networkId)
-  )
```
PetromirDev, line 330: With this change, users will see error banners on irrelevant networks

gergana95: Updated and removed this.

src/controllers/portfolio/portfolio.ts
```diff
@@ -138,6 +139,11 @@ export class PortfolioController extends EventEmitter {
       await this.#accounts.initialLoadPromise
       this.tokenPreferences = await this.#storage.get('tokenPreferences', [])
       this.#previousHints = await this.#storage.get('previousHints', {})
+      const networksWithAssets = await this.#storage.get('networksWithAssetsByAccount', {})
+
+      if (!Array.isArray(networksWithAssets)) {
+        this.#networksWithAssetsByAccounts = networksWithAssets
+      }
```
PetromirDev, line 146: This check is incorrect. `networksWithAssetsByAccount` was never an array- its values were. We should check every key's value or introduce a `@v2` storage key and no longer use the old one.

PetromirDev: image

PetromirDev (changes requested):

PetromirDev (approved):

PR #1136, opened 2024-11-28
Fix / Selected Account Controller Not Updated On Preferences Change

gergana95:
Change: dont reset portfolio if account is the same in order to set it when preferences are changed


gergana95:
Closing this in favor of this better universal fix [1140](https://github.com/AmbireTech/ambire-common/pull/1140), due to the issue may be resolved for preferences change in the current PR, but the account may be updated from another properties.

PR #1145, opened 2024-12-02
Fix / Exclude routes with protocol fees for which the user does not have sufficient funds to pay

sonytooo:
* The PR title is misleading. We decided not to exclude routes requiring a native fee when the account has no native balance. Instead, we will show an error message for these routes. The message will explain why the user can’t continue with the selected route and how he can resolve the issue. This error will appear in the swap& bridge form
* Currently we do the checks only for these bridge protocols because we do not have a dynamic way of getting from Socket API which ones require contract fees: **['stargate', 'arbitrum-bridge', 'zksync-native']**


src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -745,8 +747,57 @@ export class SwapAndBridgeController extends EventEmitter {
...
+          let routes = quoteResult.routes || []
+
+          // Exclude routes that require an additional protocol fee in a specific token
+          // if the user does not have sufficient balance of that token.
+          try {
+            routes = routes.filter((route) => {
+              let shouldFilterOut = false
+              if (!route.userTxs) return !shouldFilterOut
+
+              const bridgeTx = route.userTxs.find((tx) => tx.userTxType === 'fund-movr') as
```
superKalo, line 759: You can use the `getIsBridgeTxn`, making this `find((tx) => getIsBridgeTxn(tx.userTxType))`

src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -745,8 +747,57 @@ export class SwapAndBridgeController extends EventEmitter {
...
+
+              // if the required protocolFee token is not in the users portfolio
+              // filter out that route to prevent errors on the signAccountOp screen due to
+              // failing simulation and estimation because the user doesn't have sufficient balance to pay the fee
+              if (!tokenToPayFeeWith) shouldFilterOut = true
+
+              return !shouldFilterOut
+            })
+          } catch (error) {
+            console.error(error)
```
superKalo, line 792: Should continue further if the logic breaks, right? Because it will behave just like it was before these changes. I tend to agree.

sonytooo: yes, the idea is to use a try-catch block to handle cases where certain properties in the nested structure might be missing, possibly due to a socket mistake. If the code inside the block throws, the quote will default to the original routes array

sonytooo:
In most cases I’ve tested, the problematic route is Stargate, and it is often the only route found for the given pair. How would you suggest displaying this to the user while still preventing them from proceeding?

PR #1149, opened 2024-12-04
Update/ Portfolio hints banner improve ux and clean up hints logic

PetromirDev:
## Changes:
- Logically separate hints- from external API and additional hints, instead of adding learned NFTs to previousHints
- Only display a portfolio error banner if fallback hints(from storage) are missing or older than 10 minutes
- Don't store empty structures in storage. We used to store empty hints for networks, unsupported by velcro, instead of not storing anything. This PR changes that- hints will only be stored if there is a successful call to velcro
- Improves naming and adds docs


jordan-enev (approved):
lgtm :clap:

one small comment left and already discussed

gergana95 (approved):
Approved! I tested and I suggest to test the banners more

PR #1157, opened 2024-12-06
Feature / Swap & Bridge API error handling and messaging

superKalo:
Improves the Swap & Bridge error handling and messaging:

- Changed: A common handler for all Socket API responses (wired up to all API calls now ✅), that processes Socket API responses and throws custom errors for various failures, including handling the API's unique response structure.
- Changed: Do not swallow the upstream errors coming from our Swap & Bridge service provider (would greatly help us debug if the glitch is coming on our end or not when someone reports a bug):
  Screenshot 2025-01-29 at 17 57 59
- Changed: Do not swallow the errors in case something in our logic breaks, display it with sensible (fallback) messages.
  Screenshot 2025-01-29 at 21 11 34


src/services/socket/api.ts
```diff
@@ -93,20 +96,68 @@ export class SocketAPI {
...
+    errorPrefix: string
+  }): Promise<T> {
+    let response: CustomResponse
+
+    try {
+      response = await fetchPromise
+    } catch (e: any) {
+      const message = e?.message || 'no message'
+      const status = e?.status || 'unknown'
+      const error = `${errorPrefix} Upstream error: <${message}>, status: <${status}>`
```
PetromirDev, line 117: Perhaps we can hide Upstream error: `no message` and status: `unknown`.

src/services/socket/api.ts
```diff
@@ -225,33 +268,30 @@ export class SocketAPI {
...
-                toAsset: normalizeIncomingSocketToken(step.toAsset)
-              }))
-            : undefined
+          ...('steps' in userTx && {
+            steps: userTx.steps.map((step) => ({
+              ...step,
+              fromAsset: normalizeIncomingSocketToken(step.fromAsset),
+              toAsset: normalizeIncomingSocketToken(step.toAsset)
+            }))
+          })
```
PetromirDev, line 294: I'm pretty sure that we have to fallback to an object because we are spreading ...

src/services/socket/api.ts
```diff
@@ -299,19 +339,16 @@ export class SocketAPI {
...
-      method: 'POST',
-      headers: this.#headers,
-      body: JSON.stringify(params)
+    const response = await this.#handleResponse<SocketAPISendTransactionRequest>({
+      fetchPromise: this.#fetch(`${this.#baseUrl}/route/start`, {
+        method: 'POST',
+        headers: this.#headers,
+        body: JSON.stringify(params)
+      }),
+      errorPrefix: 'Unable to start the route.'
```
PetromirDev, line 348: Could we improve this message?

src/services/socket/api.ts
```diff
@@ -344,32 +381,28 @@ export class SocketAPI {
...
-                toAsset: normalizeIncomingSocketToken(step.toAsset)
-              }))
-            : undefined
+        ...('steps' in userTx && {
+          steps: userTx.steps.map((step) => ({
+            ...step,
+            fromAsset: normalizeIncomingSocketToken(step.fromAsset),
+            toAsset: normalizeIncomingSocketToken(step.toAsset)
+          }))
+        })
```
PetromirDev, line 405: again, this could cause an error

PetromirDev (changes requested):

src/libs/swapAndBridge/swapAndBridgeErrorHumanizer.ts
```diff
@@ -0,0 +1,23 @@
...
+    e instanceof SwapAndBridgeError
+  ) {
+    return e
+  }
+
+  // Last resort (fallback) error handling
+  const message = e?.message || 'no details'
+  const errorMessage = `Unexpected error happened in the Swap & Bridge flow. Try again later or contact Ambire support. Details: <${message}>`
+
+  return new Error(errorMessage)
```
PetromirDev, line 22: Trim the message if it exceeds x characters

PetromirDev (changes requested):

src/services/socket/api.ts
superKalo: This conditional spread is safe, it handles both cases (presence and absence of `fromAsset`) gracefully without causing errors 🙄 

I asked an AI friend to write a test, that we can run in the browser console, that proofs this:

```
// Define a function that uses the conditional spread operator
function normalizeUserTx(userTx) {
  return {
    ...userTx,
    ...('fromAsset' in userTx && {
      fromAsset: `normalized_${userTx.fromAsset}`
    }),
    toAsset: `normalized_${userTx.toAsset}`
  };
}

// Test cases
const testCases = [
  { input: { toAsset: 'asset1' }, expected: { toAsset: 'normalized_asset1' } },
  { input: { fromAsset: 'asset2', toAsset: 'asset3' }, expected: { fromAsset: 'normalized_asset2', toAsset: 'normalized_asset3' } }
];

// Execute test cases
testCases.forEach(({ input, expected }, index) => {
  const result = normalizeUserTx(input);
  console.log(`Test case ${index + 1}:`, JSON.stringify(result) === JSON.stringify(expected) ? 'Passed' : 'Failed');
});
```

Tests passed ✅

src/services/socket/api.ts
superKalo: Same as https://github.com/AmbireTech/ambire-common/pull/1157#discussion_r1937557694

src/services/socket/api.ts
superKalo: The title of the card is "Pending route", the call to action is "Proceed", that's why "Unable to start the route" makes sense to me. I can sync up for all these messages at batch with Stefan down the road, closing this one now so as not to block the PR.

src/services/socket/api.ts
superKalo: I changed it to displaying the status only if one exists in https://github.com/AmbireTech/ambire-common/pull/1157/commits/f251912586546a977c52002fa8451efea15db7a2, but left the message as is, since I haven't yet had a situation with no message (and it might be actually useful if we see no message from the upstream provider while getting bug reports).

src/libs/swapAndBridge/swapAndBridgeErrorHumanizer.ts
superKalo: That's an excellent idea, thanks! Protection against crazy long error messages added in https://github.com/AmbireTech/ambire-common/pull/1157/commits/b573fc69ec34383e1fd6c1bb82fabecf9b5e80bc

sonytooo (approved):

PR #1159, opened 2024-12-09
Feature / Force re-estimate account op upon another account op being confirmed

superKalo:
Handles a corner case - after BA signs an approval, the user may open the next transaction too soon. The new transaction could estimate before the account op with the approval gets confirmed on-chain, causing estimation errors (that's expected). But these stuck until the next scheduled re-estimate (that currently happens every 60s).

Forcing a re-estimate here updates this quicker and the user can proceed with the next transaction.

https://github.com/user-attachments/assets/51c1b5ac-346b-47be-af9c-6d6ba81859d8

Might help with https://github.com/AmbireTech/ambire-app/issues/3104


superKalo:
Closed in favor of https://github.com/AmbireTech/ambire-common/pull/1163

PR #1163, opened 2024-12-11
Fix / Basic Account Estimation Error Due to a Pending Approval

sonytooo:
* added: 'waiting-approval-to-resolve' routeStatus
* do not build both approval and swap or bridge tx at the same time to prevent bugs - first build the approval tx and when confirmed build the swap or bridge tx 

* Refactoring part resolves https://github.com/AmbireTech/ambire-app/issues/3542:
    * Previously, the routeStatus was updated after signing a transaction in resolveAccountOpAction within the main function. However, with smart accounts, this approach caused issues because the relayer would modify or batch the transaction with others, resulting in a different txnId. This mismatch led to errors in Socket, as the incorrect txnId was passed, preventing proper status updates for some routes indefinitely.
    * Now, all routeStatus updates for the whole lifecycle of the activeRoute occur after receiving updates in the SubmittedAccountOp from the activity controller. This logic has been consolidated into a single function, making the update process for activeRoutes more robust, less prone to errors, and easier to maintain.

### DIFF
--- a/src/controllers/activity/activity.ts
+++ b/src/controllers/activity/activity.ts
@@ -304,17 +304,19 @@ export class ActivityController extends EventEmitter {
   async updateAccountsOpsStatuses(): Promise<{
     shouldEmitUpdate: boolean
     shouldUpdatePortfolio: boolean
+    updatedAccountsOps: SubmittedAccountOp[]
   }> {
     await this.#initialLoadPromise
 
     if (!this.#selectedAccount.account || !this.#accountsOps[this.#selectedAccount.account.addr])
-      return { shouldEmitUpdate: false, shouldUpdatePortfolio: false }
+      return { shouldEmitUpdate: false, shouldUpdatePortfolio: false, updatedAccountsOps: [] }
 
     // This flag tracks the changes to AccountsOps statuses
     // and optimizes the number of the emitted updates and storage/state updates.
     let shouldEmitUpdate = false
 
     let shouldUpdatePortfolio = false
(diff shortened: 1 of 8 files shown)


Ivshti (approved):

superKalo (approved):

superKalo:
Solves https://github.com/AmbireTech/ambire-app/issues/3542 as well 🚀

src/controllers/activity/activity.ts
```diff
@@ -354,10 +356,12 @@ export class ActivityController extends EventEmitter {
             if (fetchTxnIdResult.status === 'rejected') {
               this.#accountsOps[selectedAccount][networkId][accountOpIndex].status =
                 AccountOpStatus.Rejected
+              updatedAccountsOps.push(this.#accountsOps[selectedAccount][networkId][accountOpIndex])
               return
             }
             if (fetchTxnIdResult.status === 'not_found') {
               declareStuckIfQuaterPassed(accountOp)
+              updatedAccountsOps.push(this.#accountsOps[selectedAccount][networkId][accountOpIndex])
```
borislav-itskov, line 364: `declareStuckIfQuaterPassed` is called two times in `activity.ts` and we do `updatedAccountsOps.push` only at one of them, here. At code line 394 we also call `declareStuckIfQuaterPassed`.

I suggest moving `updatedAccountsOps.push` to `declareStuckIfQuaterPassed` directly and hence remove this line of code if this makes sense

borislav-itskov (approved):
Approve, but let's just change the `declareStuckIfQuaterPassed`

PR #1166, opened 2024-12-12
Add / Unit test for v1 acc sign message on unsupported network

Maskln:


borislav-itskov (approved):

superKalo (approved):
Approved.

Minor: we discussed offline to try the sexier jest syntax (`expect().rejects. toThrow`).

PR #1169, opened 2024-12-13
Improve/estimation slow request

borislav-itskov:
Add a timeout limit:
* 3.5s to pimlico estimate gas prices
* 3.5s to standard gas price call (provider.getBlock())
* 5s to estimate requests (estimate gas, estimate deployless, bundler estimate)

If the timeout is reached, it retries each request 5 times and then it declares a failure


src/controllers/main/main.ts
```diff
@@ -582,6 +583,7 @@ export class MainController extends EventEmitter {
         return this.isSignRequestStillActive
       }
     )
+    estimationErrorEmitter.init(this.signAccountOp)
```
superKalo, line 586: We synced up offline with Bobby and Simo, that introducing another approach for emitting sounds the easiest way to do this. In future, could be refactored if we came up with better approach.

PR #1179, opened 2024-12-17
Simulation: prioritize enumerable data

JIOjosBG:
we used to check only the passed nft id and if there are no passed nft ids we check the functions from the enumerable nft standard
now we first check the contract supports the enumerable interfaceId. if so we rely on `tokenOfOwnerByIndex`, if not enumerable we use the passed nft ids if there are any

resolves https://github.com/AmbireTech/ambire-app/issues/3621


src/libs/portfolio/getOnchainBalances.ts
```diff
@@ -261,7 +261,7 @@ export async function getTokens(
       ],
       deploylessOpts
     )
-
+  console.log(before[0])
```
borislav-itskov, line 264: let's remove this

src/services/assetInfo/assetInfo.ts
```diff
@@ -1,8 +1,9 @@
-import { JsonRpcProvider, ZeroAddress } from 'ethers'
+import { JsonRpcProvider, Wallet } from 'ethers'
```
borislav-itskov, line 1: unused import

PR #1181, opened 2024-12-18
Fix / Active Route Removal

sonytooo:
* Added removeActiveRoute function to the MainController
    * When called the active route and all related userRequests will be removed

resolves: https://github.com/AmbireTech/ambire-app/issues/3028


superKalo (approved):

Ivshti (commented):
lgtm - i don't like that we're hardcoding edge cases that deep into the code, but we can figure it out at a later point in time, right now the most important thing is for things to work

PR #1183, opened 2024-12-18
Updates/ Deployless timeout to 5000 for invictus RPCs and 15000 ms for custom ones

PetromirDev:
The implementation may seem ugly but this is the easiest way to implement RPC-specific timeouts.


borislav-itskov (approved):
it's perfect for the time being

PR #1188, opened 2024-12-19
Enhancement/ Improve error humanization

PetromirDev:
## Changes:
- Add CustomError error decoder
- Update generic errors
- Handle more error reasons
- Update `The transaction cannot be estimated because` to `The transaction will fail because`
- Fix `toUtf8String` throws an error (credits to Bobby)

Closes https://github.com/AmbireTech/ambire-app/issues/3634


borislav-itskov (approved):
:rocket:

Ivshti (commented):
this should be a fantastic improvement1

PR #1194, opened 2025-01-02
Humanizer: hidden calls inside a call to the current account

JIOjosBG:
the methods `tryCatch`, `tryCatchLimit`, `executeBySelfи`, `executeBySelfSingle` (and `setAddrPrivilege`) can be executed without a signature if the contracts calls itself which is done if the method call is requested as a normal contract call. We used to not display the inner calls. This PR addresses this issue.

resolves https://github.com/AmbireTech/ambire-app/issues/3667

A second attack vector with `execute`, `executeMultiplel` and `executeBySender` is present when the current account is an authorized signer for another ambire account. the only case when a dapp might request such calls is in spear fishing attacks

the 7 cases:
- 4 functions triggered by self can be displayed as the calls within them(5 here because one of them is a batch of 2)
- 3 functions towards other contracts - when the current account is a signer (for simplicity we do not display the inner calls)
 
image


Ivshti (approved):
looks alright, just make sure it can't blow up for some stupid reason. I read it a few times and it seems ok

JIOjosBG:
One more vuln:
what if the drain call/calls is inside a tryCatch inside a tryCatch inside a tryCatch inside....
we need to unwrap recusively

PR #1198, opened 2025-01-06
Min 10$ balance to use the gas tank

borislav-itskov:
add: a minimum requirement of 10$ in the account to use the gas tank

### DIFF
--- a/src/controllers/signAccountOp/signAccountOp.ts
+++ b/src/controllers/signAccountOp/signAccountOp.ts
@@ -396,16 +396,13 @@ export class SignAccountOpController extends EventEmitter {
     let balance = 0
     Object.keys(currentPortfolio).forEach((networkName) => {
       const networkPortfolio = currentPortfolio[networkName]
-      if (!networkPortfolio) return
-      if (!networkPortfolio.result) return
-      if (!networkPortfolio.result.total) return
-      if (!networkPortfolio.result.total.usd) return
+      if (!networkPortfolio?.result?.total?.usd) return
 
       balance += networkPortfolio.result.total.usd
     })
     if (balance < 10 && this.accountOp.gasFeePayment && this.accountOp.gasFeePayment.isGasTank) {
       errors.push(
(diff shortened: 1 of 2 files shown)


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -393,19 +393,20 @@ export class SignAccountOpController extends EventEmitter {
...
+      if (!networkPortfolio.result.total.usd) return
 
-      return networkPortfolio.result.total.usd > 0
+      balance += networkPortfolio.result.total.usd
     })
-    if (!hasBalance && this.accountOp.gasFeePayment && this.accountOp.gasFeePayment.isGasTank) {
-      errors.push("Gas tank isn't allowed on an empty account. Please add funds to your account")
+    if (balance < 10 && this.accountOp.gasFeePayment && this.accountOp.gasFeePayment.isGasTank) {
+      errors.push(
+        "Gas tank isn't allowed on accounts with < 10$ balance. Please add funds to your account"
```
superKalo, line 408: Minor: it should be `$10`.

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -393,19 +393,20 @@ export class SignAccountOpController extends EventEmitter {
...
+    Object.keys(currentPortfolio).forEach((networkName) => {
       const networkPortfolio = currentPortfolio[networkName]
-      if (!networkPortfolio) return false
-      if (!networkPortfolio.result) return false
-      if (!networkPortfolio.result.total) return false
-      if (!networkPortfolio.result.total.usd) return false
+      if (!networkPortfolio) return
+      if (!networkPortfolio.result) return
+      if (!networkPortfolio.result.total) return
+      if (!networkPortfolio.result.total.usd) return
```
superKalo, line 402: Minor, this could be shrinked to 1 condition with ts optional chaining `!networkPortfolio?.result?.total?.usd`. Also in case the total is `0`, we may (or may not) want to be truethful in this case.

PR #1201, opened 2025-01-07
Config / Run Unit tests against NodeJS v22.x (to be in sync with the extension v)

superKalo:
Self-explanatory.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Install Node.js ⚙️
       uses: actions/setup-node@v3
       with:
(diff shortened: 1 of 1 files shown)


superKalo:
Couple of tests cause TS warns (that ultimately fail the tests), I'll see what's going on.

superKalo:
Tests were failing not caused by TS warns. It was the structuredClone polyfil from @ungap/structured-clone. Unit tests were prev ran against an older NodeJS (v16.x). In Node 22+, `structuredClone` is already defined and marked read-only, so trying to reassign it throws that error. Polyfil no longer needed, so I removed it altogether with the package.

PR #1210, opened 2025-01-10
UX/ Reduce RPC error banners

PetromirDev:
## The problem
Account states are updated every 5 minutes automatically and on manual refresh on the dashboard. When the user is simply browsing on his laptop, the account state update fails in the background and the user decides to open his wallet to check how his portfolio is going he will see an RPC error banner, even if there was a successful update 5 minutes ago.
## The solution
Save an update timestamp and compare it before displaying an RPC error banner. If the last update is from > 8 minutes ago, an error banner is displayed. Also, the banner is displayed immediately in case of a manual refresh on the dashboard as a bad RPC means that there will be portfolio and defi positions banners, which must be replaced with a single RPC error banner for better UX. 

Closes https://github.com/AmbireTech/ambire-app/issues/3105

### DIFF
--- a/src/controllers/signAccountOp/signAccountOp.ts
+++ b/src/controllers/signAccountOp/signAccountOp.ts
@@ -59,6 +59,7 @@ import { AccountOpAction } from '../actions/actions'
 import EventEmitter from '../eventEmitter/eventEmitter'
 import { KeystoreController } from '../keystore/keystore'
 import { PortfolioController } from '../portfolio/portfolio'
+import { ProvidersController } from '../providers/providers'
 import {
   getFeeSpeedIdentifier,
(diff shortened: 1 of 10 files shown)


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -189,6 +193,7 @@ export class SignAccountOpController extends EventEmitter {
     super()
 
     this.#accounts = accounts
+    this.#providers = providers
```
borislav-itskov, line 196: @PetromirDev maybe you had an idea here and forgot to remove the providers?

borislav-itskov (changes requested):

PetromirDev:
This is no longer needed

PR #1213, opened 2025-01-13
Feature / Handle hidden tokens in the Swap & Bridge token lists

superKalo:
Hidden tokens 1) Won't be displayed in the "From" token list and 2) They won't be added to the "Receive" token list as additional tokens from the portfolio (example: hiding SKYA token on Base), BUT 3) They will STILL appear in the "Receive" list if they are present in service provider's to token list (example: hiding cbBTC on Base).

Resolves https://github.com/AmbireTech/ambire-app/issues/3743


gergana95 (approved):
better this way 👍

PR #1225, opened 2025-01-15
Fix / flatten and move all tokens into one array after pagination

gergana95:
Fixes issue [3680](https://github.com/AmbireTech/ambire-app/issues/3680)

Changelog:
- Passing another kind of structure with `[tokens, { metadata: blockNumber, beforeNonce, afterNonce} ]` from the getOnChainBalances after the deployless call.
- Update the Flatten logic and flatten the results no matter how many arrays we receive into only one with tokens and metadata into an object.


src/libs/portfolio/portfolio.ts
```diff
@@ -246,7 +251,10 @@ export class Portfolio {
       )
     ])
 
-    const [tokensWithErrResult, blockNumber, beforeNonce, afterNonce] = tokensWithErr
+    console.log(`tokensWithErr after flattening on network ${networkId}:`, tokensWithErr)
```
Ivshti, line 254: is the console.log intentional? we shouldn't have console logs in libraries imo, unless it's on errors

Ivshti (changes requested):

src/libs/portfolio/pagination.ts
```diff
@@ -8,6 +8,30 @@ export function paginate(input: any[], limit: number): any[][] {
   return pages
 }
 
-export async function flattenResults(everything: Promise<any[]>[]): Promise<any[]> {
-  return Promise.all(everything).then((results) => results.flat())
+export function flattenResults(everything: Promise<any[]>[]): Promise<any[]> {
```
Ivshti, line 11: we can add types to this to make this code safer.

And to the function that returns the results that we pass into here

src/libs/portfolio/portfolio.ts
gergana95: Its only for testing before merge, will remove it

jordan-enev (approved):
:rocket: great finding & fix

PR #1230, opened 2025-01-16
Fix / Swap & Bridge Route Prepare Returns Status 500

sonytooo:
Fixed: route/prepare returns status 500 because of a pending tx

Resolves: https://github.com/AmbireTech/ambire-app/issues/3757


borislav-itskov (approved):

superKalo:
👏

PR #1231, opened 2025-01-17
Refactor/ Improve selected acc portfolio

PetromirDev:
### Reduce redundant data
Instead of repeating the same `tokens` and `collections` 3 times- in `latest`, `pending` and at the object's root, this PR leaves `tokens` and `collections` only at the root of the selected account portfolio structure. 
### Reduce object iterations
This PR reduces the iterations over `pendingStateSelectedAccount` from 3 to 1.
### Remove `tokenAmounts` 
`tokenAmounts` stored `latestAmount` and `pendingAmount` for each token in a separate variable. This PR attaches both properties to every token.

PR #1245, opened 2025-01-23
Fix / pass additional portfolio tokens to array in calculating account portfolio

gergana95:
Fixes an issue with gas tank tokens not visible on the Dashboard. Buf after refactoring from [this](https://github.com/AmbireTech/ambire-common/pull/1231) PR

### DIFF
--- a/src/libs/selectedAccount/selectedAccount.ts
+++ b/src/libs/selectedAccount/selectedAccount.ts
@@ -238,6 +238,11 @@ export function calculateSelectedAccountPortfolio(
       const networkTotal = Number(result?.total?.usd) || 0
       newTotalBalance += networkTotal
 
+      if (network === 'gasTank' || network === 'rewards') {
+        const networkTokens = result?.tokens || []
+        tokens.push(...networkTokens)
+      }
+
       const networkCollections = result?.collections || []
       collections.push(...networkCollections)
     }


sonytooo (approved):

superKalo (approved):
I tested with the same account on PROD and with these changes, I confirm Gas Tank tokens are now back on the Dashboard after this change ✅

src/libs/selectedAccount/selectedAccount.ts
```diff
@@ -238,6 +238,11 @@ export function calculateSelectedAccountPortfolio(
       const networkTotal = Number(result?.total?.usd) || 0
       newTotalBalance += networkTotal
 
+      if (network === 'gasTank' || network === 'rewards') {
+        const networkTokens = result?.tokens || []
+        tokens.push(...networkTokens)
+      }
```
PetromirDev, line 244: I think that the bug is larger and not limited to gasTank and rewards tokens. When doing this refactoring I haven't thought of the case where we have `latest` state but not `pending` state- in this case, there will be no tokens in `tokens`, as we are always looping through `pendingTokens`. I wrote the logic this way to limit the number of loops but it introduces this bug.

PetromirDev (changes requested):

src/libs/selectedAccount/selectedAccount.ts
PetromirDev: Yep, this is after manually disabling pending updates
image

PR #1246, opened 2025-01-23
Config / Enable Swap & Bridge convenience fees (0.5%)

superKalo:
Resolves https://github.com/AmbireTech/ambire-app/issues/2877


superKalo:
Bobby said money are in the bank! So it should be all good.

Ivshti (commented):
awesome that socket lets us do this so easily!

PR #1250, opened 2025-01-24
Bug/ Selected account portfolio tokens will be missing if there is no pending state

PetromirDev:
This is a potential bug that you can force by adding the following line in `deployless.call`:
```
 if (opts.blockTag === 'pending') throw new Error('pending blockTag not supported')
```
The problem stems from the fact that we always loop through `pendingTokens` when building the tokens for selectedAccount's `tokens` array.


gergana95 (approved):
Tested and no issues found! I believe this fixes issues we may have had with missing pending state 🎉

Ivshti:
good catch

PR #1252, opened 2025-01-24
Humanizer: swap amount out one more case

JIOjosBG:
resolves: https://github.com/AmbireTech/ambire-app/issues/3836
image


Ivshti (commented):
lgtm but it's easy to induce an error here by making data that starts with this prefix but isn't valid

JIOjosBG:
> lgtm but it's easy to induce an error here by making data that starts with this prefix but isn't valid

Well, that is an issue present in every module of the humanizer. Instead of addressing it we simply have each module in a try catch

PR #1257, opened 2025-01-28
Enhancement/ Refactor token preferences

PetromirDev:
### Refactor token preferences so custom tokens and tokens with preferences can be differentiated
The previous implementation stored custom tokens and tokens with preferences in the same structure(variable) and didn't mark custom tokens in any way. When a new preference was added, we started treating the token as a custom token. Example: When you hide a token that is not custom it will be displayed as hidden, but also as custom(with a trash icon next to it)
### Optimize custom tokens
Custom tokens stored unnecessary information like- `decimals` and `symbol` in the best case and the whole token in the worst case. Worst case example:
image
This information is not needed as the portfolio lib fetches it anyway.
### Move `isHidden` to flags
`isHidden` wasn't a flag. It was being added to the root of the token structure.


src/controllers/portfolio/portfolio.test.ts
```diff
@@ -743,59 +736,121 @@ describe('Portfolio Controller ', () => {
...
-    })
+    const tokenIsRemoved = controller.customTokens.find(
+      (token) => token.address === customToken.address && token.networkId === customToken.networkId
+    )
+    expect(tokenIsRemoved).toBeFalsy()
+    expect(getCustomTokenFromPortfolio()).toBeFalsy()
   })
 
-  test('Update Token Preferences - hide a token and portfolio returns isHidden flag', async () => {
+  test('Cannot add the same custom token twice', async () => {
```
gergana95, line 775: Do you think it would be better to write a test for the migration?

gergana95 (approved):
Tested and switched a few times to the current implementation in v2 and it all seems to work good. Nice one 🎉

PR #1258, opened 2025-01-28
Main controller: method for deleting individual calls

JIOjosBG:
Adds the ability to delete a single call from the accountOp


src/controllers/actions/actions.ts
```diff
@@ -123,7 +123,13 @@ export class ActionsController extends EventEmitter {
...
 
         this.actionsQueue = this.actionsQueue.filter((a) => a.type === 'accountOp')
-        if (this.actionsQueue.length) {
+        if (
+          this.actionsQueue.filter(
+            (a) =>
+              // we do not want notification if THE CURRENT USER does not have pending-to-sign accountOps
+              a.type === 'accountOp' && a.accountOp.accountAddr === selectedAccount.account?.addr
+          ).length
+        ) {
```
PetromirDev, line 132: I'm pretty sure we can use visibleActionsQueue as it is specifically filtered by account

src/controllers/main/main.ts
```diff
@@ -1512,6 +1512,78 @@ export class MainController extends EventEmitter {
...
+      networkId,
+      nonce: accountState.nonce,
+      userRequests: this.userRequests,
+      actionsQueue: this.actions.actionsQueue
+    })
+
+    this.actions.addOrUpdateAction(accountOpAction)
+    this.signAccountOp?.update({ calls: accountOpAction.accountOp.calls })
+    this.estimateSignAccountOp()
+  }
```
PetromirDev, line 1585: Please use emitError everywhere to make this code future-proof. When we integrate Sentry emitError will likely be used to report all errors

PetromirDev (changes requested):

PR #1262, opened 2025-01-29
Feature/eip 7702

borislav-itskov:
Everything necessary to enable eip-7702


contracts/deployless/IAmbireAccount.sol
```diff
@@ -4,8 +4,10 @@ pragma solidity ^0.8.7;
 import '../libs/Transaction.sol';
 
 interface IAmbireAccount {
-	function privileges(address addr) external returns (bytes32);
-	function nonce() external returns (uint);
+	struct AmbireStorage {
```
Ivshti, line 7: identation issue

Ivshti (changes requested):

contracts/libs/Eip712HashBuilder.sol
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  function getUserOp712Hash(
+    PackedUserOperation calldata op,
+    bytes32 hash
+  ) internal view returns (bytes32) {
+    // decode the calls if any
+    Transaction[] memory calls;
+    if (op.callData.length >= 4) {
+      bytes4 executeSelector = 0x6171d1c9;
```
Ivshti, line 81: you can use the interface here; IAmbireAccount.execute.selector

contracts/libs/Eip712HashBuilder.sol
```diff
@@ -0,0 +1,122 @@
...
+      bytes4 executeBySenderSelector = 0xabc5345e;
+      bytes4 functionSig = bytes4(op.callData[0:4]);
+
+      if (functionSig == executeBySenderSelector) {
+        calls = abi.decode(op.callData[4:], (Transaction[]));
```
Ivshti, line 86: this means that transactions where the calldata starts with those selectors but cannot be parsed, won't be validate-able

This is fine, but what if it's a DoS attack as per 4337?

borislav-itskov: Reverting in validateUserOp is fine, we asked the 4337 group

package.json
```diff
@@ -33,6 +33,7 @@
     "bip44-constants": "^128.0.0",
     "ethereumjs-util": "^7.1.3",
     "react": "^18.1.0",
+    "secp256k1": "^5.0.1",
```
superKalo, line 36: Since this package is used in ambire-common (src/libs/keystoreSigner/keystoreSigner.ts), shouldn't we put this package as a "dependency" here? And then - the extension should not have this package installed separately?

src/services/bundlers/bundlerSwitcher.ts
```diff
@@ -42,7 +43,10 @@ export class BundlerSwitcher {
     return this.noStateUpdateStatuses.includes(this.getSignAccountOpStatus())
   }
 
-  canSwitch(bundlerError: Error | null): boolean {
+  canSwitch(acc: Account, bundlerError: Error | null): boolean {
+    // no fallbacks for EOAs
+    if (!acc.creation) return false
```
superKalo, line 48: Maybe `isBasicAccount(acc, state)`?

superKalo:
So far so good 👏

src/controllers/portfolio/portfolio.ts
```diff
@@ -793,13 +793,14 @@ export class PortfolioController extends EventEmitter {
...
                 blockTag: 'pending',
-                ...(currentAccountOps && {
-                  simulation: {
-                    account: selectedAccount,
-                    accountOps: currentAccountOps
-                  }
-                }),
-                isEOA: !isSmartAccount(selectedAccount),
+                ...(currentAccountOps &&
+                  state && {
```
jordan-enev, line 797: Why do we need to check for `state` (AccountOnchainState) here?  When we run a simulation, isn’t the state already fetched and defined in earlier steps?

src/services/bundlers/bundlerSwitcher.ts
borislav-itskov: Haven't started the bundler switchers, yet

src/controllers/accounts/accounts.ts
```diff
@@ -61,14 +70,16 @@ export class AccountsController extends EventEmitter {
...
 
   async #load() {
     await this.#networks.initialLoadPromise
     await this.#providers.initialLoadPromise
-    const [accounts, accountPreferences] = await Promise.all([
+    const [accounts, accountPreferences, storageSignedMessages] = await Promise.all([
       this.#storage.get('accounts', []),
-      this.#storage.get('accountPreferences', undefined)
+      this.#storage.get('accountPreferences', undefined),
+      this.#storage.get('signedMessages', {})
```
jordan-enev, line 82: `signedMessages` should be read from `ActivityController` instead of directly accessing localStorage.

src/controllers/accounts/accounts.ts
```diff
@@ -43,6 +49,9 @@ export class AccountsController extends EventEmitter {
   // Holds the initial load promise, so that one can wait until it completes
   initialLoadPromise: Promise<void>
 
+  // all SignedMessage type 7702-authorization the user has signed
+  #authorizations: InternalSignedMessages
```
jordan-enev, line 53: The above 7702 comment is great, but the first time I saw #authorizations in the code, I thought we stored all types of auth messages there.

src/controllers/accounts/accounts.ts
jordan-enev: However, I've seen that we need #authorizations only in the case we call `#updateAccountStates()`. If so, I would suggest to simply get it from ActivityController, in a similar way we do it here:

```
const entryPointAuthorizationMessageFromHistory = await this.activity.findMessage(
        account.addr,
        (message) =>
          message.fromActionId === ENTRY_POINT_AUTHORIZATION_REQUEST_ID &&
          message.networkId === meta.networkId
      )
```

Doing this, there is not need to read the signed messages from localStorage and keeping `#authorizations` variable up to date.

src/controllers/featureFlags/featureFlags.ts
```diff
@@ -0,0 +1,25 @@
...
+import EventEmitter from '../eventEmitter/eventEmitter'
+
+/**
+ * Responsible for managing (enable/disable) feature flags within the app. The
+ * long-term vision for this is to support dynamic management, currently it
+ * enables or disables features only at runtime. Can be useful for feature
+ * toggling, A/B testing, and gradual feature roll-outs.
+ */
+export class FeatureFlagsController extends EventEmitter {
+  flags = { ...featureFlags }
```
jordan-enev, line 11: I guess this should be private, as we have a setter function and the same time we can mutate its value outside of the class

src/controllers/portfolio/portfolio.ts
borislav-itskov: You are right, it is defined in earlier steps.  
I see this more as a precaution of a rare case where the RPC doesn't work for some reason and returns an invalid/empty state for `AccountOnChainState`. There's risk an error is thrown here instead of just skipping the simulation which would be the lesser evil in my opinion

src/controllers/main/main.ts
```diff
@@ -773,6 +798,66 @@ export class MainController extends EventEmitter {
...
+
+    const accountOpAction = makeSmartAccountOpAction({
+      account: this.accounts.accounts.find((acc) => acc.addr === accountAddr)!,
+      networkId,
+      nonce: this.accounts.accountStates[accountAddr][networkId].nonce,
+      userRequests: this.userRequests,
+      actionsQueue: this.actions.actionsQueue
+    })
+
+    this.actions.addOrUpdateAction(accountOpAction, 'first')
```
jordan-enev, line 820: 1. BA -> add new txn on Odyssey
2. The same BA -> add new txn on Odyssey
3. Sign 7702 auth message.
4. The first txn is removed and replaced with a new AccOp that contains the both transactions.
5. But, the second transaction is still part of the Actions queue.

I would suggest in that case, if we sign an 7702 auth message, first to stack the existing txns into an acc op (as we did it), but the same time to remove all txns those are on the network we authorized.

Screenshot from 2025-02-20 17-46-38
Screenshot from 2025-02-20 17-46-23

src/controllers/main/main.ts
```diff
@@ -1625,6 +1765,168 @@ export class MainController extends EventEmitter {
...
+      has7702(network) &&
+      canBecomeSmarterOnChain(
+        account,
+        accountState,
+        this.keystore.keys.filter((key) =>
+          this.selectedAccount.account!.associatedKeys.includes(key.addr)
+        )
+      )
+    ) {
+      // check if entry point auth is already visible
```
jordan-enev, line 1886: minor, I guess you reffer to 7702, not entry point

src/libs/signMessage/signMessage.ts
```diff
@@ -559,3 +635,52 @@ export function adjustEntryPointAuthorization(signature: string): string {
...
+  signature: EIP7702Signature | string,
+  account: Account,
+  accountState: AccountOnchainState
+): EIP7702Signature | Hex {
+  if (isHexString(signature)) {
+    return account.creation && !accountState.isDeployed
+      ? // https://eips.ethereum.org/EIPS/eip-6492
+        (wrapCounterfactualSign(signature, account.creation) as Hex)
+      : (signature as Hex)
+  }
```
jordan-enev, line 683: 6492 code is duplicated here and into getVerifyMessageSignature, we could reuse it

src/libs/signMessage/signMessage.ts
```diff
@@ -559,3 +635,52 @@ export function adjustEntryPointAuthorization(signature: string): string {
...
+
+  const sig = signature as EIP7702Signature
+  // ethereum v is 27 or 28
+  const v = get7702SigV(sig)
+  return concat([sig.r, sig.s, v]) as Hex
+}
+
+export function getSavedSignature(
```
jordan-enev, line 673: what does 'saved' means here, as I could not get the context

src/controllers/portfolio/portfolio.ts
jordan-enev: On the other hand, when we do the simulation, we don't use this passed `state: AccountOnchainState` anywhere. If it is not used in the portfolio, how this will reflect the simulation and may result in an error?

If you have a strong arguments that we need to do a strong validation for state existance, we could do it in the 2 places we call `this.portfolio.updateSelectedAccount()` in main ctrl , instead of passing the whole state and not using it in the logic. But would be nice to include how the missing state could break the simulation.

src/controllers/featureFlags/featureFlags.ts
borislav-itskov: Done just now

src/controllers/main/main.ts
borislav-itskov: This flow is fixed for the time being, albeit ugly. We're changing the flow with the authorization message completely so I think this part of the code will get deleted. If not, let's discuss it at a later stage

src/controllers/networks/networks.ts
```diff
@@ -131,6 +132,9 @@ export class NetworksController extends EventEmitter {
       )
       if (!predefinedNetwork) {
         this.#networks[networkName].predefined = false
+
+        if (this.#networks[networkName].chainId === 911867n)
```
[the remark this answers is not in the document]
borislav-itskov: I added it to ambire-common's `consts/networks.ts` as this logic is in common :)

PR #1265, opened 2025-01-30
Bug/ NetworkError displayed as TypeError

PetromirDev:
TypeError implies that there is something wrong with our code, which looks bad to users. NetworkError is technically a TypeError which caused the previous behaviour.

When I try to broadcast after stopping my WiFi:
image


PetromirDev:
This turned out to be a bad idea as it breaks the case in which the relayer is actually down.

PR #1267, opened 2025-01-30
Fix / Deployless - Out of gas

jordan-enev:
Closes https://github.com/AmbireTech/ambire-app/issues/3681.

We reduced the token limits for Deployless->StateOverride mode only, as the limit for Proxy mode has already been reduced to the minimum due to the 24KB of calldata.


JIOjosBG (approved):

Ivshti (commented):
lgtm. it's a little bit luck based as some tokens may consume more gas, so we need to check if we have a per-token gas limit. We probably don't, if it's failing like this often

PR #1270, opened 2025-01-31
Update/ Better UX when the portfolio is slow

PetromirDev:
## Changes:
- Add: `isReadyToVisualize` flag to the portfolio in `selectedAccount`. It's true when all networks have loaded or when a timeout has passed. This allows the UI to display tokens when some networks are taking too long (As opposed to the previous logic which required all networks to have loaded)
- Change: Selected account portfolio error to return `networkNames` instead of `networkIds`
- Change: Deployless timeouts to 15000/20000 from 5000/15000
- Change: Update the portfolio on popup open only if 5 minutes have passed since the last update (Imo we should implement similar logic for `defiPositions` and `accountState` as we are updating them every time a popup is opened)
- Fix: `portfolioStartedLoadingAtTimestamp` not reset after changing the selected account

Closes https://github.com/AmbireTech/ambire-app/issues/3853

### DIFF
--- a/src/interfaces/selectedAccount.ts
+++ b/src/interfaces/selectedAccount.ts
@@ -27,6 +27,7 @@ export interface SelectedAccountPortfolio {
   collections: CollectionResultInterface[]
   totalBalance: number
   isAllReady: boolean
+  isLoading: boolean
   networkSimulatedAccountOp: NetworkSimulatedAccountOp
   latest: SelectedAccountPortfolioState
   pending: SelectedAccountPortfolioState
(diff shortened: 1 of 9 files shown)


src/interfaces/selectedAccount.ts
```diff
@@ -26,6 +26,12 @@ export interface SelectedAccountPortfolio {
   tokens: SelectedAccountPortfolioTokenResult[]
   collections: CollectionResultInterface[]
   totalBalance: number
+  /** Either all portfolio networks have loaded or a timeout has been reached and there are tokens.
+   * @example - If the user has 3 networks and 2 of them have loaded, but the third has not and a timeout has been reached
+   * the value of isReadyToVisualize will be true.
+   */
+  isReadyToVisualize: boolean
```
jordan-enev, line 33: nice doc and example :+1:

jordan-enev (approved):
LGTM. One thing we can improve in the case of a forced reload is passing a network filter. Currently, the skeleton is shown for an extended period, waiting 5 seconds or until all networks are fetched. However, in this case, we only need the result for a single network.

jordan-enev (approved):

PR #1274, opened 2025-02-03
Refactoring / Swap & Bridge Form Reset on Submit

sonytooo:
Preserve the state of fromSelectedToken, fromChainId, and toChainId after form submission to enhance the UX and avoid confusion

resolves: https://github.com/AmbireTech/ambire-app/issues/3405

### DIFF
--- a/src/libs/swapAndBridge/swapAndBridge.ts
+++ b/src/libs/swapAndBridge/swapAndBridge.ts
@@ -74,25 +74,20 @@ export const sortPortfolioTokenList = (accountPortfolioTokenList: TokenResult[])
  * Not all tokens in the portfolio are eligible.
  */
 export const getIsTokenEligibleForSwapAndBridge = (token: TokenResult) => {
-  const getAmount = (t: TokenResult): bigint => {
-    if (typeof t.amountPostSimulation === 'bigint') {
-      // Prevents the token from disappearing from the list if the
-      // amountPostSimulation equals the full token amount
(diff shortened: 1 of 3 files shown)


src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -74,13 +74,25 @@ export const sortPortfolioTokenList = (accountPortfolioTokenList: TokenResult[])
...
+    if (typeof t.amountPostSimulation === 'bigint') {
+      // Prevents the token from disappearing from the list if the
+      // amountPostSimulation equals the full token amount
+      if (t.amountPostSimulation === 0n && t.amount > 0n) return t.amount
+
+      return t.amountPostSimulation
+    }
+
+    return t.amount
+  }
```
superKalo, line 87: IMO this is confusing, it conflicts with the way the other helper `getTokenAmount` works (gets `amountPostSimulation` or the `amount` of a token).

Either:

1) Maybe move this next to `src/ambire-common/src/libs/portfolio/helpers.ts` and name it getTokenAmountRelevantForSwapAndBridge maybe (to indicate the diff)?

or

2) How about keeping `Number(getTokenAmount(token)) > 0` but also `&& shouldPreventTokenFromDissapearing` where this is something like:

```
const shouldPreventTokenFromDissapearing = typeof token.amountPostSimulation === 'bigint' && token.amountPostSimulation === 0n && t.amount > 0n
```

PR #1282, opened 2025-02-04
CI: disable hardhat tests in ci

JIOjosBG:
Since hardhat tests do not test:
- code that is under active development
- code that is influenced by external factors

we can keep those tests as just tool that we use during development and not as something that runs on every push on a PR


JIOjosBG:
The main idea was to reduce the time it takes to run the common actions, as we do not need hardhat tests in the CI, but it was the contract compilation which was time consuming and we need it for the jest tests.
Closing this PR

PR #1290, opened 2025-02-07
Add/ token isCustom flag and pass tokenPreferences as hints

PetromirDev:
Needed for https://github.com/AmbireTech/ambire-app/pull/3975


gergana95 (approved):
Tested and seems to work good 💯

PR #1295, opened 2025-02-10
Feature / Keystore Extra Entropy

sonytooo:
* Resolves: https://github.com/AmbireTech/ambire-app/issues/1306
* adds extra entropy to the random seed phrase generation when creating a wallet from a new Seed Phrase


src/libs/entropyGenerator/entropyGenerator.ts
```diff
@@ -0,0 +1,54 @@
...
+      const encoder = new TextEncoder()
+      const uint8Array = encoder.encode(extraEntropy)
+      this.addEntropy(uint8Array)
+    }
+
+    if (this.#entropyPool.length === 0) throw new Error('Entropy pool is empty')
+
+    const hash = getBytes(keccak256(this.#entropyPool))
+    const randomBytesGenerated = randomBytes(length)
+    // ensures non-deterministic final output
```
Ivshti, line 22: this comment is incorrect - not sure what you meant but it is deterministic - the same input will always generate the same output

src/libs/entropyGenerator/entropyGenerator.ts
```diff
@@ -0,0 +1,54 @@
...
+    const now = performance.now()
+
+    if (!now) return
+
+    const timeEntropy = new Uint8Array(new Float64Array([now]).buffer)
+    this.addEntropy(timeEntropy)
+  }
+
+  #collectSystemNoiseEntropy(): void {
+    const systemNoise = randomBytes(16)
```
Ivshti, line 40: that's not really system noise though

src/libs/entropyGenerator/entropyGenerator.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.addEntropy(systemNoise)
+  }
+
+  addEntropy(newEntropy: Uint8Array): void {
+    const combined = new Uint8Array(this.#entropyPool.length + newEntropy.length)
```
Ivshti, line 45: is there a better way to concat?

src/controllers/keystore/keystore.ts
```diff
@@ -301,16 +293,14 @@ export class KeystoreController extends EventEmitter {
...
     // We are not unlocked
     if (!mainKey) {
       if (!this.#keystoreSecrets.length) {
-        const key = getBytes(keccak256(concat([randomBytes(32), toUtf8Bytes(extraEntropy)]))).slice(
-          0,
-          16
-        )
         mainKey = {
-          key,
-          iv: randomBytes(16)
```
Ivshti, line 313: where do we take the extra entropy from?

Ivshti (changes requested):

src/libs/entropyGenerator/entropyGenerator.ts
sonytooo: yes

`this.#entropyPool = new Uint8Array(Buffer.concat([this.#entropyPool, newEntropy]))`

src/libs/entropyGenerator/entropyGenerator.ts
sonytooo: then will rename it to: **collectCryptographicEntropy**

src/controllers/keystore/keystore.ts
sonytooo: https://github.com/AmbireTech/ambire-app/pull/3995/files#diff-a1ec8ef4e2ce0e8518394860a92c5a7c37b390319f38a99a9bb6ecfca0c5b553R1

src/libs/entropyGenerator/entropyGenerator.ts
```diff
@@ -0,0 +1,58 @@
...
+  generateRandomMnemonic(wordCount: 12 | 24, extraEntropy: string): Mnemonic {
+    const wordCountToBytesLength = { 12: 16, 24: 32 }
+    const bytesLength = wordCountToBytesLength[wordCount] || 16 // defaults to 12-word phrase
+    const entropy = this.generateRandomBytes(bytesLength, extraEntropy)
+    const mnemonic = Mnemonic.fromEntropy(entropy, '', LangEn.wordlist())
+    return mnemonic
+  }
+
+  #collectTimeEntropy(): void {
+    const now = performance.now()
```
superKalo, line 39: For the future mobile apps, we need to add a polyfil (not sure if in this repo or in the mobile one), because the Performance API is not available on React Native out of the box. Maybe add a TODO here?

src/libs/entropyGenerator/entropyGenerator.ts
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable no-bitwise */
+import { getBytes, keccak256, LangEn, Mnemonic, randomBytes } from 'ethers'
+
+export class EntropyGenerator {
```
superKalo, line 4: Please add a few notes for context as to why this extra Generator is needed.

src/libs/entropyGenerator/entropyGenerator.ts
superKalo, line 39: If the extension uses `performance.now()` as well in the `extraEntropy`, should we use it here as well?

src/controllers/keystore/keystore.ts
```diff
@@ -301,16 +293,14 @@ export class KeystoreController extends EventEmitter {
       })
 
     let mainKey: MainKey | null = this.#mainKey
+    const entropyGenerator = new EntropyGenerator()
```
superKalo, line 296: Both the extension front-end and the background process initiate `new EntropyGenerator`, meaning it won't be a singleton. Could this be a problem?

src/libs/entropyGenerator/entropyGenerator.ts
sonytooo: Yes, I think that by default, there should be an additional layer of entropy to be added to the randomBytes from Ethers, especially in the cases when extra entropy is omitted or it is an empty string

src/controllers/keystore/keystore.ts
sonytooo: Not an issue—it’s more of a library than a controller, so it can be constructed and used wherever randomness is needed.

src/libs/entropyGenerator/entropyGenerator.ts
sonytooo: added a step by step comment

superKalo (approved):

PR #1303, opened 2025-02-11
Fix / Significant Balance Decrease Banner

jordan-enev:


src/interfaces/signAccountOp.ts
```diff
@@ -6,4 +6,12 @@ type Warning = {
   displayBeforeSign: boolean
 }
 
+export enum TraceCallDiscoveryStatus {
```
PetromirDev, line 9: exports are at the bottom in this file

src/controllers/main/main.ts
```diff
@@ -719,28 +741,30 @@ export class MainController extends EventEmitter {
...
         level: 'silent',
         message: 'Error in main.traceCall',
         error: e
       })
     }
+
+    this.signAccountOp?.calculateWarnings()
+    this.#traceCallTimeoutId = null
+    clearTimeout(timeoutId)
   }
```
PetromirDev, line 768: You've deleted
```
// fire an update request to refresh the warnings if any
 .then(() => this.signAccountOp?.update({}))
```
Please make sure that replacing it with `this.signAccountOp?.calculateWarnings()` is enough and there isn't some edge case. @borislav-itskov should know better

PetromirDev (changes requested):
Good job. I have some minor comments

Ivshti:
@jordan-enev can we please increase the 2 secs to 5 secs

PetromirDev:
@Ivshti @jordan-enev In my opinion, if we decide to increase the timeout there should be some sort of loading state or warning that we are not sure if the user's balance will decrease substantially. I'm not a degen by any means and don't operate with large amounts but I don't take too much time on the sign account op screen- it loads, I scan everything, select a fee speed and sign, which may take less than 5 seconds.

This problem is still present but not that prominent with a 2-second timeout.

Ivshti (approved):

src/controllers/main/main.ts
borislav-itskov: The `calculateWarnings` call has been moved inside `traceCall` and now we're awaiting `updateSelectedAccount`. That leads be to believe we're covered here 👍

borislav-itskov (approved):

PR #1310, opened 2025-02-13
Update/ Swap ctrl to fallback to cache token list and add portfolio tokens on error

PetromirDev:
## Changes and fixes:
- Fallback to cache regardless of age if we fail to fetch the to token list
- "Monkey patch" the token list with WALLET tokens even if we fail to fetch the token list
- Add portfolio tokens even if we fail to fetch the token list


superKalo (approved):
I think it's fine as is, wonder if we could `SocketAPI.addCustomTokens` once, wired up to a promise "finally".

PR #1311, opened 2025-02-14
Refactor gas tank v2

Maskln:


src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -433,10 +504,14 @@ export class SelectedAccountController extends EventEmitter {
...
     return getFirstCashbackBanners({
       selectedAccountAddr: this.account.addr,
-      cashbackStatusByAccount: this.#portfolio.cashbackStatusByAccount
+      cashbackStatusByAccount: this.cashbackStatusByAccount
     })
   }
 
+  get cashbackStatus(): CashbackStatusByAccount {
+    return this.cashbackStatusByAccount
+  }
```
PetromirDev, line 513: Imo this getter should return the cashback status of the selected account

src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -83,6 +88,8 @@ export class SelectedAccountController extends EventEmitter {
   // Holds the initial load promise, so that one can wait until it completes
   initialLoadPromise: Promise<void>
 
+  cashbackStatusByAccount: CashbackStatusByAccount = {}
+
```
PetromirDev, line 92: This state should be private and only used to calculate what is needed for the UI. The UI doesn't have to worry about timestamps and calculate things- it only needs a banner (handled by `firstCashbackBanner`) and a function that should be called when the user clicks on the button in the banner.

src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -267,11 +275,74 @@ export class SelectedAccountController extends EventEmitter {
...
+          currentAccountStatus.cashbackWasZeroAt
+        ),
+        firstCashbackReceivedAt: getUpdatedTimestamp(
+          shouldSetFirstCashbackReceivedAt,
+          currentAccountStatus.firstCashbackReceivedAt
+        ),
+        firstCashbackSeenAt: getUpdatedTimestamp(
+          shouldSetFirstCashbackSeenAt,
+          currentAccountStatus.firstCashbackSeenAt
+        ),
```
PetromirDev, line 314: I'm thinking that we should strongly consider refactoring this logic. Why? Because it could be made dead simple with a string value:
```
[accountId]?: 'no-cashback' | 'unseen-cashback' | 'seen-cashback'
```
or something like this. Timestamps complicate the logic more than they add value.

Also, why do we need `seenModalAt`? Can't we use FE state to display the modal when the user clicks on the button in the banner?

@superKalo

PetromirDev: Pseudo code as an example:
```
async updateCashbackStatus() {
    if (!this.#portfolio || !this.account || !this.portfolio.latest.gasTank) return

    const accountId = this.account.addr
    const gasTankResult = this.portfolio.latest.gasTank.result as PortfolioGasTankResult

    const isCashbackZero = gasTankResult.gasTankTokens?.[0].cashback === 0n
    const cashbackWasZeroBefore = !!this.cashbackStatusByAccount[accountId]?.cashbackWasZeroAt
    const notReceivedFirstCashbackBefore =
      !this.cashbackStatusByAccount[accountId]?.firstCashbackReceivedAt

    if (isCashbackZero) {
      this.cashbackStatusByAccount = {
      ...this.cashbackStatusByAccount,
        [accountId]: 'zero-cashback'
      }
    } else if (!isCashbackZero && cashbackWasZeroBefore && notReceivedFirstCashbackBefore) {
      this.cashbackStatusByAccount = {
      ...this.cashbackStatusByAccount,
        [accountId]: 'unseen-cashback'
      }
    }
  }

onFirstCashbackSeen() {
      this.cashbackStatusByAccount = {
      ...this.cashbackStatusByAccount,
        [accountId]: 'seen-cashback'
      }
      this.emitUpdate()
}
```

src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -447,6 +522,7 @@ export class SelectedAccountController extends EventEmitter {
       ...this,
       ...super.toJSON(),
       firstCashbackBanner: this.firstCashbackBanner,
+      cashbackStatusByAccount: this.cashbackStatus,
```
PetromirDev, line 525: Why is that? `cashbackStatusByAccount` can be used directly. This pattern is only used for getters.

PetromirDev (changes requested):

PetromirDev (changes requested):

PetromirDev (approved):
Good job :100:

src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -1,10 +1,15 @@
 import { getAddress } from 'ethers'
+import { PortfolioGasTankResult } from 'libs/portfolio/interfaces'
```
superKalo, line 2: Bad import ref.

src/libs/selectedAccount/selectedAccount.ts
```diff
@@ -313,3 +317,42 @@ export function calculateSelectedAccountPortfolio(
     pending: stripPortfolioState(pendingStateSelectedAccount)
   } as SelectedAccountPortfolio
 }
+
+export const migrateCashbackStatus = (
```
superKalo, line 321: Please add some context why this is needed and since which extension version migration is needed.

src/interfaces/selectedAccount.ts
```diff
@@ -37,3 +38,16 @@ export interface SelectedAccountPortfolio {
   latest: SelectedAccountPortfolioState
   pending: SelectedAccountPortfolioState
 }
+
+// An old CashbackStatus type before the migration
```
superKalo, line 42: Before which migration? Please give more context in this note.

superKalo (approved):

PR #1313, opened 2025-02-17
Update/ Sign account op error if a SA can't pay with any fee option

PetromirDev:
Closes https://github.com/AmbireTech/ambire-app/issues/3735

image

### DIFF
--- a/src/consts/gasTankFeeTokens.ts
+++ b/src/consts/gasTankFeeTokens.ts
@@ -39,7 +39,8 @@ export default [
     symbol: 'aeth',
     networkId: 'arbitrum',
     decimals: 18,
-    icon: 'https://assets.coingecko.com/coins/images/279/large/ethereum.png'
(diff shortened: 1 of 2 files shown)


src/consts/gasTankFeeTokens.ts
```diff
@@ -39,7 +39,8 @@ export default [
     symbol: 'aeth',
     networkId: 'arbitrum',
     decimals: 18,
-    icon: 'https://assets.coingecko.com/coins/images/279/large/ethereum.png'
+    icon: 'https://assets.coingecko.com/coins/images/279/large/ethereum.png',
+    isAave: true
```
borislav-itskov, line 43: This one is not an AAVE token, it's the native ETH on Arbitrum. For reasons I don't know we've named it AETH.  
I guess you've made it isAave: true because you want to hide it from the error message.  
In that case, we can rename this to "hiddenOnError" for now. Just a proposal

borislav-itskov (changes requested):

PR #1316, opened 2025-02-18
Fix / Always Pending Active Route due to a Failed Transaction

sonytooo:
* Fixed: active route status remains pending forever due to a failed transaction because Socket does not update the route status to Failed in that case


superKalo (approved):

superKalo:
Resolves https://github.com/AmbireTech/ambire-app/issues/4047

PR #1319, opened 2025-02-18
Upgrade deployless so it works with the relayer

borislav-itskov:
Wrap _getConnection in a try/catch as not all RPCs have it.


Ivshti (changes requested):
I have a few issues with this

- if it fails, isProviderInvictus won't even be set - it will be a missing property; fix: initialize it to false beforehand
- if it fails in an unexpected way, the error is swallowed; fix: instead of using try/catch, check if the property is there
- there's a likelyhood that, in the RPC that DOES NOT have getConnection, it has ANOTHER method to get the URL
- the fact that we have `isProviderInvictus` to begin with - behavior shouldn't depend on this. I'll review in detail

Ivshti:
@borislav-itskov I take back/modify two feedbacks

- the property is actually initialized, I didn't see that
- isProviderInvictus only affects the timeouts which is fine, but I'd prefer that this is set on the top level from whoever uses the library, as the library user has more information on whether the provider is invictus and doesn't have to resort to hacks like checking the URL

borislav-itskov:
> @borislav-itskov I take back/modify two feedbacks
> 
> * the property is actually initialized, I didn't see that
> * isProviderInvictus only affects the timeouts which is fine, but I'd prefer that this is set on the top level from whoever uses the library, as the library user has more information on whether the provider is invictus and doesn't have to resort to hacks like checking the URL

We will discuss this with @PetromirDev and fix it

PetromirDev (approved):

Oxbobby:
this is too old and we won't merge it

PR #1322, opened 2025-02-19
Feature / Mechanism to update network(s) config externally

superKalo:
Tweaks for the predefined networks logic so that a new build is NOT needed when adding new network or changing an existing predefined network config. Moreover:

- Mechanism to override network config
- Update networks storage by chainId. Since we decided to update the networkId to ChainId in storage for better handling of newly added by relayer and to keep in mind we need to be assured the network is added correctly since on adding a custom network we fetch the chain id from RPC, but networkId = name of network is added openly.
- Mechanism to update network coming from relayer config which does not support smart accs and check if its supporting them - on predefined version change
- Update custom networks properties from rpc on 24hours
- Take the icons for all networks from the `iconUrls` (currently, for predefined networks, it's hardcoded)
- Leave a network removed from the config as custom network

Big change:
- The migration of all storages to chainId will be left out from this PR, but essentially we want to update all controllers to use chainId instead of networkId. This can be done on all controllers load and with migration with replacing script all networkIds to chainId in storage. Otherwise we would need to have a replacement function on each call of  storage data by networkId.

Relayer part coming in https://github.com/AmbireTech/relayer/pull/1037

Part of https://github.com/AmbireTech/ambire-app/issues/3840

How to test the features:
Prerequisites:
- Add Unichain as custom network on v2 branch, when you switch to `config/additional-networks` branch this chain must be added as predefined and can't be deleted.
- Add custom, hidden tokens, learned tokens in the structure previousHints
After switching to new branches:
- Unichain network is not supporting smart accounts so we update this properties from RPC and it should be properly updated and added a lastUpdatedNetworkInfo timestamp
- Unichain should not be able to be removed
- All  custom, hidden tokens, learned tokens should be updated in storage with chain Id


borislav-itskov (approved):

src/controllers/networks/networks.ts
```diff
@@ -31,7 +31,9 @@ export class NetworksController extends EventEmitter {
 
   #fetch: Fetch
 
-  #networks: { [key: NetworkId]: Network } = {}
+  #callRelayer: Function
+
+  #networks: { [key: string]: Network } = {}
```
superKalo, line 36: Minor: key is chain id, right? So this could be "key in ChainId" or "key in Network['chainId']".

src/controllers/networks/networks.ts
```diff
@@ -68,69 +73,239 @@ export class NetworksController extends EventEmitter {
...
+   * 3. If no networks are found in storage, sets predefined networks and emits an update.
+   * 4. Merges the networks from the Relayer with the stored networks.
+   * 5. Ensures predefined networks are marked correctly and handles special cases (e.g., Odyssey network).
+   * 6. Sorts networks with predefined ones first, followed by custom networks, ordered by chainId.
+   * 7. Updates the networks in storage.
+   * 8. Asynchronously updates network features if needed.
+   *
+   * This method ensures that the application has the most up-to-date network configurations,
+   * handles migration of legacy data, and maintains consistency between stored and relayer-provided networks.
+   */
```
superKalo, line 118: Well documented 👏 , we discussed offline that it would be great if we have time to write unit tests that would cover all these scenarios.

src/controllers/networks/networks.ts
```diff
@@ -68,69 +73,239 @@ export class NetworksController extends EventEmitter {
...
+        Date.now() - network.lastUpdatedNetworkInfo <= 24 * 60 * 60 * 1000
+      )
+        return
+
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      getNetworkInfo(
+        this.#fetch,
+        network.selectedRpcUrl,
+        network.chainId,
+        async (info) => {
```
superKalo, line 286: The getNetworkInfo function is called within a loop, and since it is asynchronous, it can potentially cause issues related to closures and scope.

I'm not sure if this is problematic, but please double-check 🙏 The best practice with these is wrapping all those in a Promise.all I think.

superKalo:
Tested on an extension with custom networks already added. Went though the logic. Great job 👏 

Let's make a call to action to everyone else to test as well once this is merged.

src/controllers/networks/networks.ts
gergana95: Тhe thing is it is chainId.toString(). How do we add this as key?

src/controllers/networks/networks.ts
gergana95: I added a test for the merging relayer logic, will add more later on.

PR #1324, opened 2025-02-21
Feature / Phishing Detection Controller

sonytooo:
* Added PhishingController that fetches the blacklist URLs of https://github.com/MetaMask/eth-phishing-detect and https://github.com/phantom/blocklist and merges them together in a private Set

Resolves: 
* https://github.com/AmbireTech/ambire-app/issues/1897
* https://github.com/AmbireTech/ambire-app/issues/1465
* https://github.com/AmbireTech/ambire-app/issues/1035
* https://github.com/AmbireTech/ambire-app/issues/2643


src/controllers/phishing/phishing.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  async getIsBlacklisted(url: string) {
+    this.emitUpdate()
+    await this.initialLoadPromise
+
+    try {
+      const hostname = new URL(url).hostname
+
+      // blacklisted if it has `ambire` in the hostname but it is not a pre-approved ambire domain
+      if (hostname.includes('ambire') && !hostname.includes('ambire.com')) return true
```
PetromirDev, line 138: I'm pretty sure that this could be bypassed
image

src/controllers/phishing/phishing.ts
```diff
@@ -0,0 +1,161 @@
...
+    const sixHoursInMs = 6 * 60 * 60 * 1000
+
+    if (this.#latestStorageUpdate && Date.now() - this.#latestStorageUpdate < sixHoursInMs) return
+    const storedPhishingDetection = await this.#storage.get('phishingDetection', null)
+
+    if (Date.now() - storedPhishingDetection.timestamp >= sixHoursInMs) {
+      await this.#update(storedPhishingDetection)
+    }
+  }
+
```
PetromirDev, line 129: Does the list have to be updated this often? Why don't we generate the list before builds and use it as a static JSON file? That way we:
- Don't need to use storage
- Don't have to rely on HTTP requests
- Don't have to do any parsing in the extension

We are

src/controllers/phishing/phishing.ts
```diff
@@ -0,0 +1,161 @@
...
+  #windowManager: WindowManager
+
+  #blacklist: Set<string> = new Set() // list of blacklisted URLs
+
+  #latestStorageUpdate: number | null = null
+
+  isReady: boolean = false
+
+  updateStatus: 'LOADING' | 'INITIAL' = 'INITIAL'
```
PetromirDev, line 27: Why not rename it to `isLoading` and make it a boolean?

PetromirDev (changes requested):
This is a high-overview review. I have yet to test the PR and review the code in detail.

src/controllers/phishing/phishing.ts
sonytooo: just followed the pattern of the rest of the update statuses in the other controllers. It can be updated with additional statuses like "DONE" or "ERROR" in the future if needed

src/controllers/phishing/phishing.ts
sonytooo: It’s not ideal. Honestly, I think we should update it even more often than every six hours, but it’s a decent starting point. New blacklisted apps pop up every minute, and if a hyped dApp blows up, we might go a whole week without an update—leaving a lot of users at risk of getting phished.

See how often mm updates its list: https://github.com/MetaMask/eth-phishing-detect/commits/main/src/config.json

PetromirDev:
Long app name
image

PetromirDev (changes requested):
Please write a couple of positive and negative test cases

PetromirDev (approved):

PR #1328, opened 2025-02-24
Bug/ portfolio `PriceFetchError` is never displayed

PetromirDev:
The error is not displayed because `lastSuccessfulUpdate` is updated even when there are portfolio errors with the level `warning`. That is because we are checking if there has been a successful update in the last 10 minutes, and because `lastSuccessfulUpdate` is being updated, the error is not displayed unless there is another critical error.


gergana95 (approved):

Ivshti (commented):
good find

PR #1336, opened 2025-02-25
Add token name to structure

JIOjosBG:
Resolves: https://github.com/AmbireTech/ambire-app/issues/3951

adds token name property to:
- portfolio tokens
- defi positions
- networks (for native asset)


contracts/deployless/BalanceGetter.sol
```diff
@@ -42,7 +44,7 @@ contract BalanceGetter is Simulation {
     TokenInfo[] memory results = new TokenInfo[](len);
     for (uint256 i = 0; i < len; i++) {
       if (tokenAddrs[i] == address(0)) {
-        results[i] = TokenInfo('ETH', address(account).balance, 18, bytes(''));
+        results[i] = TokenInfo('ETH', 'Ether', address(account).balance, 18, bytes(''));
```
Ivshti, line 47: not directly related to the PR, but how do we handle this for non-eth networks? as far as I remember we change the hardcoded value somewhere in the portfolio library in common

jordan-enev (approved):

contracts/deployless/BalanceGetter.sol
JIOjosBG: good catch, it should have been done here https://github.com/AmbireTech/ambire-common/blob/v2/src/libs/portfolio/getOnchainBalances.ts#L214

PR #1337, opened 2025-02-26
UX/ Sign account op - Show only major error instead of stacking few ones

PetromirDev:
Error toasts have been reduced using several improvements:  
1. Gas price toasts now only appear when the background process updates the gas price.  
2. Added a UI error for gas prices.
3. No redundant gas price toasts - they won’t be displayed if there’s an estimation error or if the estimation is loading.  
4. Improved the gas price toast text
5. Improved error messaging - connectivity-related issues now show more relevant errors.  
6. Streamlined error handling - `errorCallback` now displays errors directly in the UI instead of triggering both a toast and a UI error with different messages. This ensures only one clear error message appears.  
7. Fixed: `TypeError` on broadcast.


Ivshti (commented):
I'm not against this but we're making the main controller even more complicated. Nice touch with re-estimation

PetromirDev:
@Ivshti, I agree, but the current behaviour when the user is offline isn't great. Right now, multiple error toasts appear, each showing different messages. Ideally, we should streamline this to:  
- 1 error toast on reestimation  
- 1 error toast on broadcast  
- 1 alert error after X failed estimations  

I haven't achieved this yet which is why the PR is draft. I plan to discuss these changes with Bobby in the next weeks.

borislav-itskov (approved):

PR #1338, opened 2025-02-26
Fix / Sign Message - Overwrite bug 

jordan-enev:
Closes https://github.com/AmbireTech/ambire-app/issues/3263.

- Fix: The SignMessage controller was incorrectly overriding the original message string with a hex value, resulting in a non-human-readable message when signing fails.
- Change: Simplified `PlainTextMessage.message` to accept only `string` values.
- Change: Improved TypeScript parameter definitions for `verifyMessage` by making the main properties required. Enforced a strict requirement that either `message` or `typedData` must be provided, but not both.
- Remove: Removed `finalDigest` parameter as it was never passed to `verifyMessage`. Refactored `verifyMessage` to ensure it only accepts either `message` or `typedData`, never both.

### DIFF
--- a/src/controllers/signMessage/signMessage.ts
+++ b/src/controllers/signMessage/signMessage.ts
@@ -159,15 +159,16 @@ export class SignMessageController extends EventEmitter {
 
       const accountState = this.#accounts.accountStates[account.addr][network.id]
       let signature
+      // It is defined when messageToSign.content.kind === 'message'
+      let hexMessage: string | undefined
+
       try {
         if (this.messageToSign.content.kind === 'message') {
           const message = this.messageToSign.content.message
-          this.messageToSign.content.message = isHexString(message)
-            ? message
-            : hexlify(toUtf8Bytes(message.toString()))
+          hexMessage = isHexString(message) ? message : hexlify(toUtf8Bytes(message))
 
           signature = await getPlainTextSignature(
-            this.messageToSign.content.message,
+            hexMessage,
             network,
             account,
             accountState,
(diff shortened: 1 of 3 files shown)


jordan-enev:
I couldn't find any `finalDigest` usages in @ambire-app and @ambire-common, but it's used in Relayer (credits to @JIOjosBG), so I'm going to restore the parameter.

src/controllers/signMessage/signMessage.ts
```diff
@@ -159,15 +159,16 @@ export class SignMessageController extends EventEmitter {
...
+      // It is defined when messageToSign.content.kind === 'message'
+      let hexMessage: string | undefined
+
       try {
         if (this.messageToSign.content.kind === 'message') {
           const message = this.messageToSign.content.message
-          this.messageToSign.content.message = isHexString(message)
-            ? message
-            : hexlify(toUtf8Bytes(message.toString()))
+          hexMessage = isHexString(message) ? message : hexlify(toUtf8Bytes(message))
```
JIOjosBG, line 168: if the message is not a hexString we should throw err

This function should always receive a message converted once to hex string, regardless if it was hex string in the first place.

if a dapp wants the user to sign the message '0x1234', this func should receive `0x307831323334` and we should not allow `0x1234` to be signed just like this in the first place. we should sign `0x307831323334` 

by the same logic
if for some reason this function receive non-hex string message `Some message`, we should not simply convert it to hex and we should throw some error

jordan-enev: Legends also passes a plain string message. I guess other dapps do it as well. For now, I leave it as it was before.

PR #1340, opened 2025-02-28
Fix / Ensure Account Info

jordan-enev:
Fix: Prevent adding a new request (txn) if the current network account state is not fetched, as this would cause a signing and broadcasting error. Instead, block the request and display an appropriate error toast.

I found the bug, thanks to this test (which failed а single time on my PR): https://github.com/AmbireTech/ambire-common/actions/runs/13547855901/job/37863825321#step:6:312

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -1021,8 +1021,25 @@ export class MainController extends EventEmitter {
     if (!this.accounts.accountStates[accountAddr]?.[networkId])
       await this.accounts.updateAccountState(accountAddr, 'pending', [networkId])
     // If this still didn't work, throw error: this prob means that we're calling for a non-existent acc/network
-    if (!this.accounts.accountStates[accountAddr]?.[networkId])
-      this.signAccOpInitError = `Failed to retrieve account info for ${networkId}, because of one of the following reasons: 1) network doesn't exist, 2) RPC is down for this network`
+    if (!this.accounts.accountStates[accountAddr]?.[networkId]) {
+      // TODO: improve message in case of existing signAccountOp
+      const errorMessage = this.signAccountOp
+        ? `We couldn't retrieve your account information for ${networkId}. Please try reloading your account from the Dashboard. If the issue persists, contact support for assistance.`
+        : `We couldn't complete your last action because we couldn't retrieve your account information for ${networkId}. Please try reloading your account from the Dashboard. If the issue persists, contact support for assistance.`
+
+      this.signAccOpInitError = errorMessage
+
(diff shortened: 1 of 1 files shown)


src/controllers/main/main.ts
```diff
@@ -1010,19 +1010,34 @@ export class MainController extends EventEmitter {
...
+      }
     }
     // If this still didn't work, re-load
     if (!this.accounts.accountStates[accountAddr]?.[networkId])
       await this.accounts.updateAccountState(accountAddr, 'pending', [networkId])
     // If this still didn't work, throw error: this prob means that we're calling for a non-existent acc/network
-    if (!this.accounts.accountStates[accountAddr]?.[networkId])
-      this.signAccOpInitError = `Failed to retrieve account info for ${networkId}, because of one of the following reasons: 1) network doesn't exist, 2) RPC is down for this network`
+    if (!this.accounts.accountStates[accountAddr]?.[networkId]) {
+      const networkName = networkId[0].toUpperCase() + networkId.slice(1)
```
PetromirDev, line 1030: Get the name from this.networks.networks and fallback to the networkId
image

src/controllers/main/main.ts
```diff
@@ -1010,19 +1010,34 @@ export class MainController extends EventEmitter {
...
+        hasAccountInfo: false,
+        errorMessage: `Account ${accountAddr} does not exist`
+      }
     }
     // If this still didn't work, re-load
     if (!this.accounts.accountStates[accountAddr]?.[networkId])
       await this.accounts.updateAccountState(accountAddr, 'pending', [networkId])
     // If this still didn't work, throw error: this prob means that we're calling for a non-existent acc/network
-    if (!this.accounts.accountStates[accountAddr]?.[networkId])
-      this.signAccOpInitError = `Failed to retrieve account info for ${networkId}, because of one of the following reasons: 1) network doesn't exist, 2) RPC is down for this network`
```
PetromirDev, line 1025: Perhaps we should remove `this.signAccOpInitError` entirely? It's still used in `initSignAccOp` and we know the current implementation is problematic, right?

jordan-enev: It was problematic, only in the case I fixed. In `initSignAccOp` we have several validations, and if we have an errors it would be written to `this.signAccOpInitError`. Later, we are showing this error on SignAccOp screen.

src/controllers/main/main.ts
jordan-enev: Good catch! I didn' want to overcomplicate it, but didn't consider that we have network ids having a dash.

PR #1341, opened 2025-03-04
Feature / Recover Lost Connection with Dapps

sonytooo:
* remove the redundant logic of storing dappsSessions in storage


src/controllers/dapps/dapps.ts
```diff
@@ -61,28 +61,18 @@ export class DappsController extends EventEmitter {
   }
 
   async #load() {
-    // eslint-disable-next-line prefer-const
-    let [storedDapps, dappSessions] = await Promise.all([
-      this.#storage.get('dapps', []),
-      this.#storage.get('dappSessions', {})
-    ])
+    const storedDapps = await this.#storage.get('dapps', [])
```
superKalo, line 64: I wonder if we should write a migration that deletes `dappSessions`. My main concern is that in the future, if we by chance introduce `dappSessions` storage key and forget that we prev used it - the extension could crash on older users. That's presuming we forget that we used this storage key, which is very likely. Do you agree?

PetromirDev (approved):

PR #1343, opened 2025-03-04
Fix: it adds the tokens in the list even if the network is not fully loaded

Maskln:
closes: https://github.com/AmbireTech/ambire-app/issues/4116


PetromirDev (approved):
I made a slight change to the condition with the aim of making it more user-readable https://github.com/AmbireTech/ambire-common/pull/1343/commits/f8b06e5c7fbb9afa91071b069ddd528e05018ac8

PR #1334, opened 2025-02-25
Sign Message: allow v1 accs to sign anything if OG mode is activated

JIOjosBG:
## To test
- attempt to sign a plaintext  or typed message, that does not include the address of the accounts (should not be allowed)
- activate OG mode
- attempt step 1 again, but now should be successful

Resolves: https://github.com/AmbireTech/ambire-app/issues/4034

PR #1345, opened 2025-03-05
Refactor: resolve feedback for og mode signing in v1

JIOjosBG:
Feedback here https://github.com/AmbireTech/ambire-common/pull/1334#issuecomment-2700899595


src/controllers/signMessage/signMessage.ts
```diff
@@ -1,3 +1,4 @@
+import { InviteController } from 'controllers/invite/invite'
```
superKalo, line 1: Bad import ref (that's why the unis tests break)

superKalo (approved):
Please fix the minor import problem, otherwise, looks good 👍

superKalo:
And please add a note for the `isOG` bypass in `getEIP712Signature` to give context and explain a bit further why this bypasses the restriction 🙏

PR #1347, opened 2025-03-05
Rich json: add options for indentations for stringifying

JIOjosBG:


superKalo (commented):
No longer crashes for the use-case I caught breaking.

superKalo (approved):
No longer crashes for the use-case I caught breaking.

PR #1348, opened 2025-03-05
Bug/ Duplicate error if signAccountOp is not initialized

PetromirDev:
The error is emitted once using `emitError` and a second time because of `withStatus`


superKalo (approved):
Nice catch!

PR #1357, opened 2025-03-13
Feature/drop eoa mode

PetromirDev:


borislav-itskov (approved):

.eslintrc.js
```diff
@@ -41,6 +41,7 @@ module.exports = {
     'react/no-unstable-nested-components': 'off',
     'react-hooks/exhaustive-deps': 'warn',
     'react/react-in-jsx-scope': 'off',
-    '@typescript-eslint/no-floating-promises': 'warn'
+    '@typescript-eslint/no-floating-promises': 'warn',
+    "semi": ["error", "never"]
```
Ivshti, line 45: quote mismatch

contracts/deployless/Estimation.sol
```diff
@@ -138,12 +139,19 @@ contract Estimation is FeeTokens, Spoof {
     address relayer,
     address oracle
   ) external returns (EoaEstimationOutcome memory eoa) {
-    // simulate the transactions
     SimulationOutcome memory simulation;
-    (simulation, , ) = simulateUnsigned(op, associatedKeys);
```
Ivshti, line 143: this seems like an un-elegant way to do things

why don't we introduce per-call limits in the `simulation` itself and not touch the top-level code

or just, considering you don't use those limits, drop/revert this altogether

src/consts/7702.ts
```diff
@@ -25,3 +25,13 @@ export const networks7702: Custom7702Settings = {
     implementation: SEPOLIA_7702
   }
 }
+
+export interface EIP7702Auth {
```
Ivshti, line 29: putting an interface in consts doesn't seem like the best place for it

src/controllers/main/main.ts
```diff
@@ -2475,13 +2291,34 @@ export class MainController extends EventEmitter {
...
     this.signAccountOp.update({
-      gasPrices: this.gasPrices[accOp.networkId],
+      gasPrices: this.gasPrices[accOp.networkId] || null,
       bundlerGasPrices: this.bundlerGasPrices[accOp.networkId],
       blockGasLimit: gasData && gasData.blockGasLimit ? gasData.blockGasLimit : undefined
     })
     this.emitUpdate()
   }
 
+  async getPortfolioSimulationPromise(op: AccountOp) {
```
Ivshti, line 2301: Dancho should review this part

src/controllers/main/main.ts
Ivshti, line 1: nice job removing logic from this one

src/interfaces/network.ts
Ivshti, line 1: this needs to be synced with Geri/Kalo cause they're implementing network hot loading

src/libs/account/EOA.ts
Ivshti, line 1: I love how those classes worked out.

But maybe the interface could be a bit more generic, like maybe activator/authorization should be in some common setup function?

Ivshti (commented):
wip review, clicking submit so i can send it

PR #1358, opened 2025-03-13
Config / Migrate to single transaction bridging quotes ONLY for the build-in Swap & Bridge

superKalo:
Prev we've used Socket's "multi transaction" bridging, because it gave the most options across many pairs we've tried with. It involves 2 or more transactions for the user either on the source or destination chain. Ref: https://docs.bungee.exchange/bungee-manual/socket-api/guides/multi-tx-bridging

To simplify the build-in Swap & Bridge, as part of https://github.com/AmbireTech/ambire-app/issues/4151, we're migrating to the "single transaction" bridging now - a single transaction on the sending chain and **no transaction on the destination chain**. Ref: https://docs.bungee.exchange/bungee-manual/socket-api/guides/single-tx-bridging

TODO: Sync with the others and test different scenarios, to make sure the shortcoming of having no routes for most of the pairs (that we experienced before) is not a game changer.

### DIFF
--- a/src/services/socket/api.ts
+++ b/src/services/socket/api.ts
@@ -273,7 +273,7 @@ export class SocketAPI {
       userAddress,
       isContractCall: isSmartAccount.toString(), // only get quotes with that are compatible with contracts
       sort,
-      singleTxOnly: 'false',
+      singleTxOnly: 'true',
       defaultSwapSlippage: '1',
       uniqueRoutesPerBridge: 'true'
     })


superKalo:
Closing in favor of https://github.com/AmbireTech/ambire-common/pull/1359. Moreover, Socket's single-txn mode was resulting in no routes for many popular pairs so it was a no-go.

PR #1369, opened 2025-03-17
Fix: token discovery for EOAs

borislav-itskov:
Fix: debug_traceCall was reverting on getBalances when the account was an EOA


Ivshti (commented):
awesome find

PR #1374, opened 2025-03-18
Refactor/estimate controller

Oxbobby:
Move the estimation into it's own controller, init it in `signAccountOp` and call it whenever necessary.

This is the initial step to make `signAccountOp` instantiable from within `swapAndBridge.ts`


src/controllers/estimation/estimation.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  // it's initialized if it has estimated at least once
+  isInitialized() {
+    return this.hasEstimated
+  }
+
+  // has it estimated at least once without a failure
+  isLoadingOrFailed() {
+    return !this.hasEstimated || !!this.error
+  }
```
PetromirDev, line 213: Calling it `isLoadingOrFailed` is a bit confusing because it returns false when the estimation is loading (after the first load)

src/controllers/gasPrice/gasPrice.ts
```diff
@@ -0,0 +1,114 @@
...
+        speeds: bundlerGas,
+        bundler: bundler.getName()
+      }
+
+    this.emitUpdate()
+
+    if (!this.gasPriceTimeout)
+      this.gasPriceTimeout = createRecurringTimeout(() => this.fetch('major'), 12000)
+    this.gasPriceTimeout.stop()
+    this.gasPriceTimeout.start()
```
PetromirDev, line 108: Why do you call stop before start? Also, shouldn't start be called in an else statement?

if (!this.gasPriceTimeout) {
...
} else {
this.gasPriceTimeout.start()
}

src/controllers/gasPrice/gasPrice.ts
```diff
@@ -0,0 +1,114 @@
...
+
+        if (type === ErrorType.ConnectivityError) {
+          message = 'Network connection issue prevented us from retrieving the current network fee.'
+        }
+
+        this.emitError({
+          level: emitLevelOnFailure,
+          message,
+          error: new Error(`Failed to fetch gas price on ${this.#network.name}: ${e?.message}`)
+        })
```
PetromirDev, line 76: Does `emitError` work for these nested controllers? Where is the listener?

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -259,10 +265,61 @@ export class SignAccountOpController extends EventEmitter {
...
+      estimation: this.estimation,
+      readyToSign: this.readyToSign
+    }))
+    this.#traceCall = traceCall
+
+    this.#load()
+  }
+
+  #load() {
+    this.learnTokensFromCalls()
```
PetromirDev, line 289: Nice

src/controllers/estimation/estimation.ts
```diff
@@ -0,0 +1,267 @@
...
+
+      return errors
+    }
+
+    if (!this.isInitialized()) return []
+
+    if (this.error) {
+      errors.push({
+        title: this.error.message,
+        code: 'ESTIMATION_ERROR'
```
PetromirDev, line 253: The code should be `typeof cause === 'string' && cause.length ? cause : 'ESTIMATION_ERROR'`

src/controllers/estimation/estimation.ts
```diff
@@ -0,0 +1,267 @@
...
+    if (this.isLoadingOrFailed() && this.estimationRetryError) {
+      // If there is a successful estimation we should show this as a warning
+      // as the user can use the old estimation to broadcast
+      errors.push({
+        title: `${this.estimationRetryError.message} ${
+          this.error
+            ? 'We will continue retrying, but please check your internet connection.'
+            : 'Automatically retrying in a few seconds. Please wait...'
+        }`,
+        code: 'ESTIMATION_ERROR'
```
PetromirDev, line 242: Imo we shouldn't display a code in this case as, technically, there is no error yet

src/controllers/estimation/estimation.ts
```diff
@@ -0,0 +1,267 @@
...
+    if (this.estimationRetryError && this.status === EstimationStatus.Success) {
+      warnings.push({
+        id: 'estimation-retry',
+        title: this.estimationRetryError.message,
+        text: 'You can try to broadcast this transaction with the last successful estimation or wait for a new one. Retrying...',
+        promptBeforeSign: false
+      })
+    }
+
+    return warnings
```
PetromirDev, line 227: This warning is missing:
```
if (
      this.estimation?.bundlerEstimation?.nonFatalErrors?.find(
        (err) => err.cause === '4337_ESTIMATION'
      )
    ) {
      warnings.push({
        id: 'bundler-failure',
        title:
          'Smart account fee options are temporarily unavailable. You can pay fee with a Basic account or try again later',
        promptBeforeSign: false
      })
    }
```

and so is `'feeTokenPriceUnavailable'`

PetromirDev (changes requested):

PetromirDev:
When a new call is added, the simulation starts loading, but the skeleton is removed too early. Consequently, the old simulation is displayed briefly without a skeleton.
https://github.com/user-attachments/assets/e8c2b94a-9748-4de8-92fd-2b886cb0e694

PetromirDev:
Start a batch is enabled when there is an estimation error:
image

src/controllers/gasPrice/gasPrice.ts
Oxbobby: The easy answer here is that we're guaranteed to not have leeks and double timeouts this way. Performance wise I don't think we're losing something?

src/controllers/estimation/estimation.ts
Oxbobby: I changed this logic and I will test it again

src/controllers/estimation/estimation.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  // this is mainly for the bundler switcher but in general
+  // if the estimation wants to know the status of the outside
+  // controller, this is the function to set up intiially
+  #getOutsideControllerStatus: Function = () => {}
+
+  // this is mainly for the bundler switcher but in general
+  // if the estimation wants to know on which statuses in should
+  // disregard updates, this is the place
+  #outsideControllerNoUpdateStatuses: any[] = []
```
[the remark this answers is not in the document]
Oxbobby: I simplified this greatly by making everything into a single function and making the estimation controller accept the bundler switcher as an constructor argument instead of initializing it by itself.  
You can check it again and if it's not good enough, let's think together how we can improve this

src/controllers/gasPrice/gasPrice.ts
PetromirDev: What I mean is that the start function calls stop() when invoked:
```
  const start = () => {
    if (timeoutId) stop()

    timeoutId = setTimeout(async () => {
      await fn()
      start()
    }, timeout)
  }
```

PR #1378, opened 2025-03-19
Update/ Resolve selected account portfolio loading ux issues

PetromirDev:
Handle automatic and manual portfolio reloading properly in the selected account controller and library.


Ivshti (approved):
seems OK.

I thought about documenting changes in loading logic a human/developer style text, such that we don't forget why we made certain changes and we don't go in circles

however I decided it's not needed at the time since we're working off user feedback and that's the better way to do it rather than set decisions in stone

PR #1379, opened 2025-03-19
Fix: debug_traceCall fails on every first call on each network

JIOjosBG:
When i open the wallet and try to do a swap on base, on my first attempt debug trace call will not work, then on second, third... it works, When switching to optimism, for example, the same happens, the first attempt fails, then succeeds. 

On first attempts we still do not have gasPrice data and we pass undefined instead of array, later resulting in attempting to access `.find` of undefined.

- setting the gasPrice to 1 fails on some chains
- awaiting gasPrices before starting simulation will be slow

instead we simply do not pass gas data to debug_traceCall and leave the rpc t o figure it out


Ivshti (approved):
awesome fix

JIOjosBG:
> makes sense to me, but was there any reason for passing this in the first place?

The very first commit that introduced debug trace call in the extension just had it and we went along with it.

PR #1385, opened 2025-03-21
Add: 7702 activator gas

borislav-itskov:
Add the 7702 activator gas to the initial estimation to make sure the paymaster doesn't complain for low payment IF the account has not become a 7702 account, yet


Ivshti (commented):
lgtm

PR #1398, opened 2025-03-28
Feature / New Onboarding Flow

sonytooo:
Description: WIP


superKalo (approved):
I'm pre-approving this. Based on my tests - it reached a stable state 💪 

Last thing before merging: either bypass failing unit tests or sync them with the latest changes.

Bug with stored seed we discussed - HD path doesn't update, in case user changes it  (the prev `changeSavedSeedHdPathTemplateIfNeeded` logic is missing). But this can be fixed in a follow up PR as well.

PR #1415, opened 2025-04-11
Fix/faster next broadcast

Oxbobby:
Goal:
* don't wait for broadcast to confirm in order to be able to broadcast again

How:
* we resolve the user request before the final txnId (confirmation) is found but we resolve the dapp requests after the final, confirmed txnId is found
* we scan for nonce discrepancy issues and try to counter them
* fetch the newest account nonce wherever possible

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -217,8 +217,6 @@ export class MainController extends EventEmitter {
 
   #signAccountOpSigningPromise?: Promise<AccountOp | void | null>
 
-  #signAccountOpBroadcastPromise?: Promise<SubmittedAccountOp>
-
   #traceCallTimeoutId: ReturnType<typeof setTimeout> | null = null
 
   constructor({
@@ -686,17 +684,7 @@ export class MainController extends EventEmitter {
     // Error handling on the prev step will notify the user, it's fine to return here
     if (this.signAccountOp?.status?.type !== SigningStatus.Done) return
 
-    return this.withStatus(
-      'broadcastSignedAccountOp',
-      async () => {
-        // Reset the promise in the `finally` block to ensure it doesn't remain unresolved if an error is thrown
-        this.#signAccountOpBroadcastPromise = this.#broadcastSignedAccountOp().finally(() => {
-          this.#signAccountOpBroadcastPromise = undefined
-        })
-        return this.#signAccountOpBroadcastPromise
-      },
-      true
-    )
+    this.#broadcastSignedAccountOp()
   }
 
   destroySignAccOp() {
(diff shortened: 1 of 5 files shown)


src/controllers/main/main.ts
```diff
@@ -2451,6 +2449,15 @@ export class MainController extends EventEmitter {
       } else if (originalMessage.includes('Failed to fetch') && isRelayer) {
         message =
           'Currently, the Ambire relayer seems to be down. Please try again a few moments later or broadcast with a Basic Account'
+      } else if (originalMessage.includes('user nonce') && isRelayer) {
```
Ivshti, line 2452: i like this, good idea

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -641,6 +641,27 @@ export class SignAccountOpController extends EventEmitter {
...
+    // refetch the account state, resimulate and put the correct nonce in accountOp
+    const portfolioState = this.#portfolio.getPendingPortfolioState(this.accountOp.accountAddr)
+    const pendingPortfolioState = portfolioState
+      ? portfolioState[this.accountOp.chainId.toString()]
+      : null
+    if (
+      this.estimation.status === EstimationStatus.Success &&
+      pendingPortfolioState &&
+      pendingPortfolioState.criticalError?.simulationErrorMsg &&
+      pendingPortfolioState.criticalError?.simulationErrorMsg.indexOf('nonce did not increment') !==
```
Ivshti, line 654: i am worried that this error message may be different depending on the RPC (actually where does it come from?)

Ivshti: but there's not much we can do about this now anyway

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -1433,6 +1458,14 @@ export class SignAccountOpController extends EventEmitter {
       }
     }
 
+    const errorResponse = response as PaymasterErrorReponse
+    if (errorResponse.message.indexOf('invalid account nonce') !== -1) {
```
Ivshti, line 1462: same comment about error string applies

Ivshti (approved):
cool

src/controllers/signAccountOp/signAccountOp.ts
Oxbobby: We're setting this in `getOnChainBalances`:
```
  if (simulationOps.length && afterNonce === beforeNonce)
    throw new SimulationError(
      'Account op passed for simulation but the nonce did not increment. Perhaps wrong nonce set in Account op',
      beforeNonce,
      afterNonce
    )
```
a `SimulationError` records `simulationErrorMsg` in its constructor:
```
constructor(message: string, beforeNonce: bigint, afterNonce: bigint) {
    super(`simulation error: ${message}`)
    this.simulationErrorMsg = message
}
```
and the try/catch is in the portfolio controller:
```
catch (e: any) {
    ...
      state.criticalError = {
        message: e?.message || 'Error while executing the get function in the portfolio library.',
        simulationErrorMsg: e?.simulationErrorMsg,
        stack: e?.stack,
        name: e?.name
      }
```

In other words, we have `simulationErrorMsg` persisted into the state when the simulation detects a nonce discrepancy

PR #1430, opened 2025-04-24
Fix/ The action window is resized on focus (Firefox)

PetromirDev:
Change needed after https://github.com/AmbireTech/ambire-app/pull/4541


PetromirDev:
The changes introduce other problems

PR #1433, opened 2025-04-24
Fix/ App connection and remove legends.ambire.com from dapp list

PetromirDev:
## Changes:
- Reverts one change from https://github.com/AmbireTech/ambire-common/commit/b819801e301b44820bbc3685913a728145c86227 which fixes app connections
- Removes legends.ambire.com from the storage


superKalo (approved):
Tested on my local instance + connecting and disconnecting other than Ambire Rewards dapps, all worked like a charm 👍

PR #1438, opened 2025-04-25
Fix/ Portfolio batching test

PetromirDev:


PetromirDev:
The test was fixed in another way

PR #1439, opened 2025-04-28
UX/ switch any/any token and switch amounts properly (Swap and Bridge)

PetromirDev:
## Changes:
- Change: Allows the user to switch tokens, no matter if there is a toToken or not, and if the toToken is in the user's portfolio
- Fix: Switch token amounts properly


src/interfaces/swapAndBridge.ts
```diff
@@ -1,3 +1,5 @@
+import { TokenResult } from 'libs/portfolio'
```
superKalo, line 1: Opsi, bad import.

superKalo (approved):

PR #1451, opened 2025-05-05
Performance: sooner debug trace call

Oxbobby:
Put debug trace call before simulation/estimation for better performance: it gets batched along with the simulation/estimation rpc calls.  
We didn't do this before because we were working at the assumption that gasPrice / gasLimit was needed for debug trace call to work which is not true.  
A little bit more testing is needed before merging this as it actually might be a problem with some RPCs if they don't support batching of debug_traceCall along other requests


Ivshti (approved):
fantastic change. I can't think of any way it can go wrong from the code, but we'll certainly test

PR #1454, opened 2025-05-05
Config / Resolve TypeScript compiler errors

superKalo:
Missing `outDir` was causing errors on my Mac (VS Code) for some reason (although I read that this should fallback gracefully https://www.typescriptlang.org/tsconfig/#outDir)

Screenshot 2025-05-05 at 18 36 40

Also fixes a minor inconsistency in package-lock v.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ambire-common",
-  "version": "2.47.1",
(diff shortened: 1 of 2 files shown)


PetromirDev (approved):
Awesome!
Before:
image
After:
image

PR #1463, opened 2025-05-08
Feature / Associate Account Keys with the Legacy Saved Seed

sonytooo:
* accounts/keys added from the single saved seed with the legacy format had no association with that saved seed. This PR adds a migration to map all previously added accounts to the legacy saved seed.


superKalo (approved):
On the latest commit, tested against my local instance, all worked 👍

PR #1473, opened 2025-05-13
Improve / Keystore Encrypt and Decrypt Performance on Firefox

sonytooo:
* Improves the performance of unlocking and adding secrets to the Keystore on Firefox.

Adds scrypt from **@noble/hashes**—a **package with 10× more downloads than scrypt-js**—which **delivers 6–7× faster performance on Gecko-based browsers** (like Firefox), but performs worse on Webkit (e.g., Chrome).
To handle this, we introduce a ScryptAdapter library that selects the optimal implementation based on the user’s platform.


src/controllers/keystore/keystore.ts
```diff
@@ -2,15 +2,15 @@
...
 import {
   decryptWithPrivateKey,
   Encrypted,
   encryptWithPublicKey,
   publicKeyByPrivateKey
 } from 'eth-crypto'
 import { concat, getBytes, hexlify, keccak256, Mnemonic, toUtf8Bytes, Wallet } from 'ethers'
-import scrypt from 'scrypt-js'
+// import { entropyToMnemonic } from 'bip39'
+import { Platform } from 'interfaces/platform'
```
PetromirDev, line 13: fix import

PetromirDev (changes requested):
Crazy improvement :clap: 
image

superKalo (approved):
Great efforts 👏 

The @noble/hashes npm package sounds legit, claims to be audited, widely-used and MIT-licensed.

PR #1476, opened 2025-05-14
Fix / loss of keys after execution of the legacy saved seed migration

sonytooo:
There is a bug introduced with the migration that associates keys to the legacy saved seed. After running the migration for each account only a single key will remain due to a problem in the mapping where the keys are mapped by addr only and not by addr and type resulting to an array with unique keys by addr at the end.

Affected users are those that had a legacy saved seed and at the same time account/s with multiple signers

**This PR fixes that so all future execution of the migration will keep the same number of keys in the storage**


superKalo (approved):
I tested by doing the following:

1. remove extension
2. checkout ambire-app: v4.59.2 and ambire-common: v2.49.2
3. import seed + same acc from 2 diff HW
4. stop the server
5. checkout ambire-app: v2 and ambire-common: fix/legacy-seed-phrase-migration-bug

Migration succeeded, all accounts and keys look fine 👍

PR #1489, opened 2025-05-20
Fix / AccountPicker AddAccounts Bug

sonytooo:
**Added logic to prevent the accountPicker from resetting while addAccounts is in progress.**
---
Тhis fixes an issue where accounts added from a seed, pk or HW could end up in a view-only state. This is possible because we reset the accountPicker if the current screen is not one of: account-personalize or account-picker. Doing this during the accounts and keys import process, leads to a state where the accounts are added first but their corresponding keys are not because the accountPicker state is empty due to the reset.


superKalo (approved):
Great catch 👍

PR #1510, opened 2025-05-29
Config / Warn for significant account balance decrease when the decrease is more than 20% of the total account balance AND more than $1k

superKalo:
Closes https://github.com/AmbireTech/ambire-app/issues/4739


gergana95 (approved):
Tested and works good 🙌

PR #1535, opened 2025-06-11
UX/ Swap state consistency, UI responsiveness and error handling

PetromirDev:
### Old promises affect the new state:
We run a bunch of promises when fetching the quote/estimating the route. If the changes the properties of a swap while it's being fetched/estimated, the new swap SHOULD NOT be affected by the old promises. This is done using multiple techniques:
1. There are two `signAccountOpController` variables - a private one that may be outdated and a public one that is always up to date. This is done to prevent the split second in which there is a quote and the old `signAccountOp` estimation's status is `success` - during this split-second, the status of the swap and bridge controller is `ReadyToSubmit`, while it should be `ReadyToEstimate`
2. `updateQuoteId` is checked during estimation, too - this fixes another split-second window caused by the quote fetch debounce
3. Other minor changes you can see in the diff
Before:
image
### No error when the user hits rate limits
After:
image
### Unsupported token error not humanised

Before:
image
After:
image

### DIFF
--- a/src/services/lifi/api.ts
+++ b/src/services/lifi/api.ts
@@ -1,10 +1,10 @@
 import {
   ExtendedChain as LiFiExtendedChain,
-  Step as LiFiIncludedStep,
+  LiFiStep,
   Route as LiFiRoute,
   RoutesResponse as LiFiRoutesResponse,
   StatusResponse as LiFiRouteStatusResponse,
-  LiFiStep,
+  Step as LiFiIncludedStep,
   Token as LiFiToken,
   TokensResponse as LiFiTokensResponse
 } from '@lifi/types'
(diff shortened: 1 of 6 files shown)


Oxbobby (approved):

src/services/lifi/api.ts
```diff
@@ -271,8 +272,9 @@ export class LiFiAPI {
     }
 
     if (response.status === 429) {
-      const error = `Our service provider received too many requests, temporarily preventing your request from being processed. ${errorPrefix}`
-      throw new SwapAndBridgeProviderApiError(error)
+      const error =
+        'Our service provider received too many requests, temporarily preventing your request from being processed.'
+      throw new SwapAndBridgeProviderApiError(error, 'Too many requests')
```
superKalo, line 277: The short note is too generic. You want to keep it concise so that in the UI it is on 1 line, right?

If so, I'd suggest: "Rate limit reached, try again in a while" maybe?

PR #1537, opened 2025-06-12
Fix: Celo chain LiFi

Oxbobby:
Fix: LiFi doesn't work on Celo with the chain's native token as it expects another address than addr 0. So we map it

### DIFF
--- a/src/libs/swapAndBridge/swapAndBridge.ts
+++ b/src/libs/swapAndBridge/swapAndBridge.ts
@@ -357,6 +357,11 @@ const lifiMapTokenAddr = (chainId: number, tokenAddr: string) => {
   return '0x471EcE3750Da237f93B8E339c536989b8978a438'
 }
 
+const lifiTokenListFilter = (t: SwapAndBridgeToToken) => {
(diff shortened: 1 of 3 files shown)


src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -346,11 +346,23 @@ const addCustomTokensIfNeeded = ({
...
+// the celo native token is at an address 0x471EcE3750Da237f93B8E339c536989b8978a438
+// and LiFi doesn't work if we pass address 0 for this. We map it only for
+// lifi to make the swap work in this case
+const getLiFiTokenAddress = (chainId: number, tokenAddr: string) => {
+  if (tokenAddr !== ZeroAddress) return tokenAddr
+  // celo chain
+  if (chainId !== 42220) return tokenAddr
+
+  return '0x471EcE3750Da237f93B8E339c536989b8978a438'
+}
```
PetromirDev, line 358: Please make the comment public so it's displayed when the function is hovered:
image

PetromirDev (changes requested):
the rest of the feedback is in dm

PetromirDev (approved):

PR #1541, opened 2025-06-13
Feature / Debank DeFi

sonytooo:
* Integrated Debank DeFi positions while preserving our custom DeFi protocol implementations on top of the Debank list
* Added Ambire stkWallet DeFi position for users holding a positive stkWallet token balance
* Improved performance of the updatePositions function
* Adjusted DeFi types for better consistency and accuracy

### DIFF
--- a/src/controllers/defiPositions/defiPositions.ts
+++ b/src/controllers/defiPositions/defiPositions.ts
@@ -278,34 +278,22 @@ export class DefiPositionsController extends EventEmitter {
 
     prepareNetworks()
 
+    let debankPositions: PositionsByProvider[] = []
+
     try {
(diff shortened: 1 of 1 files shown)


src/controllers/defiPositions/defiPositions.ts
```diff
@@ -113,120 +115,217 @@ export class DefiPositionsController extends EventEmitter {
...
+        isLoading: false,
+        positionsByProvider: Array.from(positionMap.values()),
+        updatedAt: Date.now()
+      }
+    }
+
+    prepareNetworks()
+
+    try {
+      const resp = await this.#fetch(`https://cena.ambire.com/api/v3/defi/${selectedAccountAddr}`)
```
superKalo, line 282: Update: problem confirmed (chatting offline with Simo), if this fails, it should fallback to the custom defi positions, but it doesn't.

PR #1545, opened 2025-06-17
Feature / Force Update DeFi Positions

sonytooo:
Add logic to the DeFi controller to force-refresh DeFi positions when the user manually reloads their balance from the dashboard. The force update should be triggered only if all the following conditions are met:
	•	The user clicks the refresh button on the dashboard
	•	The account is not view-only
	•	The DeFi tab is currently focused on the dashboard


superKalo (approved):
Sounds good. If we want to further prevent abuse, I'd suggest we add a timestamp with the last force update requested and to trigger another only if it's 30s later? In case someone clicks reload many times in a row.

PR #1548, opened 2025-06-18
Fixes / Dapps

sonytooo:
* Added dapps getter one more filtration condition by name to ensure only unique dapps are returned

### DIFF
--- a/src/controllers/dapps/dapps.ts
+++ b/src/controllers/dapps/dapps.ts
@@ -49,8 +49,13 @@ export class DappsController extends EventEmitter {
       })
     )
 
-    return [...this.#dapps, ...predefinedDappsParsed].reduce((acc: Dapp[], curr: Dapp): Dapp[] => {
-      if (!acc.some(({ url }) => url === curr.url)) return [...acc, curr]
+    const combined = [...this.#dapps, ...predefinedDappsParsed]
+
+    return combined.reduce((acc: Dapp[], curr: Dapp): Dapp[] => {
+      if (!acc.some((dapp) => dapp.url === curr.url || dapp.name === curr.name)) {
+        acc.push(curr)
+      }
+
       return acc
     }, [])
   }


src/controllers/dapps/dapps.ts
```diff
@@ -72,78 +84,72 @@ export class DappsController extends EventEmitter {
...
+    id?: string,
     skipPermissionCheck?: boolean
   ) => {
     await this.initialLoadPromise
 
     let dappSessions: { key: string; data: Session }[] = []
     Object.keys(this.dappSessions).forEach((key) => {
       const hasPermissionToBroadcast =
-        skipPermissionCheck || this.hasPermission(this.dappSessions[key].origin)
+        skipPermissionCheck || this.hasPermission(this.dappSessions[key].id)
```
PetromirDev, line 146: rename key to sessionId

src/utils/getDomainFromUrl.ts
```diff
@@ -0,0 +1,21 @@
...
+  const domain = parse(url).domain
+
+  if (domain) return domain
+
+  try {
+    const hostname = new URL(url).hostname
+    const cleaned = hostname.startsWith('www.') ? hostname.slice(4) : hostname
+    const parts = cleaned.split('.')
+
+    if (parts.length >= 2) return parts.slice(-2).join('.')
```
PetromirDev, line 13: What if there is a subdomain or the tld is multi-level?
`sigtool.ambire.com`
`somedomain.co.uk`

image

PetromirDev (changes requested):

src/utils/getDomainFromUrl.ts
sonytooo: removed the fallback logic because from what I've researched tldts is very reliable and consistent

src/interfaces/dapp.ts
```diff
@@ -14,7 +15,7 @@ export interface Dapp {
 export interface DappProviderRequest {
   method: string
   params?: any
-  session: { name: string; origin: string; icon: string }
+  session: { id: string; origin: string; name: string; icon: string }
```
PetromirDev, line 18: Update the session type in `interfaces/userRequest` row 66 and 84

PetromirDev (changes requested):

PR #1572, opened 2025-07-01
UX/ Lifi api minor improvements

PetromirDev:
## Changes:
- Timeout API requests after 5 seconds. I tested swap and bridge with network throttling set to Regular 3G, and 5 seconds were enough even then
- Updates the non-JSON response error message. The previous message was way too long and confusing for non-devs

The old message:
image

Potentially fixes https://github.com/AmbireTech/ambire-app/issues/4964, tho I'm not sure so let's not close it for now

### DIFF
--- a/src/services/lifi/api.ts
+++ b/src/services/lifi/api.ts
@@ -281,7 +281,7 @@ export class LiFiAPI {
           setTimeout(() => {
             reject(
               new SwapAndBridgeProviderApiError(
-                'Our service provider is temporarily unavailable. Error details: Request timeout out'
+                'Our service provider is temporarily unavailable or your internet connection is too slow. Error details: Request timeout out'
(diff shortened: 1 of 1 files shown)


src/services/lifi/api.ts
```diff
@@ -291,8 +304,8 @@ export class LiFiAPI {
     try {
       responseBody = await response.json()
     } catch (e: any) {
-      const message = e?.message || 'no message'
-      const error = `${errorPrefix} Error details: <Unexpected non-JSON response from our service provider>, message: <${message}>`
+      const error =
+        'Our service provider is temporarily unavailable. Error details: Non-JSON response received.'
```
Oxbobby, line 308: Let's remove "Error details: Non-JSON response received."

PR #1591, opened 2025-07-08
Feature / Ability to re-arrange accounts

Maskln:
Paired with ambire-app PR: https://github.com/AmbireTech/ambire-app/pull/5042

### DIFF
--- a/src/controllers/accounts/accounts.ts
+++ b/src/controllers/accounts/accounts.ts
@@ -224,6 +224,18 @@ export class AccountsController extends EventEmitter {
     this.emitUpdate()
   }
 
+  async reorderAccounts(reordered: Account[]) {
+    const reorderedAddrSet = new Set(reordered.map((acc) => getAddress(acc.addr)))
+
+    const remainingAccounts = this.accounts.filter(
+      (acc) => !reorderedAddrSet.has(getAddress(acc.addr))
+    )
+
+    this.accounts = getUniqueAccountsArray([...reordered, ...remainingAccounts])
+    await this.#storage.set('accounts', this.accounts)
+    this.emitUpdate()
+  }
+
   get areAccountStatesLoading() {
     return Object.values(this.accountStatesLoadingState).some((isLoading) => isLoading)
   }
--- a/src/controllers/accounts/accounts.ts
+++ b/src/controllers/accounts/accounts.ts
@@ -232,8 +232,8 @@ export class AccountsController extends EventEmitter {
     )
 
     this.accounts = getUniqueAccountsArray([...reordered, ...remainingAccounts])
-    await this.#storage.set('accounts', this.accounts)
     this.emitUpdate()
+    await this.#storage.set('accounts', this.accounts)
   }
 
   get areAccountStatesLoading() {


src/controllers/accounts/accounts.ts
```diff
@@ -216,6 +216,18 @@ export class AccountsController extends EventEmitter {
...
+    const reorderedAddrSet = new Set(reordered.map((acc) => getAddress(acc.addr)))
+
+    const remainingAccounts = this.accounts.filter(
+      (acc) => !reorderedAddrSet.has(getAddress(acc.addr))
+    )
+
+    this.accounts = getUniqueAccountsArray([...reordered, ...remainingAccounts])
+    this.emitUpdate()
+    await this.#storage.set('accounts', this.accounts)
+  }
```
PetromirDev, line 229: I don't think it's necessary to call `getAddress` for accounts that are stored in the extension- their addresses should already be checksummed and even if they are not for whatever reason they will remain lowercase and won't affect the order.

src/controllers/accounts/accounts.ts
```diff
@@ -216,6 +216,18 @@ export class AccountsController extends EventEmitter {
...
   }
 
+  async reorderAccounts(reordered: Account[]) {
+    const reorderedAddrSet = new Set(reordered.map((acc) => getAddress(acc.addr)))
+
+    const remainingAccounts = this.accounts.filter(
+      (acc) => !reorderedAddrSet.has(getAddress(acc.addr))
+    )
+
+    this.accounts = getUniqueAccountsArray([...reordered, ...remainingAccounts])
```
PetromirDev, line 226: As far as I understand this code serves as guarantee that there will be no removed accounts if `reorderAccounts` is not called with all accounts. What concerns me is that we spread `remainingAccounts` after `reordered`. Let's discuss this offline

PetromirDev (changes requested):

src/controllers/accounts/accounts.ts
```diff
@@ -216,6 +216,18 @@ export class AccountsController extends EventEmitter {
     await this.#storage.set('accounts', this.accounts)
   }
 
+  reorderAccounts(reordered: Account[]) {
+    if (!reordered.length) {
+      throw new Error('Reordered accounts array is empty')
+    }
```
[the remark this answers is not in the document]
PetromirDev: Currently, this error will bubble to the background process and be humanised to `Something went wrong! Please contact support. Error:...`

src/controllers/accounts/accounts.ts
```diff
@@ -248,6 +248,32 @@ export class AccountsController extends EventEmitter {
...
+      fromIndex < 0 ||
+      toIndex < 0 ||
+      fromIndex >= this.accounts.length ||
+      toIndex >= this.accounts.length
+    ) {
+      return this.emitError({
+        level: 'minor',
+        message: `Invalid fromIndex (${fromIndex}) or toIndex (${toIndex})`,
+        error: new Error('Invalid fromIndex or toIndex')
+      })
```
PetromirDev, line 262: Let's make this error major so that we can display an error toast to the user. Make the current message the message of the error and modify the message so it's human-readable - something along the lines of: "Failed to reorder accounts. Please reload the page and try again"

PetromirDev (changes requested):

PR #1592, opened 2025-07-09
Config / Activate api key for our Swap & Bridge service provider (LiFi) when we hit the rate limit #5055

Oxbobby:
We set the api key for 2 hours when we encounter a 429.  
We try to remove it on each new request if 2 hours have passed


src/services/lifi/api.ts
```diff
@@ -233,13 +237,26 @@ export class LiFiAPI {
...
+  removeApiKey() {
+    if (!this.#apiKey || !this.#apiKeyActivatedTimestamp) return
+
+    const twoHoursPassed = Date.now() - this.#apiKeyActivatedTimestamp >= 120 * 60 * 1000
+    if (!twoHoursPassed) return
+
+    this.#apiKey = undefined
+    this.#apiKeyActivatedTimestamp = undefined
+  }
+
```
PetromirDev, line 259: This method doesn't remove the api key from `#headers`, meaning that if there is a rate limit once, the API key will remain until the user restarts their extension. Additionally, I don't see the point in resetting `this.#apiKey` as it will be missing for subsequent API failures. Imo we should never reset `this.#apiKey`, but instead remove the key from `this.#headers` and reset the timestamp

PetromirDev (changes requested):

src/services/lifi/api.ts
Oxbobby: should've been this.headers, it's a mistake

src/services/lifi/api.ts
```diff
@@ -225,6 +225,10 @@ export class LiFiAPI {
 
   isHealthy: boolean | null = null
 
+  #apiKey?: string
```
superKalo, line 228: We can simplify this a bit I think. How about:

There is a logic in the background that does:

```
// Temporarily use NO API key in production, until we have a middleware to handle the API key
swapApiKey: isProd ? undefined : LI_FI_API_KEY,
```

So this is now obsolete. How about if we change this to always pass the `swapApiKey`.

Then in here, just a boolean flag: `#shouldSendApiKeyHeader` or something with a comment why this is needed to be send only in some scenarios?

Oxbobby: Added a comment to #apiKeyActivatedTimestamp :)

superKalo (approved):

PR #1604, opened 2025-07-17
Bug / Extension crash caused by missing `accountOp` in case of Sign Account Op controller error

superKalo:
Ref: https://monitor.ambire.com/organizations/ambire/issues/78

### DIFF
--- a/src/controllers/transfer/transfer.ts
+++ b/src/controllers/transfer/transfer.ts
@@ -635,7 +635,9 @@ export class TransferController extends EventEmitter {
       this.emitUpdate()
     })
     this.signAccountOpController.onError((error) => {
-      this.#portfolio.overridePendingResults(this.signAccountOpController!.accountOp)
+      if (this.signAccountOpController?.accountOp)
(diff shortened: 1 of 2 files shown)


src/controllers/transfer/transfer.ts
```diff
@@ -635,7 +635,9 @@ export class TransferController extends EventEmitter {
       this.emitUpdate()
     })
     this.signAccountOpController.onError((error) => {
-      this.#portfolio.overridePendingResults(this.signAccountOpController!.accountOp)
+      if (this.signAccountOpController?.accountOp)
```
Oxbobby, line 638: if statement should be:
```
if (this.signAccountOpController)
```
as accountOp is always true if signAccountOpController is initialized

superKalo: Thanks, you're right, improved in https://github.com/AmbireTech/ambire-common/pull/1604/commits/b4c7a116942c3fc321f31457eee2186dbe9e8594

PR #1614, opened 2025-07-21
Update/ Portfolio price caching

PetromirDev:
## Changes:
- Changes price recency from 1 min to 5 min
- Change: Don't display an error banner on price fetch if there is a cached price 

The aim is to reduce the number of calls to cena and display error banners less frequently.


gergana95 (approved):
Tested and works well!

PR #1626, opened 2025-07-25
Fix / Transfer - Skip tracking on close

jordan-enev:
If the Transfer popup is closed during signing or broadcasting, we don't want to show the tracking screen next time. Previously, the last account op was kept and shown, even after closing mid-process (mainCtrl.#broadcastSignedAccountOp). With shouldTrackLatestBroadcastedAccountOp, we can now prevent this and show a fresh form instead.

Linked PR: https://github.com/AmbireTech/ambire-app/pull/5166

### DIFF
--- a/src/controllers/transfer/transfer.ts
+++ b/src/controllers/transfer/transfer.ts
@@ -201,7 +201,6 @@ export class TransferController extends EventEmitter {
 
   set shouldTrackLatestBroadcastedAccountOp(value: boolean) {
     this.#shouldTrackLatestBroadcastedAccountOp = value
-    this.emitUpdate()
   }
 
   // every time when updating selectedToken update the amount and maxAmount of the form
@@ -266,7 +265,6 @@ export class TransferController extends EventEmitter {
   }
 
   resetForm(shouldDestroyAccountOp = true) {
-    console.log('Transfer -> resetForm() called')
     this.selectedToken = null
     this.amount = ''
     this.amountInFiat = ''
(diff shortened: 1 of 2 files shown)


src/controllers/transfer/transfer.ts
```diff
@@ -193,6 +195,15 @@ export class TransferController extends EventEmitter {
...
   }
 
+  get shouldTrackLatestBroadcastedAccountOp() {
+    return this.#shouldTrackLatestBroadcastedAccountOp
+  }
+
+  set shouldTrackLatestBroadcastedAccountOp(value: boolean) {
+    this.#shouldTrackLatestBroadcastedAccountOp = value
+    this.emitUpdate()
+  }
```
PetromirDev, line 205: This is not a commonly used pattern and can result in unnecessary emit updates if someone doesn't know that there is a setter. Also, this results in an unnecessary emit update in `update()` every time it's called. Finally, an update isn't needed in `handleCleanUpOnPortDisconnect` as opening the popup again will get the entire state from the controller.

src/controllers/transfer/transfer.ts
```diff
@@ -254,7 +265,8 @@ export class TransferController extends EventEmitter {
     )
   }
 
-  resetForm() {
+  resetForm(shouldDestroyAccountOp = true) {
+    console.log('Transfer -> resetForm() called')
```
PetromirDev, line 269: are we going to keep this log, or is it forgotten?

PetromirDev (changes requested):

PR #1627, opened 2025-07-25
Refactor/ Portfolio lib hints and prices logic

PetromirDev:
## Changes:
- Adjust the logic in the portfolio lib and controller so tokens can be filtered in the lib. This is needed so we don't fetch prices and map 0 value tokens that aren't needed
- Remove tokens from `toBeLearned` after learning them
- Makes the lib return a new property- `toBeLearned` that includes tokens that have to be learned by the controller. These tokens can be non-zero `toBeLearned` tokens passed from the controller or tokens on networks that don't support velcro discovery or have limited support. (On some networks, velcro returns a list of the top 100-200 tokens by marketcap, and as this list may change in the future, we have to learn non-zero tokens from it)

## Results:
Testing with `0xd8F3676A7567842295170B3f6AE4Bd466A878374` and filtering the network requests by URL `https://cena.ambire.com/api/v3/simple/`, we reduced the network requests for prices from ~120 to ~50

### DIFF
--- a/src/controllers/portfolio/portfolio.ts
+++ b/src/controllers/portfolio/portfolio.ts
@@ -845,11 +845,12 @@ export class PortfolioController extends EventEmitter {
   }
 
   addTokensToBeLearned(tokenAddresses: string[], chainId: bigint) {
-    if (!tokenAddresses.length) return false
-    if (!this.#toBeLearnedTokens[chainId.toString()])
-      this.#toBeLearnedTokens[chainId.toString()] = []
+    const chainIdString = chainId.toString()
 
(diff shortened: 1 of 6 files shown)


src/controllers/portfolio/portfolio.ts
```diff
@@ -785,7 +769,8 @@ export class PortfolioController extends EventEmitter {
                       state
                     }
                   }),
-                ...allHints
+                ...allHints,
+                specialErc20Hints
```
jordan-enev, line 773: https://github.com/AmbireTech/ambire-common/pull/1627/files#r2236366531

jordan-enev:
for some reason, portfolio balance on this PR and on Prod differs, let's investigate 
image_720
image_720

PR #1644, opened 2025-08-01
Feature / Testnet networks config

sonytooo:
To enable testnet-only mode, pass `defaultNetworksMode = 'testnet'` when constructing the `NetworksController` in the MainController. On a fresh installation of the extension, the `testnetNetworks` constants will be used to initialize the `NetworksController`. Adding custom networks remains possible in testnet mode, as no network filtering is applied. However, dynamic network updates are disabled in this mode.

### DIFF
--- a/src/controllers/networks/networks.ts
+++ b/src/controllers/networks/networks.ts
@@ -32,6 +32,9 @@ const STATUS_WRAPPED_METHODS = {
  * for adding, updating, and removing networks.
  */
 export class NetworksController extends EventEmitter {
+  // To enable testnet-only mode, pass defaultNetworksMode = 'testnet' when constructing the NetworksController in the MainController.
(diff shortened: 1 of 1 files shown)


src/controllers/networks/networks.ts
```diff
@@ -26,6 +28,8 @@ const STATUS_WRAPPED_METHODS = {
  * for adding, updating, and removing networks.
  */
 export class NetworksController extends EventEmitter {
+  defaultNetworksMode: 'mainnet' | 'testnet' = 'mainnet'
```
superKalo, line 31: Please add a few notes with context why this is needed (basically, parts of the PR description can go here).

superKalo (changes requested):
Looks good! I'm only marking it as "request changes" for the third comment.

PR #1651, opened 2025-08-04
Fix: swap&bridge sign account op

Oxbobby:
In the swap&bridge controller, we have a signAccountOp getter. That getter's goal is to check on request if we're fetching a quote and if the quote is different than the one in the signAccountOp as time is needed to sync both places. If there were differences between both, signAccountOp resolved in null.

While the getter is helpful in certain cases (determining the real form status upon request), it was used in the quote fetch function. That meant that signAccountOp was always treated as null as the original signAccountOp's quote was different than the new one. This, on the other hand, lead to the creation of a new signAccountOp instance without destroying the old one from memory until a background refresh is requested


superKalo (approved):
Based on my understanding, I agree with these changes 👍 Great catch!

As a sum up, in this controller - the getter seems helpful only when determining the real form status upon request (in the get formStatus method).

I'll appreciate if @PetromirDev also takes a look before we merge this.

PR #1655, opened 2025-08-05
Initial token balances / nfts retry mechanism

Oxbobby:
Add: a retry mechanism in deployless to allow quick refetching of initial portfolio requests if they are failing (they should not on initial load)

---

Ran the failing tests locally, they pass


Oxbobby:
Closing as it might cause hiding issues instead as per internal discussion

PR #1656, opened 2025-08-05
Cut off bundler gas failures

Oxbobby:
Wait a max of 4s for the bundler gas prices in the gas price controller.  
Switch the bundler if fetching gas prices from it is slow (4s max)


Ivshti (commented):
lgtm

PR #1658, opened 2025-08-05
Fix / AAVE protocol console error

sonytooo:
Resolves:
Screenshot 2025-08-05 at 13 46 12
This resolves the issue by initializing providers for both enabled and disabled networks, instead of only initializing them for enabled networks. This will also help us for the development of the tokens discovery on disabled networks


superKalo (approved):
JsonRpcProvider instances in Ethers v6 don't have performance bottleneck (doesn’t open a connection or perform a handshake immediately), so I'm fine with this approach also.

PR #1662, opened 2025-08-05
SignAccOp: Disallow singing calls to self

JIOjosBG:
## Problem 
Since smart accounts support an account to call it's own multicall functions, attackers might request a call that is a multicall to the same account. Before we mitigated that vuln by humanizing the call. But since 7702 users might want to import account from other wallets that we have not implemented a humanizer for. Also requesting such calls makes no sense....

That why we will disable signing of calls that have `to == sender`. 

image

### DIFF
--- a/src/libs/humanizer/modules/embeddedAmbireOperationHumanizer/ambireOperation.test.ts
+++ b/src/libs/humanizer/modules/embeddedAmbireOperationHumanizer/ambireOperation.test.ts
@@ -2,13 +2,13 @@ import { Interface, ZeroAddress } from 'ethers'
 
 import { describe, it } from '@jest/globals'
 
+import { embeddedAmbireOperationHumanizer } from '.'
 import { AccountOp } from '../../../accountOp/accountOp'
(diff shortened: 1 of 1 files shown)


JIOjosBG:
Resolves: https://github.com/ambireTech/ambire-app/issues/5200

Oxbobby:
I will check this also, bookmarking

JIOjosBG:
> Those will not get humanized as the humanizer will receive a single call - executeBySender - without knowing there are legit subcalls in it that are the token approvals

we used to humanize such calls, but in this PR I am removing this functionality as the humanizer should not split multicalls 

> However, this attack is not legit for V1/V2 addresses, only for 7702 EOAs.

Also legit for V2 contracts as they also have `executeBySelf` [proof](https://www.codeslaw.app/contracts/ethereum/0x46C0C59591EbbD9b7994d10efF172bFB9325E240?file=src%2FAmbireAccount.sol&start=194&end=213)

Other than that, great summary

PR #1681, opened 2025-08-08
Update/ Activity banners

PetromirDev:
## Changes:
- Banners for confirmed account ops are automatically hidden when an account op is added/updated
- Implements some performance optimisations
- Fixes
- Adds test cases

Closes https://github.com/AmbireTech/ambire-app/issues/5206
Closes https://github.com/AmbireTech/ambire-app/issues/5259
Closes https://github.com/AmbireTech/ambire-app/issues/5199


gergana95 (approved):
I think its better this way, we agreed to remove the confirmed banner after a period of time. Tested and works good!

PR #1694, opened 2025-08-12
Add/ lifi timing optimisation

PetromirDev:
The changes are documented in the issue and code.
Issue https://github.com/AmbireTech/ambire-app/issues/5276


superKalo (approved):
I'm +1 for the "maximize results" recommendation by LiFi.

It would on theory be the best balance between actually receiving enough results + canceling the ones hanging (if any do hang). Let's give it a shot!

PR #1696, opened 2025-08-13
Feature / ContinuousUpdatesController

sonytooo:


src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -324,6 +331,39 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
...
+        return
+      }
+
+      await this.updateQuote({
+        skipPreviousQuoteRemoval: true,
+        skipQuoteUpdateOnSameValues: false,
+        skipStatusUpdate: false
+      })
+    }, UPDATE_SWAP_AND_BRIDGE_QUOTE_INTERVAL)
+
```
PetromirDev, line 347: sooo much better than before :clap:

src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -324,6 +331,39 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
...
+      this.#updateActiveRoutesInterval.updateTimeout({ timeout: minServiceTime })
+    }, UPDATE_SWAP_AND_BRIDGE_QUOTE_INTERVAL)
+
+    this.onUpdate(() => {
+      if (this.activeRoutesInProgress.length) {
+        this.#updateActiveRoutesInterval.start()
+      } else {
+        this.#updateActiveRoutesInterval.stop()
+      }
+    }, 'swap-and-bridge')
```
PetromirDev, line 366: Please do this in the setter `activeRoutes`. Make `activeRoutesInProgress` read from the private variable, update the private variable and then check 
```
    if (this.activeRoutesInProgress) {
      this.#updateActiveRoutesInterval.start()
    } else {
      this.#updateActiveRoutesInterval.stop()
    }
```

src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -324,6 +331,39 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
...
+    this.#updateActiveRoutesInterval = createRecurringTimeout(async () => {
+      if (!this.activeRoutesInProgress.length) {
+        this.#updateActiveRoutesInterval.stop()
+        return
+      }
+
+      await this.checkForNextUserTxForActiveRoutes()
+
+      const minServiceTime = getActiveRoutesLowestServiceTime(this.activeRoutesInProgress)
+      this.#updateActiveRoutesInterval.updateTimeout({ timeout: minServiceTime })
```
PetromirDev, line 357: This doesn't restart the timeout tho. Do we only want to update it?

src/controllers/continuousUpdates/continuousUpdates.ts
```diff
@@ -0,0 +1,206 @@
...
+  #retriedFastAccountStateReFetchForNetworks: string[] = []
+
+  constructor({ main }: { main: IMainController }) {
+    this.#main = main
+
+    // Postpone the portfolio update for the next interval
+    // if we have broadcasted but not yet confirmed acc op.
+    // Here's why:
+    // 1. On the Dashboard, we show a pending-to-be-confirmed token badge
+    //    if an acc op has been broadcasted but is still unconfirmed.
```
PetromirDev, line 48: Having logic that may throw an error in the constructor of a controller is not very good. If any of this code throws an error, it will bubble up to the constructor of the main controller and break the entire extension. Imo every timeout should be wrapped in try try-catch, and a crash report should be sent on error.

src/controllers/continuousUpdates/continuousUpdates.ts
```diff
@@ -0,0 +1,206 @@
...
+    })
+    this.#main.ui.uiEvent.on('removeView', () => {
+      if (!this.#main.ui.views.length) {
+        this.updatePortfolioInterval.restart({
+          timeout: INACTIVE_EXTENSION_PORTFOLIO_UPDATE_INTERVAL
+        })
+        this.fastAccountStateReFetchTimeout.stop()
+      }
+    })
+
```
PetromirDev, line 77: We shouldn't start/stop fastAccountStateReFetchTimeout on these events. Only on provider controller update

src/controllers/continuousUpdates/continuousUpdates.ts
```diff
@@ -0,0 +1,206 @@
...
+      this.fastAccountStateReFetchTimeout.updateTimeout({ timeout: updateTime })
+    }, 8000)
+
+    this.#main.onUpdate(() => {
+      if (this.#main.statuses.signAndBroadcastAccountOp === 'SUCCESS') {
+        this.accountStatePendingInterval.start({ timeout: ACCOUNT_STATE_PENDING_INTERVAL / 2 })
+        this.accountsOpsStatusesInterval.start()
+      }
+    }, 'continuous-update')
```
PetromirDev, line 192: When a transaction is broadcast, we have to stop the latest account state interval and start this one. When the pending interval is stopped, we have to start the latest one.

src/controllers/swapAndBridge/swapAndBridge.ts
sonytooo: When calling updateTimeout from within the func is like restart because right after the execution of the func finishes the new timeout will be with the updated time to wait

src/controllers/continuousUpdates/continuousUpdates.ts
sonytooo: the try catch block is in the createRecurringTimeout func so every callback is wrapped in it by default

src/utils/timeout.ts
```diff
@@ -1,28 +1,99 @@
...
-    timeoutId = undefined
+  const loop = async () => {
+    try {
+      await fn()
+      throw new Error('alabala')
+    } catch (err: any) {
+      console.error('Recurring task error:', err)
+      !!emitError &&
+        emitError({
+          error: new Error(err),
```
PetromirDev, line 33: don't recreate the error as that modifies the stack trace.

src/utils/timeout.ts
```diff
@@ -1,28 +1,99 @@
...
+  let debounceFlag = false
+  let currentTimeout = timeout
 
-  const stop = () => {
-    clearTimeout(timeoutId)
-    timeoutId = undefined
+  const loop = async () => {
+    try {
+      await fn()
+      throw new Error('alabala')
```
PetromirDev, line 28: delete this error :smile:

src/controllers/continuousUpdates/continuousUpdates.ts
```diff
@@ -0,0 +1,272 @@
...
+          'latest',
+          failedChainIds.map((id) => BigInt(id))
+        )
+
+        const failedChainIdsAfterUpdate = getNetworksWithFailedRPC({
+          providers: this.#main.providers.providers
+        })
+
+        const retriedFastAccountStateReFetchForNetworks =
+          this.#retriedFastAccountStateReFetchForNetworks
```
PetromirDev, line 205: We have to push these ids to `retriedFastAccountStateReFetchForNetworks`. Otherwise, if the account state update succeeds on the first try (after 8s), we won't update the portfolio

PetromirDev: Also `recentlyFailedNetworks` will be incorrect if we don't do this

PetromirDev (changes requested):

PR #1711, opened 2025-08-19
User op nonce check optimizations

Oxbobby:
### Problem

When an impatient user comes and tries to broadcast a couple of userOps in quick succession, depending on the chain and bundler used, it could end up in an error, too many unnecessary requests fired, or a horror story.

The problem for this is very simple - `userOp` nonce reuse. Different bundlers handle that differently, some better than others. We don't want that, we want a unified solution where we make sure the nonce is up-to-date

### Solution

We are making sure that either the blockchain or our internal state is synced before allowing the bundler broadcast to happen. This is how it works in a few steps:
- we get the last `broadcastedButNotConfirmed` userOp for this chain, account
- we check if the nonce is the same as the new userOp's nonce
- if it is, we return an error for the bundler estimation without firing an `estimate` request as... it's useless as it will either return a "nonce error" or worse - end up completing successfully which will bring other problems down the line

After returning a nonce error, it's handled in two different ways depending whether we are on Ethereum or not

#### Ethereum
Screenshot 2025-08-19 at 12 05 39

#### Other chains
On other chains, we wait 2s before fetching the newest userOp pending nonce and retry the estimate() function. If either the nonce has incremented OR the last `broadcastedButNotConfirmed` userOp has confirmed and it's no longer present, we do a bundler estimate request. It's important to ignore the last `broadcastedButNotConfirmed` userOp if it's already confirmed on the blockchain as that would mean the RPCs will handle the nonce sync better than us (as a failed bundler broadcast might mean that the userOp has not incremented

---

In this process, we skip a few bundler requests and get a unified solution handler regardless of the bundler

As a side note to this, when using gelato, we ask for a re-estimate just before broadcast always, on any chain, so we have consistent results with the bundler without hitting gas price errors

### DIFF
--- a/src/services/bundlers/gelato.ts
+++ b/src/services/bundlers/gelato.ts
@@ -61,6 +61,6 @@ export class Gelato extends Bundler {
   }
 
   public shouldReestimateBeforeBroadcast(network: Network): boolean {
-    return !!network.isOptimistic
+    return true
   }
 }
--- a/src/services/bundlers/gelato.ts
+++ b/src/services/bundlers/gelato.ts
@@ -61,6 +61,6 @@ export class Gelato extends Bundler {
   }
 
   public shouldReestimateBeforeBroadcast(network: Network): boolean {
-    return true
+    return !!network.isOptimistic
   }
 }
--- a/src/controllers/estimation/estimation.ts
+++ b/src/controllers/estimation/estimation.ts
@@ -243,7 +243,6 @@ export class EstimationController extends EventEmitter {
       })
     }
 
-    // TODO: test this
     if (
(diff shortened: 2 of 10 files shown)


src/services/bundlers/gelato.ts
```diff
@@ -61,6 +61,6 @@ export class Gelato extends Bundler {
   }
 
   public shouldReestimateBeforeBroadcast(network: Network): boolean {
-    return !!network.isOptimistic
+    return true
```
jordan-enev, line 64: as this is always `true`, we can remove `shouldReestimateBeforeBroadcast` check in all conditions we are using it

Ivshti (commented):
lgtm but maybe "empty" callRelayer needs to be documented with a comment

PR #1715, opened 2025-08-19
Feature/ Allow the user to change the fee payer key

PetromirDev:


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -475,18 +476,26 @@ export class SignAccountOpController extends EventEmitter implements ISignAccoun
       this.accountOp.signingKeyType = this.accountKeyStoreKeys[0].type
     }
 
+    if (this.accountOp.gasFeePayment && !this.accountOp.gasFeePayment.paidByKeyType) {
+      const key = this.#keystore.getFeePayerKey(this.accountOp)
+
+      if (key instanceof Error) return
+
+      this.accountOp.gasFeePayment.paidByKeyType = key.type
+    }
```
Oxbobby, line 485: I think this is a bit risky as at first, we don't have a `this.accountOp.gasFeePayment` set. So we need at least 2 updates from the fe to set this. The first one will set the `gasFeePayment` if all properties for it have been collected, and the second will then set the default `paidByKeyType`.

Perhaps it would be better to put this logic as a one time logic in `#setGasFeePayment`. Let's discuss

Oxbobby (approved):

PR #1725, opened 2025-08-25
Refactoring/ Replace Ether's decode/encode functions with viem's

PetromirDev:
## Changes:
- Replaces ether's encode/decode functions with viem, which are around 10times faster :scream: 
- Deletes `estimateEOA` (no longer used, check https://github.com/AmbireTech/ambire-app/pull/5390)

## Speed comparison:
image

### How to run locally:
- Init a project and install `viem` and `ethers`
- Run this script [index.js](https://github.com/user-attachments/files/21989225/index.js)

## How to test:
It needs to be tested heavily because deployless is used in many places:
- tokens, defi, nft
- account state
- debug trace call
- sign message validation (after signing)
- add new network
and there might be something broken. You can find all places by searching for `fromDescriptor` in the IDE.

I've added 2 "Why?" comments for defi positions - for some reason we're passing `network.rpcNoStateOverride` in reverse compared to other places, but when I flipped it, it didn't work :thinking: 
image

Additionally, there's one comment for a TS error in `signMessage`, but it works and I'm not sure if there's a more elegant way to do it. The problem is that the expected type is hex, but for typed messages we're passing a Uint8 array with length 32.

## Next steps:
https://github.com/AmbireTech/ambire-app/issues/5393


jordan-enev (commented):
Tested with EOA 7702, Smart Account, Ledger - doing txns, swaps, signed messages on MM testing dapp, adding networks - no issues at all.

Would be great if @JIOjosBG or @Ivshti could take a look to `rpcNoStateOverride` question you have and here as well https://github.com/AmbireTech/ambire-common/pull/1725/files#diff-47f85a476b255cdfd76a224dab1e9141f6e2c4d1965b2ff34b38d37778dddd7aR417

PR #1747, opened 2025-09-02
Fix: route disabling on max native

Oxbobby:
Disable routes that require a native fee when we're bridging out all out available native


superKalo (approved):

PetromirDev (approved):
Nice

PR #1748, opened 2025-09-02
Bug / Check if sign operation is still valid after each async step during signing a message

superKalo:
This PR addresses a race condition bug in the SignMessage controller where async operations could complete after the signing operation has been reset, leading to crashes. The fix introduces validation checks to ensure the signing operation is still valid after each async step.

* Adds a validation method to check if signing operation is still valid after async operations (similar to the one in Swap & Bridge controller)
* Inserts validation checks after each async operation in the signing flow

Crash report: https://monitor.ambire.com/organizations/ambire/issues/387

### DIFF
--- a/src/controllers/signMessage/signMessage.ts
+++ b/src/controllers/signMessage/signMessage.ts
@@ -130,6 +130,19 @@ export class SignMessageController extends EventEmitter implements ISignMessageC
     this.emitUpdate()
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


PetromirDev (changes requested):
Nice, but we should allow rejecting/resetting the controller after the signing process has started. I think that's the core issue

https://github.com/user-attachments/assets/1b90fe9a-9732-430d-bed8-447df70e7b31

PR #1753, opened 2025-09-03
Update / Transfer Field Validation Messages

gergana95:
Check if there are any account operations that were sent to a specific address with helper function in activity controller

Display messages in 3 stages 
1. User has not interacted with the given address -   "You're trying to send to an address you've never sent funds to before (this history is stored only in your current browser). Double-check before confirming."
2. User has not interacted, but address is in Address Book -  "Bear in mind you have not sent to this address before (this history is stored only in your current browser), but it is in your Address Book.'
3. This message stays as before - "This address isn't in your Address Book. Double-check the details before confirming."

Do not display message if:
- User has interacted before, but address is not in Address Book

Closes #927

### DIFF
--- a/src/controllers/transfer/transfer.ts
+++ b/src/controllers/transfer/transfer.ts
@@ -643,11 +643,6 @@ export class TransferController extends EventEmitter implements ITransferControl
       this.recipientAddress,
       this.#selectedAccountData.account.addr
     )
-    console.log(
-      '[TransferController] Checking recipient address',
-      'previousTransactionExists',
-      previousTransactionExists
-    )
 
     // Update state based on whether there are previous transactions to this address
     this.isRecipientAddressFirstTimeSend =
--- a/src/controllers/transfer/transfer.ts
+++ b/src/controllers/transfer/transfer.ts
@@ -652,7 +652,6 @@ export class TransferController extends EventEmitter implements ITransferControl
     // Update state based on whether there are previous transactions to this address
     this.isRecipientAddressFirstTimeSend =
       !previousTransactionExists &&
-      this.isRecipientAddressUnknown &&
       this.recipientAddress.toLowerCase() !== FEE_COLLECTOR.toLowerCase()
     this.signAccountOpController = new SignAccountOpController(
       this.#accounts,
--- a/src/controllers/transfer/transfer.ts
+++ b/src/controllers/transfer/transfer.ts
@@ -114,6 +114,8 @@ export class TransferController extends EventEmitter implements ITransferControl
 
   #relayerUrl: string
 
+  isRecipientAddressFirstTimeSend: boolean = false
+
   signAccountOpController: ISignAccountOpController | null = null
 
   /**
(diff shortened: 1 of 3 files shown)


src/services/validations/validate.ts
```diff
@@ -54,6 +54,10 @@ const validateAddAuthSignerAddress = (address: string, selectedAcc: any): Valida
 
 const NOT_IN_ADDRESS_BOOK_MESSAGE =
   "This address isn't in your Address Book. Double-check the details before confirming."
+const FIRST_TIME_SEND_MESSAGE =
+  "You're trying to send to an address you've never sent funds to before (this history is stored only in your current browser). Double-check before confirming."
+const FIRST_TIME_SEND_IN_ADDRESS_BOOK_MESSAGE =
```
superKalo, line 59: These two should look more like a warning (or a note), not an actual error. They're too scary now.

Screenshot 2025-09-15 at 15 20 32

src/services/validations/validate.ts
```diff
@@ -54,6 +54,10 @@ const validateAddAuthSignerAddress = (address: string, selectedAcc: any): Valida
...
+  'Bear in mind you have not sent to this address before (this history is stored only in your current browser), but it is in your Address Book.'
```
superKalo, line 60: Text proposal:

> First time sending to this address - this browser's local history shows no prior transactions. Please double-check before sending, even though the recipient in your Address Book.

src/services/validations/validate.ts
```diff
@@ -54,6 +54,10 @@ const validateAddAuthSignerAddress = (address: string, selectedAcc: any): Valida
 
 const NOT_IN_ADDRESS_BOOK_MESSAGE =
   "This address isn't in your Address Book. Double-check the details before confirming."
+const FIRST_TIME_SEND_MESSAGE =
+  "You're trying to send to an address you've never sent funds to before (this history is stored only in your current browser). Double-check before confirming."
```
superKalo, line 58: Text proposal:

> First time sending to this address - no prior transactions in this browser’s history. Please double-check before sending.

src/controllers/transfer/transfer.ts
```diff
@@ -114,6 +114,10 @@ export class TransferController extends EventEmitter implements ITransferControl
 
   #relayerUrl: string
 
+  isRecipientAddressFirstTimeSend: boolean = false
```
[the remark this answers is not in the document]
gergana95: Following the logic behind the previously set variables isRecipientAddressUnknown, isRecipientAddressUnknownAgreed, isRecipientHumanizerKnownTokenOrSmartContract.

src/services/validations/validate.ts
```diff
@@ -54,6 +54,20 @@ const validateAddAuthSignerAddress = (address: string, selectedAcc: any): Valida
...
+  'First time sending to this address - no prior transactions in this browser’s history. Please double-check before sending, even though the recipient is in your Address Book.'
+
+function getDaysAgo(date: Date): string {
+  const now = new Date()
+  const diffMs = now.getTime() - date.getTime()
+  const diffDays = Math.floor(diffMs / (1000 * 60 * 60 * 24))
+  if (diffDays === 0) return 'today'
+  if (diffDays === 1) return 'yesterday'
+  return `${diffDays} days ago`
+}
```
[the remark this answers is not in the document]
gergana95: Thanks! Will do!

PR #1754, opened 2025-09-03
Fixes / ContinuousUpdatesController tests

sonytooo:
* Fixed RecurringTimeout bug where restarting the timeout in the same tick caused multiple function calls, and the function’s promise was not awaited in the correct place, allowing a new execution to start before the previous one had finished.
* ContinuousUpdatesController tests refactoring and fixes to improve stability

### DIFF
--- a/src/classes/recurringTimeout/recurringTimeout.ts
+++ b/src/classes/recurringTimeout/recurringTimeout.ts
@@ -67,6 +67,7 @@ export class RecurringTimeout implements IRecurringTimeout {
   }
 
   stop() {
+    this.startScheduled = false
     this.#reset()
   }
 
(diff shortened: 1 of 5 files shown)


src/controllers/accounts/accounts.ts
```diff
@@ -49,6 +49,8 @@ export class AccountsController extends EventEmitter implements IAccountsControl
   // Holds the initial load promise, so that one can wait until it completes
   initialLoadPromise?: Promise<void>
 
+  accountStatesInitialLoadPromise?: Promise<void>
```
PetromirDev, line 52: add a comment for this

src/classes/recurringTimeout/recurringTimeout.ts
```diff
@@ -77,6 +80,7 @@ export class RecurringTimeout implements IRecurringTimeout {
 
     try {
       this.promise = this.#fn()
+      this.fnExecutionsCount += 1
```
PetromirDev, line 83: shouldn't we increment after awaiting the promise?

PetromirDev (changes requested):

PetromirDev (approved):

PR #1755, opened 2025-09-04
Feature/dual swap providers

Oxbobby:
Use both Li.Fi. and socket at the same time, combining route results.

Build a new provider service called `SwapProviderParallelExecutor`. It expects all the service providers we're going to use in parallel. In this case, those are lifi and bungee

### DIFF
--- a/src/controllers/activity/activity.ts
+++ b/src/controllers/activity/activity.ts
@@ -174,6 +174,18 @@ export class ActivityController extends EventEmitter implements IActivityControl
 
   banners: Banner[] = []
 
+  #updateAccountsOpsStatusesPromises: {
+    [accountAddr: string]:
+      | Promise<{
+          shouldEmitUpdate: boolean
+          // Which networks require a portfolio update?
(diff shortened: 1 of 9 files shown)


src/controllers/main/main.ts
```diff
@@ -401,10 +399,11 @@ export class MainController extends EventEmitter implements IMainController {
...
       activity: this.activity,
       invite: this.invite,
-      // TODO: This doesn't work, because the invite controller is not yet loaded at this stage
-      // serviceProviderAPI: this.invite.isOG ? lifiAPI : socketAPI,
-      serviceProviderAPI: lifiAPI,
       storage: this.storage,
+      swapProvider: new SwapProviderParallelExecutor([
+        new LiFiAPI({ fetch }),
+        new SocketAPI({ fetch })
+      ]),
```
superKalo, line 406: Tweaked a bit in https://github.com/AmbireTech/ambire-common/pull/1787

PR #1759, opened 2025-09-05
Bridge interval time

Oxbobby:
Configure the bridge interval time. Until now, it was once each minute.  
From now on:
* we get the service time for the bridge from the provider and start the interval after that time has passed. This could mean a couple of seconds if it's Relay, 30-40s on a bit slower bridges, and 5+ minutes on Stargate
* if a long wait time bridge is initiated and after a new one with a few seconds in queued, we check if the new bridge's time is more than twice lower than the one we had until now. If it is, we stop the interval and start it anew with the new service time

Also:
* changed the text of pending bridges to show the actual number of pending bridges instead of a count of all the bridges


src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -599,6 +602,21 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
     this.#storage.set('swapAndBridgeActiveRoutes', value)
 
     if (this.activeRoutesInProgress.length) {
+      const minServiceTime = getActiveRoutesLowestServiceTime(this.activeRoutesInProgress)
```
sonytooo, line 605: here is how I think the logic should be:

`  set activeRoutes(value: SwapAndBridgeActiveRoute[]) {
    this.#activeRoutes = value
    // eslint-disable-next-line @typescript-eslint/no-floating-promises
    this.#storage.set('swapAndBridgeActiveRoutes', value)

    if (!this.activeRoutesInProgress.length) {
      this.#updateActiveRoutesInterval.stop()
      return
    }

    const minServiceTime = getActiveRoutesLowestServiceTime(this.activeRoutesInProgress)

    if (!this.#updateActiveRoutesInterval.running) {
      this.#updateActiveRoutesInterval.start({ timeout: minServiceTime })
      return
    }

    // If the interval is running, check if minServiceTime * 2 is still less than currentTimeout.
    // If it is, restart it with the new minServiceTime, as the difference makes it worth it.
    if (minServiceTime * 2 < this.#updateActiveRoutesInterval.currentTimeout) {
      this.#updateActiveRoutesInterval.restart({ timeout: minServiceTime })
    }
  }`

PR #1797, opened 2025-09-19
Fix: outdated logic that checked if the asset is a bridge

Oxbobby:


superKalo (approved):

superKalo:
PS: Minor note, branch name should have been `release/v2.74.2`.

PR #1805, opened 2025-09-30
Fix / Slow response from the parallel providers execution

sonytooo:
Resolves: https://github.com/AmbireTech/ambire-app/issues/5576

### DIFF
--- a/src/services/swapIntegrators/swapProviderParallelExecutor.ts
+++ b/src/services/swapIntegrators/swapProviderParallelExecutor.ts
@@ -9,6 +9,7 @@ import {
   SwapProvider
 } from '../../interfaces/swapAndBridge'
 import { TokenResult } from '../../libs/portfolio'
+import wait from '../../utils/wait'
 
 export class SwapProviderParallelExecutor {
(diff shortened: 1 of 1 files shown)


src/services/swapIntegrators/swapProviderParallelExecutor.ts
```diff
@@ -36,16 +37,61 @@ export class SwapProviderParallelExecutor {
...
+
+    // Figure out how long we've already waited
+    const elapsed = Date.now() - startTime
+    // If first was too quick, extend wait time so total ≥ MIN_WAIT
+    const remainingMinWait = Math.max(0, MIN_WAIT - elapsed)
+
+    const secondResult = (await Promise.race([
+      Promise.any(remainingTasks),
+      wait(MAX_WAIT_AFTER_FIRST_COMPLETED + remainingMinWait)
+    ])) as { provider: SwapProvider; result: Error | T }
```
PetromirDev, line 79: We must include `absoluteTimeout` in the race. If the first provider takes 14 sec, then this can exceed the timeout by going to 16.

PR #1812, opened 2025-10-01
Show projected rewards in dashboard

Maskln:
related with PR: https://github.com/AmbireTech/ambire-app/pull/5595


JIOjosBG (approved):
already added two normal comments

PR #1817, opened 2025-10-02
Fixes / ContractNamesController tests

sonytooo:
* Fixed ContractNamesController failing tests
* Fixed ContractNamesController getName logic – previously, calling getName with the same address within the debounce period would add the address to the fetch queue multiple times

### DIFF
--- a/src/controllers/contractNames/contractNames.test.ts
+++ b/src/controllers/contractNames/contractNames.test.ts
@@ -12,6 +12,8 @@ const contracts = {
 let finishPromise
 let errorPromise
 
+const DEFAULT_DEBOUNCE = 100
+
 describe('Contract Names', () => {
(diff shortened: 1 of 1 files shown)


src/controllers/contractNames/contractNames.test.ts
```diff
@@ -73,30 +73,38 @@ describe('Contract Names', () => {
     // request a non contract address that will fail
     contractNamesController.getName(randomAddress, 8453n)
 
-    // The default time for debounce is 50ms
-    jest.advanceTimersByTime(50)
+    // The default time for debounce is 100ms
+    jest.advanceTimersByTime(100)
```
superKalo, line 77: If that's a constant from the logic, take it from there instead of hardcoding it?

PR #2124, opened 2026-01-21
Add Colibri verified RPC provider (env-gated) 

simon-jentzsch:
- Adds optional Colibri-stateless integration as a verified RPC path for Sepolia (11155111) behind USE_COLIBRI=true.
- Introduces ColibriRpcProvider (extends ethers.JsonRpcProvider) and routes send(method, params) through Colibri’s EIP‑1193 request().
- Keeps JsonRpcProvider compatibility (destroy(), _getConnection().url, etc.) to avoid broader refactors.
- eth_call with state override (3rd param object) is bypassed and executed via the underlying RPC provider to preserve deployless/state-override behavior.
- Default prover: https://sepolia.colibri-proof.tech/ (override via COLIBRI_PROVER_URLS).
- Colibri docs: JS/TS bindings


Oxbobby:
We started working on colibri on a separate branch: https://github.com/AmbireTech/ambire-common/pull/2506

PR #2147, opened 2026-01-28
Fix/ Email vault tries to remove 'password' secret when it doesn't exist

PetromirDev:
Please review my change and the test error meticulously!

Closes https://github.com/AmbireTech/ambire-app/issues/6123


Ivshti:
no longer relevant with the email vault deprecation

PR #2195, opened 2026-02-18
Fix / Sign Screen - App Catalog validation

jordan-enev:
- Add: DappsController - `hasUnverifiedDappsAndSendResToUi` method for verifying if passed `dapps` as parameter are part of the verified DefiLlama dapps.

### DIFF
--- a/src/controllers/dapps/dapps.ts
+++ b/src/controllers/dapps/dapps.ts
@@ -761,10 +761,17 @@ export class DappsController extends EventEmitter implements IDappsController {
     })
   }
 
-  async hasUnverifiedDappsAndSendResToUi({ dapps }: { dapps: string[] }) {
+  async hasUnverifiedDappsAndSendResToUi({
+    requestId,
+    dapps
+  }: {
+    requestId: string
+    dapps: string[]
+  }) {
     const hasUnverifiedDapps = this.hasUnverifiedDappUrls(dapps)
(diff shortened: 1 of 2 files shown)


src/controllers/dapps/dapps.ts
```diff
@@ -748,6 +749,36 @@ export class DappsController extends EventEmitter implements IDappsController {
     this.#ui.message.sendUiMessage(message)
   }
 
+  protected hasUnverifiedDappUrls(dapps: string[]): boolean {
+    const verifiedDapps = Array.from(this.#dapps.values()).filter(
+      (dapp) => dapp.blacklisted === 'VERIFIED'
+    )
+    const verifiedDappUrlsSet = new Set(verifiedDapps.map((dapp) => dapp.url.toLowerCase()))
```
superKalo: Minor, `this.#dapps` list is relatively huge list and this logic iterates 2 times though that big list (`Array.from` and `filter`), you can achieve the same thing with 1 iteration:

```
const verifiedDappUrlsSet = new Set<string>()
for (const dapp of this.#dapps.values()) {
  if (dapp.blacklisted === 'VERIFIED') verifiedDappUrlsSet.add(dapp.url.toLowerCase())
}
```

That's because .values() is returns an iterator, not an array ⚡

superKalo (approved):

PR #2220, opened 2026-03-04
Feature/ ENS v2

PetromirDev:
Prepares us for ENS v2 by implementing ENS's ethers monkey patch. Deletes custom logic as it's now handled by the monkey patch. 

Docs:
https://docs.ens.domains/web/ensv2-readiness


superKalo (approved):
Great that we removed the prev magic :)

PR #2268, opened 2026-03-20
Asset info/fix blacklisted tokens

JIOjosBG:
## Problem
The asset info uses the portfolio to get token decimals and symbol, but for blacklisted tokens, the portfolio does not return them

## This adds a 

after
image

before
image


JIOjosBG:
i probably double clicked `create`

closing in favor of https://github.com/AmbireTech/ambire-common/pull/2267

PR #2269, opened 2026-03-20
Feature / Biometrics

sonytooo:
* allow concurrent executions of same method when the first execution is in success status


superKalo (approved):
Makes sense, but let's test this carefully in `v2` against the extension also.

PR #2276, opened 2026-03-24
Fix/ Simulation race conditions and clearing on portfolio error

PetromirDev:
### `overrideSimulationResult` called during portfolio update
If `overrideSimulationResult` was called while the portfolio is being updated the simulation was removed but then added back by the portfolio update on success. To fix this, we add `overrideSimulationResult` to the update queue.

Video:
https://github.com/user-attachments/assets/d2fe1933-f6cf-4cbb-938b-c75faa9a4ed7

### simulation not stored/removed on error
If an account op is passed for simulation but the portfolio update fails we weren't storing it in the portfolio state. This had two implications:
- Discarding an account op doesn't work if the update fails (possibly https://github.com/AmbireTech/ambire-app/issues/6851)
- Simulated account ops are not persisted on failure (if the first simulation fails, subsequent calls were ok)

To reproduce:
1. Prepare a transaction
2. Prepare a network request block for the RPC of the network
3. Broadcast it
4. The MOMENT that the transaction is confirmed enable network blocking (before the portfolio update)
5. `discardAccountOp` will be called, it will fail and the simulation will not be cleared (this was happening before)

## Tests:
I added two tests to prevent regressions.

Closes https://github.com/AmbireTech/ambire-app/issues/6849

### DIFF
--- a/src/controllers/portfolio/portfolio.test.ts
+++ b/src/controllers/portfolio/portfolio.test.ts
@@ -657,7 +657,7 @@ describe('Portfolio Controller ', () => {
       expect(areAccountOpsEqual(stateBefore.accountOps!, accountOp['1']!)).toBe(true)
       expect(collectionBefore?.amountPostSimulation).toBe(0n)
 
-      controller.overrideSimulationResults(accountOp['1']![0]!)
+      await controller.overrideSimulationResults(accountOp['1']![0]!)
 
       const stateAfter = getEthereumPortfolioState(controller)
       const collectionAfter = getSimulatedCollection(controller)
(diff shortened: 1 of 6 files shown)


Oxbobby (changes requested):
In signAccountOp, there's a missing await on overrideSimulationResults:

```
    if (this.estimation.status === EstimationStatus.Error) {
      this.#portfolio.overrideSimulationResults(this.accountOp)
    }
```

Oxbobby:
If overrideSimulationResults is always waiting for the whole queue, what happens if a call in the queue gets stuck? Does each call in the queue have a timeout?

PR #2280, opened 2026-03-25
Qr code based hw integration

Maskln:
paired with: https://github.com/AmbireTech/ambire-app/pull/6803

Summary
- QR HW wallet support (base layer): types, interfaces, and extensibility for multiple QR wallets
- Signer updates: QR-specific state + request/response flow handling
- Account Picker: QR import flow via QrKeyIterator + new init handler
- Signatures: added normalizeSignatureHex (supports {r,s,v} + hex)
- Misc: status wiring + type cleanups for future integrations


src/controllers/accountPicker/accountPicker.ts
```diff
@@ -863,15 +863,33 @@ export class AccountPickerController extends EventEmitter implements IAccountPic
...
+        qr: this.#externalSignerControllers.qr?.deviceModel || ''
       }
 
+      const masterFingerprint = this.#externalSignerControllers.qr?.masterFingerprint || ''
+      const qrWalletConfig = keyType === 'qr' ? this.keyIterator.walletConfig : undefined
+      const qrParsedOriginHdPath =
+        keyType === 'qr'
+          ? this.keyIterator.parsedAccount?.hdPath ||
+            this.keyIterator.parsedAccount?.accounts?.[0]?.hdPath
+          : undefined
```
Oxbobby: @superKalo I'm not proficient enough here so tagging Kalo for assistance.

I guess I'm wondering why do we have different hd paths for accounts? And is it safe that we're not using an account specific hdPath for qr accounts but defaulting to `this.keyIterator.parsedAccount?.hdPath` for each one of them?

src/controllers/accountPicker/accountPicker.ts
```diff
@@ -863,15 +863,33 @@ export class AccountPickerController extends EventEmitter implements IAccountPic
...
+      const hdPathTemplate = (
+        keyType === 'qr' && qrWalletConfig
+          ? qrWalletConfig.hdPathTemplate || this.hdPathTemplate
+          : this.hdPathTemplate
+      ) as HD_PATH_TEMPLATE_TYPE
```
Oxbobby: I see we prioritize `qrWalletConfig.hdPathTemplate` here. What if the user has changed this from the UI?  
My questions are:
1. Is the user allowed to change his hdPathTemplate from the UI
2. If yes, would be reflected in `qrWalletConfig.hdPathTemplate` (I don't think so as this happens in `setHDPathTemplate`)

Imo, it would be better to find a way to use the controller `this.hdPathTemplate` at all times. Maybe set it initially from the qrWalletConfig and allow the UI to dispatch changes after

src/controllers/accountPicker/accountPicker.ts
superKalo: Yeah, good catch.

The hd path is critical to know, since we calculate the next accounts ourselves from the xpub of the first account we receive from the QR code.

So in this case I'd advice in the key iterator - we make sure the hdPath of the first account exists and if not - to fallback with the standard hd path in there. So on this level (in the AccountPicker) - we should always have the `hdPath` and throw if missing.

src/controllers/accountPicker/accountPicker.ts
superKalo: 1. I think the answer is no - hd path should not be changable in our UI. But we should test with different hd path set on the hardware wallet (e. g. on imToken wallet) or in general - what's the flow other wallets do for different than standard hd path.

Oxbobby (approved):

src/consts/hardwareWallets.ts
```diff
@@ -3,5 +3,6 @@ import { ExternalKey } from '../interfaces/keystore'
 export const HARDWARE_WALLET_DEVICE_NAMES: { [key in ExternalKey['type']]: string } = {
   ledger: 'Ledger',
   trezor: 'Trezor',
-  lattice: 'GridPlus'
+  lattice: 'GridPlus',
+  qr: 'Qr-based'
```
superKalo: Minor: I think we should capitalize both letters, so I'd put **QR**-based here.

src/controllers/accountPicker/accountPicker.ts
```diff
@@ -863,15 +863,43 @@ export class AccountPickerController extends EventEmitter implements IAccountPic
...
+            this.keyIterator.parsedAccount?.accounts?.[0]?.hdPath
+          : undefined
+
+      if (keyType === 'qr' && !qrParsedOriginHdPath) {
+        throw new EmittableError({
+          message: 'QR account import missing origin hdPath',
+          level: 'major',
+          error: new Error('QR account import missing origin hdPath')
+        })
+      }
```
superKalo: We need better user-facing error. I'd suggest: Can’t import account. This QR code is missing required account data (HD path) from your QR hardware wallet.

Also, this should never happen, right? So I'd put "should never happen" comment + is there any other useful information we can log on Sentry that may help us debug if for some reason it happens? Like for which air-gapped wallet it happened?

src/interfaces/keystore.ts
```diff
@@ -154,6 +163,12 @@ export type ExternalKey = {
     hdPathTemplate: HD_PATH_TEMPLATE_TYPE
     index: number
     createdAt: number | null
+
+    qrWalletType?: QrWalletType
+    qrProtocol?: QrProtocolType
+    originHdPath?: string // for accounts imported from QR, to store the original path used on the wallet, as some wallets (like Keystone) don't provide the possibility to retrieve the key by xpub, but only by path, so we need to keep track of it in order to be able to retrieve the key later on
+    masterFingerprint?: string // for Ledger, optional for others, as it can be used for additional info but is not required to retrieve the key
```
superKalo: for Ledger?

src/controllers/accountPicker/accountPicker.ts
```diff
@@ -888,8 +916,11 @@ export class AccountPickerController extends EventEmitter implements IAccountPic
             meta: {
               deviceId: deviceIds[keyType],
               deviceModel: deviceModels[keyType],
+              masterFingerprint,
               // always defined in the case of external keys
-              hdPathTemplate: this.hdPathTemplate as HD_PATH_TEMPLATE_TYPE,
+              hdPathTemplate: hdPathTemplate,
+              relativePathTemplate: relativePathTemplate,
+              originHdPath: originPath,
```
superKalo: I don't like this part. It has a lot of noise.

For all external signers we have `hdPathTemplate`, which is for example: `m/44'/60'/0'/0/<account>`

So why we need extra 2 props `relativePathTemplate` (example: `0/{index}`) and `originHdPath` (example: `m/44'/60'/0'`)? Can't we just calculate them based on the `hdPathTemplate`?

Also, why mixing `<account>` (that we already use for all) and `{index}`. They mean the same thing.

Also, it's strange that these two values are only set for QR-based wallets and would be `undefined` for all others. If they are QR-based specifics only, why don't we name them somehow indicating that (same applies for (`masterFingerprint`, naming doesn't suggest it's QR-based wallets only and for all others - would be undefined)

src/controllers/main/main.ts
```diff
@@ -1141,6 +1141,45 @@ export class MainController extends EventEmitter implements IMainController {
...
+
+      const keyIterator = new QrKeyIterator({ controller: qrCtrl })
+      await keyIterator.importAccount(payload)
+
+      this.accountPicker.setInitParams({
+        keyIterator,
+        hdPathTemplate: keyIterator.walletConfig!.hdPathTemplate,
+        pageSize: 5,
+        shouldAddNextAccountAutomatically: false,
+        shouldSearchForLinkedAccounts: false
```
superKalo, line 1183: Please add context why searching for linked accounts is disabled. Otherwise, it will get lost in the Slack discussions. Base on memory, it was decided like so for simplicity sake?

src/controllers/main/main.ts
```diff
@@ -1142,6 +1142,45 @@ export class MainController extends EventEmitter implements IMainController {
...
+      const qrCtrl = this.#externalSignerControllers.qr
+
+      if (!qrCtrl) {
+        const message =
+          'Could not initialize connection with your QR hardware wallet. Please try again later or contact Ambire support.'
+        throw new EmittableError({ message, level: 'major', error: new Error(message) })
+      }
+
+      const keyIterator = new QrKeyIterator({ controller: qrCtrl })
+      await keyIterator.importAccount(payload)
```
superKalo: All other flows init key iterator and then - AccountPicker.

This one makes something strange - **imports account**. A couple of problems:

This doesn't import account, makes a bunch of other things though - sets walletConfig, sets xpub and even sets importedHdPath.

For me it's confusing (no context why this has to be done in advance) and hard to follow. Why this logic is not triggered in the key iterator retrieve method (which AccountPicker calls)?

src/interfaces/keystore.ts
```diff
@@ -154,6 +163,12 @@ export type ExternalKey = {
     hdPathTemplate: HD_PATH_TEMPLATE_TYPE
     index: number
     createdAt: number | null
+
+    qrWalletType?: QrWalletType
+    qrProtocol?: QrProtocolType
+    originHdPath?: string // for accounts imported from QR, to store the original path used on the wallet, as some wallets (like Keystone) don't provide the possibility to retrieve the key by xpub, but only by path, so we need to keep track of it in order to be able to retrieve the key later on
```
superKalo: Can you give an example how this is different than the combo between the existing `hdPathTemplate` and the `index` props?

src/interfaces/keystore.ts
```diff
@@ -154,6 +163,12 @@ export type ExternalKey = {
...
+    masterFingerprint?: string // for Ledger, optional for others, as it can be used for additional info but is not required to retrieve the key
+    relativePathTemplate?: string // QR related
```
superKalo: Can you give an example how this is different than the combo between the existing `hdPathTemplate` and the `index` props?

src/controllers/main/main.ts
Maskln: Good point, and I agree the old naming made the flow look misleading.

In the QR case, this step is not “import account into Ambire” - it parses the scanned payload and initializes iterator state (walletConfig, xpub, parsedAccount) that AccountPicker needs before first retrieve().

retrieve() is then only address derivation over a range, using that initialized state.   

I renamed the method to `initFromQrPayload` to make this explicit and added an comment in the code why we need it.

src/interfaces/keystore.ts
Maskln: Removed the prop.

src/interfaces/keystore.ts
Maskln: Removed the prop.

src/controllers/main/main.ts
Oxbobby: We can add a comment like this:

v1 accounts have never supported QR wallets. In the rare case of migration (ledger -> qr -> has created a linked account), the user should instead go to the web wallet and migrate his funds instead. v1 accounts are generally deprecated and we don't encourage users to use them

src/interfaces/keystore.ts
```diff
@@ -55,6 +55,13 @@ export interface ExternalSignerController {
   appName?: string // Lattice specific
   creds?: any // Lattice specific
   network?: any // Lattice specific
+  masterFingerprint?: string // Optional for some wallets, but can be used for additional info
+  // TODO: fix the type
```
PetromirDev: When are we going to resolve these todos?

PetromirDev (approved):
Very good!

src/utils/normalizeSignatureHex.ts
```diff
@@ -0,0 +1,16 @@
+import { Signature } from 'ethers'
+
+import { addHexPrefix } from './addHexPrefix'
+import { stripHexPrefix } from './stripHexPrefix'
+
+export function normalizeSignatureHex(input: { r?: string; s?: string; v?: number; hex?: string }) {
+  if (input.hex) return addHexPrefix(stripHexPrefix(input.hex))
+
+  const signature = Signature.from({
```
sonytooo: we should try catch here because of the optional props

superKalo (approved):

PR #2294, opened 2026-04-02
Hotfix / Swap & Bridge - stop Quote update interval when the pop-up is closed or tab is inactive

jordan-enev:
Fix: Pause `updateQuoteInterval` when the pop-up is closed or the tab is inactive. Resume it when the form becomes visible again, and trigger an immediate quote update if the last one was more than 20 seconds ago.


gergana95 (approved):

JIOjosBG:
bug, not result of this PR:
when i click proceed without signing, the next iteration of the polling will cancel the polling. Then when i click *Back* on the bottom sheet there is no more updating of the quote

PR #2299, opened 2026-04-07
Fix / RecurrungTimeout race condition

sonytooo:
**Implements a Session ID locking mechanism in RecurringTimeout to prevent stale background tasks from blocking immediate foreground requests.**

**The Issue:**
When start() and restart() were called in the same event loop tick, 
multiple microtasks were queued. A background-queued "wait" task 
(Microtask A) would execute first, setting the timer to 'running' 
and scheduling a 12s sleep. The subsequent immediate "fetch" 
(Microtask B) would then see the timer was already 'running' 
and return early, effectively being ignored.

**The Fix:**
Added an `internalSessionId` that increments on every stop/restart. 
Microtasks now self-destruct if their captured session ID doesn't 
match the current one, ensuring that immediate foreground requests 
always take priority over background noise.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ambire-common",
-  "version": "2.89.3",
+  "version": "2.91.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "ambire-common",
-      "version": "2.89.3",
+      "version": "2.91.2",
       "dependencies": {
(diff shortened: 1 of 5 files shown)


superKalo (approved):
I'm not sure I've fully grasped this one, but we discussed offline that this is a problem that was only experienced on mobile. I'll keep an eye on it.

Oxbobby (approved):

superKalo (approved):

PR #2306, opened 2026-04-09
By default, load a token in swap&bridge that's on a network supported by the account

Oxbobby:
Example: a safe account that's deployed on base but not on optimism. For some reason, the token with the highest amount in your portfolio is on optimism. You open swap&bridge from the dashboard header button and we select that token - but the network is not supported as the safe account is not deployed there.

This fixes that by checking the account networks as well before selecting the initial token

Pointing it to NEXT


PetromirDev (approved):
Nice catch

PR #2307, opened 2026-04-14
Feature/ `eth_createAccessList` and `toBeLearnedAssets` cleanup

PetromirDev:
### Adds: `eth_createAccessList` and uses it for asset discovery
The new method is used in all cases that don't require state override (because state override is not yet supported by many RPC providers). If state override is needed we fallback to debug_traceCall.

**Safe accounts**:
It was not a trivial task to get it working with Safe accounts, because `execTransaction` requires signatures (which we obviously don't have during simulation). To get around this, we use Safe's `SimulateTxAccesssor`, which simulates `execTransaction` but reverts at the end. This is perfect for our use case as `eth_createAccessList` succeeds despite reverts.

https://github.com/safe-global/safe-deployments/blob/main/src/deployments.ts

### The portfolio controller cleans toBeLearnedAssets

### DIFF
--- a/src/libs/tracer/accessListCall.ts
+++ b/src/libs/tracer/accessListCall.ts
@@ -1,0 +1,131 @@
+import { getAddress, toQuantity } from 'ethers'
+
+import { RPCProvider } from '@/interfaces/provider'
+import { getFunctionParams } from '@/libs/tracer/debugTraceCall'
+
+import { ProviderError } from '../../classes/ProviderError'
+import { AccountOnchainState } from '../../interfaces/account'
+import { Network } from '../../interfaces/network'
+import { getRpcProvider } from '../../services/provider'
+import { BaseAccount } from '../account/BaseAccount'
+import { AccountOp } from '../accountOp/accountOp'
+
+/**
+ * Parses an access list and extracts unique contract addresses
+ */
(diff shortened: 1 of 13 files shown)


src/libs/tracer/accessListCall.ts
```diff
@@ -0,0 +1,231 @@
...
+        [SAFE_CALL_OPERATION, call[0], BigInt(call[1]), BigInt(getBytes(call[2]).length), call[2]]
+      )
+    )
+
+    // For batched ops, Safe executes MultiSend via DELEGATECALL.
+    to = multiSendAddr
+    value = 0n
+    data = multiSendIface.encodeFunctionData('multiSend', [concat(multiSendCalls)])
+    operation = SAFE_DELEGATE_CALL_OPERATION
+  }
```
Oxbobby, line 84: Minor: We have the same logic in lib/safe.ts -> getSafeTxn(), but it's part of the method.

In lib/safe.ts, we can make a encodeCalls() method that accepts the AccountOp as arg, does getSignableCalls on it and encodes them.

We can then reuse it here and in getSafeTxn()

Oxbobby:
It works wonderful

Oxbobby (approved):

PR #2313, opened 2026-04-16
Decode calls in sign-acc-op

Oxbobby:
High level overview in https://github.com/AmbireTech/ambire-app/pull/7035

Adds:
- new `ContractInfo` controller that fetches the function selectors via sourcify
- `decodeCall` function that decodes the arguments in a call and structures them so the FE can display them easily


src/controllers/contractInfo/contractInfo.ts
```diff
@@ -0,0 +1,195 @@
...
+        this.#fetch,
+        cenaUrl,
+        {},
+        3000
+      ).then((r) => r.json())
+
+      if (!result.success) {
+        this.emitError({
+          error: new Error('Failed to fetch contract selectors'),
+          level: 'major',
```
Oxbobby: level might need to be silent, we don't need to alert the user that we've faield to fetch this?

src/controllers/contractInfo/contractInfo.ts
```diff
@@ -0,0 +1,195 @@
...
+            data: mappedFoundSignatures,
+            status: 'success',
+            updatedAt: Date.now()
+          }
+        else this.selectors[selector] = { status: 'not-found', updatedAt: Date.now() }
+      })
+    } catch (e: any) {
+      this.emitError({
+        error: e,
+        level: 'major',
```
Oxbobby: the same here, level should be silent if we agree

src/controllers/contractInfo/contractInfo.ts
```diff
@@ -0,0 +1,195 @@
...
+          message: 'Failed to fetch contract selectors',
+          sendCrashReport: true
+        })
+        selectorsToFetch.forEach((s) => {
+          this.selectors[s] = { status: 'error', error: result.error, updatedAt: Date.now() }
+        })
+        this.emitUpdate()
+        void this.#storeSelectorsInStorage()
+        return
+      }
```
Oxbobby: On failure, we can return the fallback, or initial result, so the user can continue - but we can make another retry to cena with a bigger timeout (10s)

JIOjosBG: very good point: we should persists old success vals when after expiration the fetch fails

also yes to the retry

both should have appropriate tests

src/controllers/contractInfo/contractInfo.ts
JIOjosBG: sounds reasonable, maybe if the fetch fails we should display some warning about this in the dropdown

Oxbobby: Let's start without warnings as I don't know how far this will be understood. For example, if you are a person that seldom to never checks this, if you get a warning "failed to fetch alabala" that you don't understand what it means, it will be net negative for us

src/controllers/contractInfo/contractInfo.ts
```diff
@@ -0,0 +1,192 @@
...
+    this.emitUpdate()
+  }
+
+  async #storeSelectorsInStorage() {
+    const selectorsToStore: Selectors = {}
+    Object.entries(this.selectors).forEach(([k, v]) => {
+      if (v.status === 'loading') return
+      selectorsToStore[k] = v
+    })
+    await this.#storage.set(FUNCTION_SELECTORS_STORAGE_KEY, selectorsToStore)
```
Oxbobby, line 77: We shouldn't expect this to grow exponentially, right?

src/libs/decodeCall/index.ts
```diff
@@ -0,0 +1,132 @@
...
+  data: Call['data'],
+  foundSignatures: { signature: string }[]
+): DecodedCall | null {
+  if (!isHex(data)) return null
+  let resultWithDiff: { diff: number; decoded: DecodedCall | null } = {
+    diff: Infinity,
+    decoded: null
+  }
+  for (const { signature } of foundSignatures) {
+    const iface = new Interface(['function ' + signature])
```
Oxbobby: maybe this should be in the try / catch block?

src/libs/decodeCall/index.ts
```diff
@@ -0,0 +1,132 @@
...
+      console.error(e)
+      console.error(`decodeCall: ${e.message}`)
+      // TODO should we ignore it?
+    }
+  }
+  // this is just a false positive MITIGATION
+  // in cases where the data part is 1 slot (32 bytes) and there is
+  // a found function that does not have arguments
+  // encountered as issue on a zero slot 0x00000.000000
+  if (data.length === '0x'.length + 64 && resultWithDiff.diff === 16) return null
```
Oxbobby: AI comment, please check if it makes sense or not:

The false-positive guard looks off. A 32-byte blob decoded as a no-arg function has diffInBytes === 28, not 16; a 20-byte address-like blob has diffInBytes === 16. So that condition likely misses the case described by the comment

src/controllers/contractInfo/contractInfo.ts
Oxbobby: yep, I don't need this:
Screenshot 2026-05-18 at 10 22 21

src/controllers/contractInfo/contractInfo.ts
JIOjosBG: not exponentially, but it will grow rather fast.

For the most used 100 function selectors the relayer returns about 205kb. so i would expect we would store maybe 1.5x that in the storage.

All this includes all those 100 selectors, and all selectors that match their first 2 bytes

JIOjosBG: we can make the user store only the requested selector instead of all selectors

src/libs/decodeCall/index.ts
JIOjosBG: i am changing to 28

src/controllers/contractInfo/contractInfo.ts
Oxbobby: half a MB seems reasonable

PR #2316, opened 2026-04-17
Built-in surveys

JIOjosBG:
## How it works
- On the relayer side we will create 1 banner an 1 survey for every survey we want the users to fill out
- Inside the banner there will be info about requirement for the user, like min/max balance or txns
- If the user has the balance/txns we display the banner 
- filling out the survey dismisses the banner
- response with the user's address is sent to the relayer so we do not serve  the survey+banner to the same account on another device PR [relayer/1772](https://github.com/AmbireTech/relayer/pull/1772)

## Why in a separate controller and not in banners controller
- so we can display surveys when user does something specific in the extension in the future
- it handles the state of the UI as well so it is persisted on popup close


src/controllers/banner/banner.ts
```diff
@@ -47,9 +73,48 @@ export class BannerController extends EventEmitter implements IBannerController
...
+
+  /**
+   * Used when account is being switched, because we might want to display
+   * different banners for different accounts.
+   * The first and only (Apr 2026) such case is survey banners that have
+   * to be filtered depending on balance, tx count and keys for acc
+   */
+  emitUpdateBanners() {
+    this.emitUpdate()
+  }
```
JIOjosBG, line 120: is this ok practice 1/2

src/controllers/main/main.ts
```diff
@@ -699,6 +739,7 @@ export class MainController extends EventEmitter implements IMainController {
       await this.requests.removeUserRequests([swapAndBridgeSigningRequest.id])
     }
     await this.selectedAccount.setAccount(accountToSelect)
+    this.banner.emitUpdateBanners()
```
JIOjosBG: is this ok practice 2/2

src/controllers/survey/survey.ts
```diff
@@ -0,0 +1,269 @@
...
+    ui.uiEvent.on('removeView', () => {
+      if (this.status === 'success-submitted') {
+        this.clearSurveyState()
+      }
+    })
+  }
+
+  async #load() {
+    const surveysRespondedTo: string[] = await this.#storage.get(ANSWERED_SURVEYS_STORAGE_KEY, [])
+    this.#surveysRespondedTo = surveysRespondedTo
```
PetromirDev, line 70: you need to emit an update to update the getter `isReady` in the UI

src/controllers/survey/survey.ts
```diff
@@ -0,0 +1,269 @@
...
+      return
+    }
+
+    const parsedSurvey = parseSurvey(res.survey as unknown)
+    if (!parsedSurvey.ok) {
+      this.emitError({
+        message: 'There was error fetching the survey.',
+        level: 'major',
+
+        sendCrashReport: true,
```
PetromirDev: sendCrashReport: true is the default value

src/controllers/survey/survey.ts
```diff
@@ -0,0 +1,269 @@
...
+        error: Error(`Error with parsing a survey ${parsedSurvey.error}`)
```
PetromirDev, line 117: Imo, pass the original error. There may be more data attached to it.

src/controllers/survey/survey.ts
```diff
@@ -0,0 +1,269 @@
...
+  // TODO: should we refactor the controller?
+  async fetchSurvey(surveyId: Survey['surveyId'], sourceBannerId?: string | number) {
+    this.status = 'loading-fetching'
+    this.emitUpdate()
+
+    let res: any
+
+    try {
+      res = await this.#callRelayer(
+        `/promotions/survey/${surveyId}`,
```
PetromirDev: TODO: Check for race conditions - not checking for `loading-fetching` before starting - the user may request a surverId, close the banner and request a new one?

src/controllers/survey/survey.ts
```diff
@@ -0,0 +1,269 @@
...
+      return
+    }
+
+    this.#surveysRespondedTo.push(surveyId)
+    await this.#storage.set('surveysRespondedTo', this.#surveysRespondedTo)
+  }
+
+  async #sendResponse(instanceId: string, address: string) {
+    if (this.#survey) {
+      if (!this.answers || !Object.keys(this.answers).length) {
```
PetromirDev: It'd be better to do 

if (!this.#survey) {
 // Emit silent error
 return
}

So that:
- Code is easier to read - less nesting
- Internal errors are not ignored

PetromirDev (changes requested):

src/controllers/survey/survey.ts
JIOjosBG: this is reached when 1. the relayer returned a survey successfully, but for some reason the survey does not have proper format. There is no actual exception at this step

src/controllers/survey/survey.ts
JIOjosBG: why should the err be silent?

src/controllers/survey/survey.ts
JIOjosBG: what do you think about:
```ts
async fetchSurvey(surveyId: Survey['surveyId'], sourceBannerId?: string | number) {
    if (this.status === 'loading-fetching') return
    this.status = 'loading-fetching'
    this.emitUpdate()
    ....
```

src/controllers/survey/survey.ts
PetromirDev: Maybe it shouldn't be, depends on whether we want the user to see it or not

src/controllers/main/main.ts
PetromirDev: I mean it's okay, because you don't just emit an update to update the getters, but add a new method that explains this. On the other hand:
1. This pattern is not used atm
2. What if there are other getters in the future? Then the name becomes incorrect

The current pattern is:
```
    // forceEmitUpdate to update the getters in the FE state of the ctrls
    await Promise.all([
      this.activity.forceEmitUpdate(),
      this.requests.forceEmitUpdate(),
      this.addressBook.forceEmitUpdate(),
      this.swapAndBridge.forceEmitUpdate(),
      this.dapps.broadcastDappSessionEvent('accountsChanged', [toAccountAddr]),
      this.forceEmitUpdate()
    ])
```

src/controllers/banner/banner.ts
```diff
@@ -1,24 +1,51 @@
+import { ISurveyController } from '@/interfaces/survey'
+
+import pkg from '../../../package.json'
 import { Banner, IBannerController } from '../../interfaces/banner'
 import { IEventEmitterRegistryController } from '../../interfaces/eventEmitter'
 import { IStorageController } from '../../interfaces/storage'
 import EventEmitter from '../eventEmitter/eventEmitter'
 
+const currentCommonVersion = pkg.version
```
PetromirDev: Use `appVersion` from `main`

PetromirDev (changes requested):

src/controllers/banner/banner.ts
JIOjosBG: we discussed offline, this will be kept in the banners controller, but the caller will `SelectedAccountController` instead of `MainController`

src/controllers/main/main.ts
JIOjosBG: we discussed in huddle that we can keep this, but emit the banner update in a different place

src/controllers/survey/survey.ts
JIOjosBG: This is a scary problem, and imo the user should be reminded to restart the extension in this case.

In theory should not happen as the dashboard and almost all other parts of the extension will not be visible, so the user should no be able to see anything about the surveys

src/controllers/survey/survey.test.ts
```diff
@@ -0,0 +1,264 @@
...
+      status: 200,
+      text: async () => JSON.stringify({ success: true })
+    }
+  }
+  return { ok: false, status: 404 }
+})
+
+const getFreshControllers = () => {
+  const { uiManager } = mockUiManager()
+
```
PetromirDev: we usually call this function prepareTest and its response is usually:

{ ...somethingElse, controller }

and not a function that returns the controller

src/controllers/survey/helpers.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+  if (q.responseType === 'openText')
+    question = {
+      id: q.id,
+      questionPosition: q.questionPosition,
+      text: q.text,
+      responseType: 'openText',
+      responseOptions: null
+    }
```
PetromirDev, line 42: Is this much validation actually needed?

src/controllers/main/main.ts
```diff
@@ -299,13 +304,55 @@ export class MainController extends EventEmitter implements IMainController {
...
     })
+
+    this.survey = new SurveyController({
+      fetch: this.fetch,
+      relayerUrl,
+      storage: this.storage,
+      ui: this.ui,
+      eventEmitterRegistry,
+      dismissBanner: (bannerId: Banner['id']) => {
+        console.log('DEBUG dismissing banner from mainCtrl', bannerId, this.banner)
```
PetromirDev: delete this log

src/controllers/banner/banner.ts
```diff
@@ -47,9 +75,53 @@ export class BannerController extends EventEmitter implements IBannerController
...
     // Always return one banner at a time
-    return this.#getValidBanners(this.#banners).slice(0, this.maxBannerCount)
+    const accData = this.#getAccountData()
+
+    return {
+      banners: this.#getValidBanners(this.#banners)
+        .filter((b) => this.#notSurveyOrValidSurvey(b, accData))
+        .slice(0, this.maxBannerCount),
+      account: accData.status === 'has-selected-account' ? accData.address : null
+    }
```
PetromirDev, line 131: Maybe we should change this getter in a way that doesn't call `this.#getAccountData()` in the root. Instead, call it only if there is a survey banner. Why? Because this getter is reevaluated on every portfolio update if portfolio.isReady=true and `this.#getAccountData()` gets all accountOps and sorts them (which is not that heavy on modern machines, but I think it's more expensive to do than to run `#getAccountData()` multiple times if there is more than one survey (which should be rare, right?)

There likely is an even better way to do this - just try to call `this.#getAccountData()` only once and only if needed.

PetromirDev (changes requested):

src/controllers/survey/survey.ts
PetromirDev: I don't think that there is a way for the user to start fetching, change the account and click on a different survey fast enough, but fyi that this doesn't handle that case. So yes, let's go like that. Also, ensure that users can't submit multiple times by spamming the submit button @JIOjosBG

src/controllers/survey/survey.test.ts
JIOjosBG: I wanted to have access to the new storage before the controller is initialized. I saw prepareTest, but thought it would not be as easy to use for this case

src/controllers/survey/helpers.ts
JIOjosBG: Imo yes, because all banners will be manually added directly in the database in the relayer. Added them just to make sure that banners with invalid structure do not break anything

src/controllers/banner/banner.ts
JIOjosBG: i tried 3 different ways:
Since we need the `account` for the banners response anyway we can 1) add another callback for getting the acc, 2) add the selectedAccController and change the callback a bit or 3) pass all controllers and no callback

All of them look a little overengineered, how about we simply add another function to the activity that returns the number of transactions without sorting them to make the existing code lighter @PetromirDev

PetromirDev: Honestly, yes, let's just add a `sortByTimestamp` param to `getAccountOpsForAccount` with a default value of true, and use false there.

src/controllers/survey/helpers.ts
PetromirDev: Soo, can't the relayer check these before returning?

src/controllers/survey/survey.test.ts
PetromirDev: You can do it. Check the `initialSetStorage` pattern in `makeMainController`

src/controllers/survey/survey.test.ts
```diff
@@ -0,0 +1,264 @@
+import { Survey } from '@/interfaces/survey'
+import { getNextQuestionForAnswers } from '@/utils/survey'
+import { expect } from '@jest/globals'
+import { produceMemoryStore } from '@test/helpers'
+import { mockUiManager } from '@test/helpers/ui'
```
PetromirDev: there are ts errors in the file

PetromirDev (changes requested):

src/controllers/survey/helpers.ts
JIOjosBG: technically yes, but we have had cases where the relayer's code is edited on the fly and deployed. Last time this caused the extension to crash because of missing props (then it was rewards related)

PetromirDev (approved):

PR #2317, opened 2026-04-19
Smooth max transfer

Oxbobby:
Change log:
* a sync mechanism between transfer & sign-account-op for MAX transfers
* enable the MAX transfered out token as a fee (if it's a legit fee token)
* suppress "not enough to pay the fee" errors during synching period in one click transfer mode

### DIFF
--- a/src/controllers/signAccountOp/signAccountOp.ts
+++ b/src/controllers/signAccountOp/signAccountOp.ts
@@ -150,6 +150,7 @@ import {
   getUnknownTokenWarning,
   SignAccountOpType
 } from './helper'
+import { isTransferredTokenFeeOption } from '../../libs/account/feeOptions'
 
 export enum SigningStatus {
   EstimationError = 'estimation-error',
@@ -857,7 +858,10 @@ export class SignAccountOpController extends EventEmitter implements ISignAccoun
       this.canBroadcast
     ) {
       const identifier = getFeeSpeedIdentifier(this.selectedOption, this.accountOp.accountAddr)
-      if (this.hasSpeeds(identifier))
+      if (
+        this.hasSpeeds(identifier) &&
+        !this.#shouldSuppressTransferFeeSelectionError(this.selectedOption)
+      )
         errors.push({
           title: 'Please select a token and an account for paying the gas fee.'
         })
@@ -868,6 +872,7 @@ export class SignAccountOpController extends EventEmitter implements ISignAccoun
       this.selectedOption &&
       this.accountOp.gasFeePayment &&
       this.selectedOption.availableAmount < this.accountOp.gasFeePayment.amount &&
+      !this.#shouldSuppressTransferFeeSelectionError(this.selectedOption) &&
       this.canBroadcast
     ) {
       const speedCoverage = []
@@ -1608,13 +1613,14 @@ export class SignAccountOpController extends EventEmitter implements ISignAccoun
   #getIsFeeOptionDisabled(feeOption: FeePaymentOption): boolean {
     const id = getFeeSpeedIdentifier(feeOption, this.accountOp.accountAddr)
     const speeds = this.feeSpeeds[id] ?? []
+    const isTransferredTokenOption = isTransferredTokenFeeOption(feeOption, this.accountOp)
 
     const coversSlow = speeds.some(
       (speed: SpeedCalc) =>
         speed.type === FeeSpeed.Slow && feeOption.availableAmount >= speed.amount
     )
 
-    if (!coversSlow) return true
+    if (!coversSlow && !isTransferredTokenOption) return true
 
     const isExternal = this.accountKeyStoreKeys.some(
       (keyStoreKey) => keyStoreKey.addr === feeOption.paidBy && keyStoreKey.isExternallyStored
@@ -1625,6 +1631,14 @@ export class SignAccountOpController extends EventEmitter implements ISignAccoun
     if (isExternal && canNotBecomeSmarter && feeOption.token.address !== ZERO_ADDRESS) return true
 
     return false
+  }
+
+  #shouldSuppressTransferFeeSelectionError(feeOption?: FeePaymentOption): boolean {
+    return (
+      this.#type === 'one-click-transfer' &&
+      !!feeOption &&
+      isTransferredTokenFeeOption(feeOption, this.accountOp)
+    )
   }
 
   /**
--- a/src/controllers/signAccountOp/signAccountOp.ts
+++ b/src/controllers/signAccountOp/signAccountOp.ts
@@ -150,7 +150,7 @@ import {
   getUnknownTokenWarning,
   SignAccountOpType
 } from './helper'
-import { isTransferredTokenFeeOption } from '../../libs/account/feeOptions'
+import { canFeeOptionCoverAmount, isTransferredTokenFeeOption } from '../../libs/account/feeOptions'
 
 export enum SigningStatus {
   EstimationError = 'estimation-error',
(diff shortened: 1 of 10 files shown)


src/controllers/transfer/transfer.ts
```diff
@@ -918,7 +983,11 @@ export class TransferController extends EventEmitter implements ITransferControl
     })
 
     this.signAccountOpController.onUpdate((forceEmit) => {
-      this.propagateUpdate(forceEmit)
+      const hasAdjustedMaxAmount = this.#syncMaxAmountWithFeeReservation(forceEmit)
+
+      if (!hasAdjustedMaxAmount) {
+        this.propagateUpdate(forceEmit)
```
jordan-enev: I'm almost sure this could cause issues if it's not emitted on every signAccountOp update. Why? Because if the front-end accesses transfer.signAccountOp.status or any other signAccountOp property, it won't be in sync with the latest signAccountOp state due to the conditional update above.

src/controllers/transfer/transfer.ts
```diff
@@ -760,6 +767,63 @@ export class TransferController extends EventEmitter implements ITransferControl
...
+    this.amountFieldMode = 'token'
+    this.#setAmount(amount, isProgrammaticUpdate)
+    this.amountFieldMode = amountFieldMode
+  }
+
+  #syncMaxAmountWithFeeReservation(forceEmit?: boolean) {
+    if (
+      !this.#isMaxAmountSelected ||
+      !this.selectedToken ||
+      typeof this.selectedToken.decimals !== 'number'
```
jordan-enev: How this could be possible? Maybe we can skip this check: `typeof this.selectedToken.decimals !== 'number'`

src/controllers/transfer/transfer.ts
```diff
@@ -760,6 +767,63 @@ export class TransferController extends EventEmitter implements ITransferControl
...
+      !!accountAddr &&
+      !!gasFeePayment &&
+      !!selectedFeeOption &&
+      !selectedFeeOption.token.flags.onGasTank &&
+      selectedFeeOption.paidBy.toLowerCase() === accountAddr &&
+      selectedFeeOption.token.chainId === this.selectedToken.chainId &&
+      selectedFeeOption.token.address.toLowerCase() === this.selectedToken.address.toLowerCase() &&
+      gasFeePayment.inToken.toLowerCase() === this.selectedToken.address.toLowerCase() &&
+      (!gasFeePayment.feeTokenChainId ||
+        gasFeePayment.feeTokenChainId === this.selectedToken.chainId)
```
jordan-enev: I would suggest moving this into a private helper method to make the flow of the main method (#syncMaxAmountWithFeeReservations) easier to read.

Something like this:

```js
  #shouldReserveFeeFromTransferredToken() {
    const gasFeePayment = this.signAccountOpController?.accountOp.gasFeePayment
    const selectedFeeOption = this.signAccountOpController?.selectedOption
    const selectedToken = this.selectedToken
    const accountAddr = this.#selectedAccount.account?.addr.toLowerCase()

    if (!accountAddr || !gasFeePayment || !selectedFeeOption || !selectedToken) return false
    if (selectedFeeOption.token.flags.onGasTank) return false
    if (selectedFeeOption.paidBy.toLowerCase() !== accountAddr) return false

    const selectedTokenAddress = selectedToken.address.toLowerCase()

    return (
      selectedFeeOption.token.chainId === selectedToken.chainId &&
      selectedFeeOption.token.address.toLowerCase() === selectedTokenAddress &&
      gasFeePayment.inToken.toLowerCase() === selectedTokenAddress &&
      (!gasFeePayment.feeTokenChainId || gasFeePayment.feeTokenChainId === selectedToken.chainId)
    )
  }
```

src/controllers/transfer/transfer.ts
```diff
@@ -760,6 +767,63 @@ export class TransferController extends EventEmitter implements ITransferControl
...
+
+    const desiredAmount = shouldReserveFeeFromTransferredToken
+      ? getAmountAfterFeeReserve(totalTokenAmount, gasFeePayment.amount)
+      : totalTokenAmount
```
jordan-enev: I would prefer to short-circuit here instead of using a ternary operation, to explicitly emphasize that we won't update the token amount:

```js
if (!shouldReserveFeeFromTransferredToken) return

const maxTransferableAmount = getAmountAfterFeeReserve(totalTokenAmount, gasFeePayment.amount)

```

src/controllers/transfer/transfer.ts
jordan-enev: Now I see that in `this.#syncMaxAmountWithFeeReservation(forceEmit)` we also call this.propagateUpdate(forceEmit), and this check here is meant to reduce the number of calls we make.

Even so, I'd prefer not to tightly couple the two updates/functions, because a small change in `syncMaxAmountWithFeeReservation` could easily break the main propagateUpdate.

If there's no performance degradation, let's keep the two updates separate and simple. What do you think?

Oxbobby: It's OK with me but I'm not sure I fully understand. You're proposing to remove both "propagateUpdate" calls and to allow the updates to happen in the next tick?

jordan-enev: No, I meant to always call `propagateUpdate` on both places - onUpdate callback and into your function as well.

src/controllers/transfer/transfer.ts
```diff
@@ -760,6 +762,104 @@ export class TransferController extends EventEmitter implements ITransferControl
...
+    if (selectedFeeOption.token.flags.onGasTank) return false
+    if (selectedFeeOption.paidBy.toLowerCase() !== accountAddr) return false
+
+    const selectedTokenAddress = selectedToken.address.toLowerCase()
+
+    return (
+      !!accountAddr &&
+      !!gasFeePayment &&
+      !!selectedFeeOption &&
+      !selectedFeeOption.token.flags.onGasTank &&
```
jordan-enev: this looks redundant, as we already had checked it above

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -857,7 +858,10 @@ export class SignAccountOpController extends EventEmitter implements ISignAccoun
       this.canBroadcast
     ) {
       const identifier = getFeeSpeedIdentifier(this.selectedOption, this.accountOp.accountAddr)
-      if (this.hasSpeeds(identifier))
+      if (
+        this.hasSpeeds(identifier) &&
+        !this.#shouldSuppressTransferFeeSelectionError(this.selectedOption)
```
jordan-enev, line 912: This method, `shouldSuppressTransferFeeSelectionError`, and the way we use it look more like a patch or a way of bypassing validation errors.

I was thinking about whether we can skip it altogether. Currently, the problem seems to come from (if I’m not mistaken) calling `this.syncSignAccountOp()` in controllers/transfer with the max available amount, without deducting the fee:

```js
if (shouldSetMaxAmount) {
  this.#isMaxAmountSelected = true
  this.amountFieldMode = 'token'
  this.#setTokenAmount(this.maxAmount, true)
}
```

What if, at this point, we set `this.maxAmount - fee` in cases where the selected fee option matches the current token? Along with that, in signAccountOpController.update, we can sync the max amount back when the fee option changes.

Wdyt?

Oxbobby: We discussed this offline and here's what we came up with:
I agree this solution is hackish and it should be avoided if possible. In our current design, it's a bit more difficult to do so. Imagine this scenario:
1. MAX transfer on USDT
2. Another fee token is selected by default in signAccountOp
3. You select USDT as the fee

At that moment, get errors() are recalculated in signAccountOp and they get displayed in the UI. The updates are down -> top, from signAccountOp to transfer. If it were the opposite, we could catch this in transfer.ts and prevent it. So a solution where this hack is deleted is a solution where we change the update cycle from down -> top to top -> down.

For this reason, we decided to go with the hackish solution until we find a better one

PR #2320, opened 2026-04-20
Config / Disable (maybe temporarily) injecting $WALLET token on Base as an option the Swap & Bridge "Receive" list

superKalo:
Because of the decision to remove $WALLET liquidity on Base and consolidate it into the Ethereum liquidity, it doesn't make much sense injecting $WALLET token on Base as an option the Swap & Bridge "Receive" list anymore.


Oxbobby (approved):

superKalo:
Side note, user would be still able to add WALLET in lists by pasting token address.

PR #2326, opened 2026-04-22
Config / Make Phishing controller blacklist updates UI-aware and improve update intervals (15m when active, 6h when inactive)

superKalo:
Updates `PhishingController` to use adaptive polling based on UI activity (15m when active, 6h when inactive) by wiring it up to the `UiController` events.

Before the update was every 3h no matter what. After these changes (and the server changes mentioned in https://github.com/AmbireTech/ambire-app/issues/6974, which updates its list every 15 minutes), if an app gets blacklisted, extension blacklist will have it between 1 min (best case) to 30 minutes (worst case).

It also fixes the throttling logic by comparing against the current interval timeout and updating in-memory `updatedAt` after successful fetches, preventing redundant requests during rapid popup reopen cycles.

Also adds unit tests for active/inactive interval switching and immediate restart behavior, plus small inline comments/debug hooks to make update flow easier to reason about :v: 

Resolves https://github.com/AmbireTech/ambire-app/issues/6974


src/controllers/phishing/phishing.ts
```diff
@@ -62,25 +67,40 @@ export class PhishingController extends EventEmitter implements IPhishingControl
...
       this.emitError.bind(this)
     )
 
+    this.#ui.uiEvent.on('addView', () => {
+      if (this.#ui.views.length === 1)
+        this.#updatePhishingInterval.restart({
+          timeout: PHISHING_ACTIVE_UPDATE_INTERVAL,
+          runImmediately: true
+        })
+    })
```
PetromirDev: The list will be updated on every popup open. Is that the desired behaviour?

superKalo: I think yes. On every popup open we need to restart the interval (and set it to the active interval, which is quicker - 15 minutes) and then on close - same but reversed - restart it with the inactive value (6h).

This will bounce in `#continuouslyUpdatePhishing` if update is not needed

```js
// This prevents redundant requests to the relayer
// when the extension reloads multiple times within a short period.
const timeSinceLastUpdate = this.#updatedAt ? Date.now() - this.#updatedAt : null
if (
  this.#updatedAt &&
  timeSinceLastUpdate !== null &&
  timeSinceLastUpdate < this.updatePhishingInterval.currentTimeout
)
```

Is there a better approach for this? 👀

superKalo: We agreed offline that this is the desired behavior ✅

src/controllers/phishing/phishing.ts
```diff
@@ -114,8 +134,27 @@ export class PhishingController extends EventEmitter implements IPhishingControl
...
+    ) {
+      // NOTE: used for debugging only
+      // console.log(
+      //   `[PhishingController] Skip update (sinceLastUpdate=${Math.floor(timeSinceLastUpdate / 1000)}s, timeout=${Math.floor(this.updatePhishingInterval.currentTimeout / 1000)}s)`
+      // )
+
+      return
+    }
+
+    // NOTE: used for debugging only
```
PetromirDev, line 185: There is no loading check. What if the user opens and closes the popup multiple times within 60 seconds and his internet is not that fast?

src/controllers/phishing/phishing.ts
```diff
@@ -62,25 +67,40 @@ export class PhishingController extends EventEmitter implements IPhishingControl
...
+      PHISHING_INACTIVE_UPDATE_INTERVAL,
       this.emitError.bind(this)
     )
 
+    this.#ui.uiEvent.on('addView', () => {
+      if (this.#ui.views.length === 1)
+        this.#updatePhishingInterval.restart({
+          timeout: PHISHING_ACTIVE_UPDATE_INTERVAL,
+          runImmediately: true
+        })
```
PetromirDev: Imo change this to any `addView` of type request window. This way:
- It will still update, even if the user has left a tab open somewhere (making length > 1)
- It will update on every added request window (where we actually need the phishing list)

PetromirDev (changes requested):

src/controllers/phishing/phishing.ts
superKalo: Yes, this was indeed a problem. Handled in https://github.com/AmbireTech/ambire-common/pull/2326/changes/029f3903740c7d386ecef40344913f5b9449ada9 by introducing a wrapper around `#continuouslyUpdatePhishing` that: 1) deduplicates concurrent triggers via a shared promise and 2) switches to the failed-retry interval when the fetch/update flow throws.

src/controllers/phishing/phishing.ts
superKalo: Improved in https://github.com/AmbireTech/ambire-common/pull/2326/changes/11a483add3a2e0db9d984d33030b0131ef88c5cd, but let's discuss further if this is optimal.

PetromirDev (approved):

sonytooo (approved):

PR #2328, opened 2026-04-23
Feature / Address poisoning detection to the transfer flow

superKalo:
This PR introduces poisoning analysis in activity.ts:

- Added: Recipient is now scanned against previously used recipients from transaction history for consecutive left/right address matches. A recipient is flagged as suspicious for address poisoning when 1) it's a first time send and 2) there's another transaction with a recipient that have a total of at 8 matching prefix and suffix characters. The detection returns metadata describing the matched trusted address plus the exact prefix/suffix match lengths.
- Changed: Transfer controller now consumes that metadata during recipient history checks, and validation logic uses it to show an address poisoning warning with a formatted preview of the trusted address (meaning the address one've prev sent to).
- Changed: The `shortenAddress` util now supports fixes size visible characters for the cases where a single `maxLength` is not enough - e.g. show exactly N symbols at the beginning and in the end.

Notes:

- I tested the performance with and without the poisoning checks with 10k account ops. It increases the `hasAccountOpsSentTo` execution about x2 times (expected), but the total execution time still seems ~ reasonable. Let's discuss this further offline:
  Screenshot 2026-04-27 at 12 10 30
- To be discussed, if what I prepared (a simple comparison at the beginning and at the end) is enough or we should make something more complicated like [this approach](https://github.com/Robert-MacWha/kohaku/blob/blocklist/packages/security/src/address-poisoning/index.ts).


src/controllers/activity/activity.ts
```diff
@@ -77,6 +77,54 @@ const paginate = (items: any[], fromPage: number, itemsPerPage: number) => {
...
+  if (matchedPrefixCharsCount + matchedSuffixCharsCount < MIN_ADDRESS_POISONING_TOTAL_MATCH_CHARS) {
+    return null
+  }
+
+  return {
+    matchedPrefixCharsCount,
+    matchedSuffixCharsCount
+  }
+}
+
```
PetromirDev: Please move this somewhere else, outside of the activity controller

src/controllers/activity/activity.ts
```diff
@@ -206,21 +254,70 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+          (lastInteractedAt ?? -1) > (bestPoisoningMatch.lastInteractedAt ?? -1))
+      ) {
+        bestPoisoningMatch = {
+          matchedAddress: address,
+          matchedPrefixCharsCount: matchCounts.matchedPrefixCharsCount,
+          matchedSuffixCharsCount: matchCounts.matchedSuffixCharsCount,
+          lastInteractedAt
+        }
+      }
+    }
```
PetromirDev, line 308: Move this outside of the controller too

src/services/validations/validate.ts
```diff
@@ -68,6 +70,39 @@ const NOT_IN_ADDRESS_BOOK_MESSAGE =
...
+}: AddressPoisoningMatch) => {
+  let normalizedAddress = matchedAddress
+
+  try {
+    normalizedAddress = getAddress(matchedAddress)
+  } catch {
+    // keep original if checksum normalization fails
+  }
+
+  const strongestSymmetricMatch = Math.min(matchedPrefixCharsCount, matchedSuffixCharsCount)
```
PetromirDev: this should be Math.max, right? We are looking for the strongest, not the weakest

PetromirDev (changes requested):

src/controllers/activity/activity.ts
superKalo: Agree, moved in https://github.com/AmbireTech/ambire-common/pull/2328/changes/b1cccec0b0c1bb1e214cfdf7e269908dc21918e6

src/controllers/activity/activity.ts
superKalo: Agreed, moved in https://github.com/AmbireTech/ambire-common/pull/2328/changes/b1cccec0b0c1bb1e214cfdf7e269908dc21918e6

src/services/validations/validate.ts
superKalo: You're right! `Math.max` was the correct one - the prev was a left over. This way the previewed chars are shortened based on the stronger match to better see the diff (example: 6+4 similarity should review 8…8 chars in the warn, not 6…6). Fixed in https://github.com/AmbireTech/ambire-common/pull/2328/changes/f2a2a10f1824d423922ae20f8d42aa1adf296a30

superKalo:
Re: AI review, missing import: it's not missing, its on line 1, imported from ethers.

```
import { getAddress, parseUnits } from 'ethers'
```

src/libs/transfer/address-poisoning.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Address poisoning lookalikes usually preserve a few consecutive chars from the left and/or
+ * right side of the original address. We consider any recipient with at least 8 matched chars
+ * in total as suspicious, while still keeping the exact prefix/suffix counts so downstream UI
+ */
+export const getAddressPoisoningMatchCounts = (candidate: string, trustedAddress: string) => {
+  const normalizedCandidate = candidate.toLowerCase()
+  const normalizedTrustedAddress = trustedAddress.toLowerCase()
+
+  if (
```
PetromirDev, line 17: I know that we have unit tests for transfer, but this logic is very important so I think we should also have extensive unit tests for the lib.

PetromirDev (changes requested):

gergana95 (approved):
I like the changes, tested and works good. I would suggest we update the message if current address has NOT send a transaction to the poisoned address ( but only imported the acc ), in this case I am not sure if the wording 'interacted with' is clear what we mean. It may be only a problem on my side, so not pushing on this.

src/libs/transfer/address-poisoning.ts
superKalo: I agree, it was useful for me too to put into words what the expected behavior should be. I added them in https://github.com/AmbireTech/ambire-common/pull/2328/commits/5af5ad98a511c28d2ce49f0dcc7e769254e87c1b

superKalo:
> I like the changes, tested and works good. I would suggest we update the message if current address has NOT send a transaction to the poisoned address ( but only imported the acc ), in this case I am not sure if the wording 'interacted with' is clear what we mean. It may be only a problem on my side, so not pushing on this.

You're absolutely right! But this is actually a bug, rather than a wording problem.

I thin it's better to scan **only against previously used recipients from transaction history**. Involving Address Book / imported accounts becomes confusing.

So for simplicity in https://github.com/AmbireTech/ambire-common/pull/2328/changes/ea8b1ac877ad04b5eb49d9b9187dacc9c4907b5f I changed it so that addresses from Address Book (including imported accounts) are not used as additional poisoning references. So only previously sent-to recipients from transaction history can trigger address poisoning detection.

Example: Imagine a "poisoned" address lands in the Address Book.

- If there are no transactions from the selected account towards a similar to this account, we show "first transaction":
  Screenshot 2026-05-08 at 14 23 20
- If there is a transaction towards a similar account:
  Screenshot 2026-05-08 at 14 23 32

PetromirDev (approved):

PR #2332, opened 2026-04-25
Feature/Balance-changes

Oxbobby:
Change log:
* remove the numbers from the uniswap humanizer as you will see them in balance changes instead. It's still up for discussion


src/libs/humanizer/modules/Uniswap/utils.ts
```diff
@@ -23,7 +29,21 @@ export const getUniRecipientText = (accAddr: string, recAddr: string): Humanizer
...
+      const previousToken = allTokens[index - 1]
+
+      return !(
+        previousToken &&
+        previousToken.type === 'token' &&
+        token.type === 'token' &&
+        previousToken.address === token.address &&
+        previousToken.value === token.value
+      )
+    })
```
JIOjosBG: since they are hidden, we should not really care if 1) they are one after another 2) the values match 3) that they duplicated at all

imo we do not need deduplication here at all, why did you add it?

src/libs/humanizer/modules/Uniswap/utils.ts
```diff
@@ -147,7 +197,9 @@ export const uniReduce = (_calls: HumanizerVisualization[][]): HumanizerVisualiz
         isSwap(callJ) &&
         callI[3].address === callJ[1].address
       ) {
-        callI?.push({ ...callI[3], isHidden: true })
+        if (!WRAPPED_NATIVE_TOKEN_ADDRESSES.has(callI[3].address?.toLowerCase())) {
```
JIOjosBG: why is it important that this is not wrapped native? 

To me it looks like this is a valid bugfix without the if statement, only with the push

src/libs/humanizer/modules/Uniswap/utils.ts
```diff
@@ -198,7 +259,9 @@ export const uniReduce = (_calls: HumanizerVisualization[][]): HumanizerVisualiz
         isTake(callJ) &&
         callI[1].address === callJ[1].address
       ) {
-        callI[1].value = callI[1].value! > callJ[1].value! ? callI[1].value : callJ[1].value
+        if (callI[1].value && callI[1].value > 0n) {
+          callI[1].value = callI[1].value > callJ[1].value! ? callI[1].value : callJ[1].value
+        }
         calls.splice(j, 1)
```
JIOjosBG, line 244: Since the logic in this file is kinda fragile, please add tests with the exact transactions that had buggy humanization that made you add those changes.

We need them because if someone else touches this file we have to be sure future changes do not break anything

JIOjosBG (changes requested):
feedback for the humanizer changes

src/controllers/storage/storage.ts
```diff
@@ -537,6 +540,27 @@ export class StorageController extends EventEmitter implements IStorageControlle
...
+      await activity.backfillRecentMissingBalanceChanges(10)
+
+      await this.#storage.set('passedMigrations', [...new Set([...passedMigrations, migrationId])])
+    })().finally(() => {
+      this.#backfillAccountOpBalanceChangesMigrationPromise = undefined
+    })
+
+    return this.#backfillAccountOpBalanceChangesMigrationPromise
+  }
+
```
PetromirDev: Nice! Can we also migrate on filter? E.g., if the user is filtering (paginating) account ops, and the result account ops are not migrated.

src/controllers/activity/activity.ts
```diff
@@ -413,6 +480,144 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+    } catch (error) {
+      console.log(error)
+      await this.setAccountOpBalanceChanges(
+        accountOp.identifiedBy,
+        accountOp.accountAddr,
+        accountOp.chainId,
+        []
+      )
+    }
+  }
```
PetromirDev, line 630: I'm not sure if we should set the balance changes to [] on failure/no rpc etc. This is not very flexible if we decide to also fetch balance changes on filter (another comment). If we decide to do that, then we can:
- Skip setting anything/fetching if there isn't an rpc (which shouldn't really happen) or if the rpc is not working
- On failure set a new property to the account op - `balanceChangesFetchRetryCount` or increment it if it already exists
- Stop refetching on 3 failures

I'm saying this only because our activity is in storage, which would require migrations if we decide to fetch all balance changes in the future.

src/libs/deployless/deployless.ts
```diff
@@ -146,6 +154,15 @@ export class Deployless {
...
   }
 
+  /**
+   * To be able to successfully pass a number as blockTag,
+   * we need to call toQuantity to it, making it a no-leading zeros hex.
+   * This is the standard to make sure RPCs don't revert
+   */
+  private static normalizeRpcBlockTag(blockTag: string | number): string {
+    return typeof blockTag === 'number' ? toQuantity(blockTag) : blockTag
+  }
```
PetromirDev, line 164: Nice :)

src/libs/accountOp/balanceChanges.ts
```diff
@@ -0,0 +1,93 @@
+import { getAddress, ZeroAddress } from 'ethers'
+
+import { TokenError, TokenResult } from '../portfolio/interfaces'
+import { BalanceChange } from './submittedAccountOp'
+
+export const getBalanceChangeTokenAddresses = (tokenAddrs: string[]) =>
+  Array.from(new Set([ZeroAddress, ...tokenAddrs].map((tokenAddr) => getAddress(tokenAddr))))
```
PetromirDev: Just fyi that getAddress throws and that it's advised to wrap it in try/catch

src/libs/accountOp/balanceChanges.ts
```diff
@@ -0,0 +1,93 @@
...
+  // if the accountOp is not an atomic batch,
+  // we will have to pass the first receipt's block number
+  // we want to start the comparisson from
+  prevBlockNumber?: number
+}) => {
+  const previousBlockNumber = prevBlockNumber
+    ? prevBlockNumber
+    : receiptBlockNumber > 0
+      ? receiptBlockNumber - 1
+      : 0
```
PetromirDev, line 96: What happens when we pass block number 0 🤔

src/controllers/activity/activity.ts
```diff
@@ -413,6 +480,144 @@ export class ActivityController extends EventEmitter implements IActivityControl
     this.emitUpdate()
   }
 
+  async setAccountOpBalanceChanges(
```
PetromirDev, line 493: Using `persistAccountsOps()` can be a huge problem - `storage.set()` shouldn't be called concurrently, which is done here because of:
```
    for (const accountOp of opsToBackfill) {
      // eslint-disable-next-line no-await-in-loop
      await this.backfillAccountOpBalanceChanges(accountOp)
    }
```
Imo set in storage at the end of the operation, instead of after every op update.

PetromirDev (changes requested):

src/libs/accountOp/submittedAccountOp.ts
```diff
@@ -47,6 +48,24 @@ export type PortfoliosToUpdate = {
...
+  | 'marketDataIn'
+  | 'meta'
+  | 'flags'
+> & {
+  amount: bigint
+  amountBefore: bigint
+  amountAfter: bigint
+  balanceChange: bigint
+}
+
```
PetromirDev, line 68: Not sure if it's possible, but it may be kind of nice to mark which asset was used for gas in the UI. For example, in this txn I used ETH to pay for gas

Image

PetromirDev (changes requested):

src/libs/humanizer/modules/Uniswap/utils.ts
Oxbobby: I ran the test without the filter and found out that this token appears one more time in the humanization:
```
{
        type: 'token',
        address: '0x1111111111166b7fe7bd91427724b487980afc69',
        value: 0n,
        id: null,
        isHidden: true,
        chainId: undefined,
        editApprovalData: undefined
      }
```

But, as you said, it's isHidden: true - so it shouldn't matter.  
I'm removing the deduplication and fixing the test to expect one more time this token

src/libs/humanizer/modules/Uniswap/utils.ts
Oxbobby: You are right, this is my mistake. I'm fixing it

src/libs/humanizer/modules/Uniswap/utils.ts
Oxbobby: If you remove this code, the tests starts failing:

Screenshot 2026-04-30 at 17 08 49

So the code is adjusted to fit the tests. The tests started failing after I changed the humanization value's to zeros.

I'm open to suggestions on how to improve this. But this code is here so that the tests work properly

src/libs/accountOp/submittedAccountOp.ts
Oxbobby: I thought how to do this for a few hours and consulted with AI. Basically, it's close to impossible to make this work reliably crosschain as RPCs do not give you the final fee in the receipt (on optimism, there's no l1 fee or operator fee included in the receipt; on base there's no l1 fee; on mantle, it's a third way to calculate everything etc).

So our only hope is calculating it by ourselves and it's very risky, with close to 0% chance to get this right cross-chain.

What you do receive though is 100% accuracy on what changed in your account between blocks A & B. For me, this is incredibly valuable, especially if you are trading. And also, you as an user know if you should be receiving balance changes in ETH between these blocks - so you can easily deduct that the ETH here has been for gas.

In comparisson, Rabby is showing some numbers but they are incorrect. We also - until now we were showing some gas numbers that were incorrect. So the trade off here is: lose the gas icon but gain 100% accuracy on balance changes

src/libs/accountOp/balanceChanges.ts
Oxbobby: an error

src/libs/humanizer/modules/Uniswap/utils.ts
```diff
@@ -161,12 +198,19 @@ export const uniReduce = (_calls: HumanizerVisualization[][]): HumanizerVisualiz
...
         isSend(callJ) &&
         isSwap(callI) &&
-        callI[3].value! / 400n >= callJ[1].value!
+        callJ[1].address === callI[3].address &&
+        UNISWAP_FEE_RECIPIENT_ADDRESSES.has(getSendRecipient(callJ) || '')
       ) {
-        callI[3].value = callI[3].value! - callJ[1].value!
+        if (callI[3].value! / 400n >= callJ[1].value!) {
+          callI[3].value = callI[3].value! - callJ[1].value!
+        }
```
JIOjosBG: since the humanizer will have 0 token amounts, do we need this?

If the tests pass with this deleted, please delete it

src/libs/humanizer/modules/Uniswap/utils.ts
```diff
@@ -161,12 +198,19 @@ export const uniReduce = (_calls: HumanizerVisualization[][]): HumanizerVisualiz
...
         calls.splice(j, 1)
       }
 
+      if (callI && isSend(callI) && callI[1].address === ZeroAddress && callI[1].value === 0n) {
```
JIOjosBG: do we need the `callI[1].address === ZeroAddress`

if the tests pass without it, please remove that part of the if statement

src/libs/humanizer/modules/Uniswap/utils.ts
```diff
@@ -187,7 +231,9 @@ export const uniReduce = (_calls: HumanizerVisualization[][]): HumanizerVisualiz
         isTake(callJ) &&
         callI[3].address === callJ[1].address
       ) {
-        callI[3].value = callI[3].value! > callJ[1].value! ? callI[3].value : callJ[1].value
+        if (callI[3].value && callI[3].value > 0n) {
+          callI[3].value = callI[3].value > callJ[1].value! ? callI[3].value : callJ[1].value
+        }
```
JIOjosBG: when will we get inside this if statement
`callI` should be Swap => `callI[3]` should be token => `callI[3].value` should be 0n always?

if the tests allow, remove it

src/libs/humanizer/modules/Uniswap/utils.ts
JIOjosBG: i misunderstood, i tough you found new cases where the utils logic is insufficient. No need to add new tests with the new cases, because there are no new cases

JIOjosBG (changes requested):

src/libs/humanizer/modules/Uniswap/utils.ts
Oxbobby: Nice. Removed it and the tests are correct 👍

src/libs/humanizer/modules/Uniswap/utils.ts
Oxbobby: Nice

src/libs/humanizer/modules/Uniswap/utils.ts
Oxbobby: Nice

src/libs/accountOp/balanceChanges.ts
Oxbobby: we shouldn't be passing block number 0 so we should be safe here

src/controllers/activity/activity.ts
Oxbobby: Done, nice catch

src/controllers/storage/storage.ts
Oxbobby: Great suggestion, I implemented it. You can check it out

src/controllers/activity/activity.ts
Oxbobby: Done, you can check it out

src/controllers/storage/storage.ts
superKalo: We've discussed offline with Bobby that after the recent changes that fill in the missing bits on filter / page change, this migration here (that executes once for the last 10) maybe is no longer needed.

src/controllers/main/main.ts
```diff
@@ -643,6 +643,8 @@ export class MainController extends EventEmitter implements IMainController {
             await this.keystore.updateKeystoreKeys()
           }
         )
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        this.storage.backfillAccountOpBalanceChangesMigration(this.activity)
```
superKalo: Note: we discussed offline that the purpose of this being here (vs being in the storage controller, initiated when extension starts) is because it needs (reads from) Activity controller up and running. Logic could be made in a way it works with the raw storage instead (so that a second exception like `associateAccountKeysWithLegacySavedSeedMigration` is not needed), but we decided to explore further the option for removing the initial backfill migration anyways. So just saying.

Oxbobby:
Adding comments to the AI review:

1. This is the comment I've left on top:
```
    // find ops with no balance changes recorded and backfill them;
    // no need to console.log anything in the catch statement here
    // as error handling is handled in backfillAccountOpBalanceChangesAndPersist.
```
The error feedback is handled in backfillAccountOpBalanceChangesAndPersist on each specific location. That's why it's safe here to leave it as it is

2. Fixed
3. A very nice catch. A timeout there is not needed if we use this carefully - I'm changing this in the activity controller as it was awaiting the balance changes. The idea is that we shouldn't be awaiting the balance changes in the activity, they auto refresh the activity cards / benzina on completion

PR #2339, opened 2026-04-28
Refactor/ Keystore cipher

PetromirDev:
Closes https://github.com/AmbireTech/ambire-app/issues/6994


jordan-enev:
A few notes - I'll add more as I continue reviewing:

1. Before running the migration, we should keep a backup of the old keystore structure in storage. We'll aim for a fail-safe migration, but in case the keystore gets corrupted, having a backup would allow users to restore access. In such incidents, we could ship an update to migrate back to the backed-up keystore or provide a manual recovery page.
2. We can benefit a lot from the existing E2E tests. They always run with the current (old) keystore storage, so the migration will effectively be exercised many times, since each test boots the extension with a predefined old keystore.

src/controllers/keystore/keystore.ts
```diff
@@ -321,15 +488,15 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
       if (!this.#keystoreSecrets.length) {
-        mainKey = {
-          key: entropyGenerator.generateRandomBytes(16, extraEntropy),
-          iv: entropyGenerator.generateRandomBytes(16, extraEntropy)
-        }
+        // Generate a new main key if this is the first secret being added
+        mainKey = await crypto.subtle.generateKey({ name: CIPHER, length: 256 }, true, [
+          'encrypt',
+          'decrypt'
+        ])
```
PetromirDev: Use entropy generator

src/libs/keystore/keystore.ts
```diff
@@ -0,0 +1,264 @@
...
+        return {
+          ...storedSeed,
+          seed: await encryptWithKey(mainKey, entropy),
+          seedPassphrase: decryptedSeedPassphrase
+            ? await encryptWithKey(mainKey, decryptedSeedPassphrase)
+            : null
+        }
+      } catch (e: any) {
+        console.error(`Failed to migrate seed with id ${storedSeed.id} to AES-GCM encryption:`, e)
+        failedMigrations.seedIds.push(storedSeed.id)
```
PetromirDev, line 309: Test Sentry

src/controllers/keystore/keystore.ts
```diff
@@ -256,33 +257,71 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+        { name: CIPHER },
+        true,
+        ['encrypt', 'decrypt']
+      )
+
+      // 3. Migrate the secret and all stored seeds/pks to use GCM
+      try {
+        await this.#migrateSecretToGCM(secretEntry.id, secretKey)
+
+        await this.#migrateStoredPayloadsToGCMIfNeeded(mainKeyOld)
```
jordan-enev: I would suggest making the storage update atomic and updating both keys at once. We can introduce a new storage controller method where we pass one or multiple keys and ensure that either all keys are stored successfully or the entire operation fails.

Otherwise, if the secret is migrated but storing the payload fails, the next time we won't enter the `isOldSecretCipher` branch, and we will end up in an intermediate state.

src/controllers/keystore/keystore.ts
```diff
@@ -297,7 +336,135 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+        message: `Failed to migrate ${failedMigrations.keyAddrs.length} keys and ${failedMigrations.seedIds.length} seeds to AES-GCM encryption.`,
+        level: 'silent',
+        error: new Error(
+          `keystore: failed to migrate ${failedMigrations.keyAddrs.length} keys and ${failedMigrations.seedIds.length} seeds to AES-GCM encryption`
+        )
+      })
+    }
+
+    await this.#storage.set('keystoreKeys', migratedKeys)
+    await this.#storage.set('keystoreSeeds', migratedSeeds)
```
jordan-enev: https://github.com/AmbireTech/ambire-common/pull/2339/changes#r3240586517

src/controllers/keystore/keystore.ts
```diff
@@ -256,33 +257,71 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+      // 3. Migrate the secret and all stored seeds/pks to use GCM
+      try {
+        await this.#migrateSecretToGCM(secretEntry.id, secretKey)
+
+        await this.#migrateStoredPayloadsToGCMIfNeeded(mainKeyOld)
+      } catch (e) {
+        this.emitError({
+          message: 'Keystore migration to GCM failed.',
+          level: 'silent',
+          error: e instanceof Error ? e : new Error('keystore: GCM migration failed')
```
jordan-enev, line 309: Let's monitor it closely in Sentry with `sendCrashReport: true`, wdyt?

src/libs/keystore/keystore.ts
```diff
@@ -0,0 +1,262 @@
...
+  mainKey: MainKey,
+  mainKeyOld: MainKeyOld,
+  storedKeys: StoredKey[],
+  storedSeeds: StoredKeystoreSeed[]
+): Promise<{
+  migratedKeys: StoredKey[]
+  migratedSeeds: StoredKeystoreSeed[]
+  failedMigrations: { keyAddrs: string[]; seedIds: string[] }
+}> => {
+  const failedMigrations: { keyAddrs: string[]; seedIds: string[] } = { keyAddrs: [], seedIds: [] }
```
jordan-enev, line 242: One more thing we discussed in today's meeting: instead of having partially migrated payloads (seeds, private keys), it would be better for the migration to be complete and atomic. In other words, we should build all migration data first, and only if everything is built successfully, update the storage at once. If any step in the data building or storage update fails, we should fall back to the current CTR algorithm and notify Sentry. On the next Keystore unlock, we will attempt the migration again.

Currently, if some seeds or private keys fail during migration, the extension will work only partially (it won't be able to sign with the seeds that failed to migrate because they will remain encrypted with the old CTR algorithm), which is bad and could force us to submit a rushed patch.

src/controllers/keystore/keystore.ts
```diff
@@ -256,33 +257,71 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+    if (isOldSecretCipher) {
+      // 1. Retrieve the main key using the old encryption method
+      const mainKeyOld = this.#unlockWithSecretOld(secretKey, secretEntry)
+
+      // Derive the new main key from the old one
+      // We cannot generate a new key, because the user may have more than one secret
+      // so if we migrate another one later, we need to be able to derive the same main key again.
+      //
+      // Another option would be to generate a new random main key, encrypt it with the old main key and store it
+      // until all secrets are migrated
```
superKalo: I'd expand the context in the comments a bit more - mentioning the `mainKeyOld` is 16 bits (because prev we had 128 bytes encryption) and then - we need to derive a new main key that is 32 bits (because we now migrate to 256 bytes encryption) where derive I'm not sure is the best word because what we simply do is concatenate the old key +  iv, (each of these is 16 bits and combined = 32 bits), right?

src/controllers/keystore/keystore.ts
superKalo: It's `true` by default I think, even for silent errors, so all good.

superKalo (commented):
Notes:

- Do we need a LavaMoat policy update for using crypto.subtle.encrypt/decrypt on the -app side? Let's check when making a prod build.
- Let's investigate a prob only visual problem I spot with seed + passphrase accounts (passphrase didn't show up correctly in the UI)
- `sendPasswordEncryptedPrivateKeyToUi` and `sendPasswordDecryptedPrivateKeyToUi` still use AES-CTR, which looks like a separate mechanism (that does not need a migration) should we migrate those as well for consistency?

PR #2344, opened 2026-04-29
Fix Swap & Bridge max handling across fiat/token modes

Maskln:
closes https://github.com/AmbireTech/ambire-app/issues/5678

### Swap & Bridge

- **Fix:** When the user taps **Max** while the from-amount field is in **USD/fiat** mode, the controller now always applies the **exact max token balance** (`maxFromAmount`) and derives fiat from that. Previously, feeding **max USD** through the fiat→token path could round down and leave **spendable dust** after a swap.
- **UX:** The user’s **fiat vs token input mode** is unchanged: we temporarily use the token conversion path only for the max calculation, then restore the previous `fromAmountFieldMode`.

### DIFF
--- a/src/controllers/swapAndBridge/swapAndBridge.ts
+++ b/src/controllers/swapAndBridge/swapAndBridge.ts
@@ -897,10 +897,12 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
     }
 
     if (shouldSetMaxAmount) {
-      const maxAmountForCurrentFieldMode =
-        this.fromAmountFieldMode === 'fiat' ? this.maxFromAmountInFiat : this.maxFromAmount
-
-      this.#setFromAmountAmount(maxAmountForCurrentFieldMode, true)
+      // Always derive amounts from exact token balance. Using max fiat here
+      // would run fiat→token rounding and often leave spendable dust.
+      const previousFieldMode = this.fromAmountFieldMode
+      this.fromAmountFieldMode = 'token'
+      this.#setFromAmountAmount(this.maxFromAmount, true)
+      this.fromAmountFieldMode = previousFieldMode
     }
 
     if (fromAmount !== undefined) {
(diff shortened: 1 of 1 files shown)


src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -897,16 +899,27 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
...
-      this.#setFromAmountAmount(this.maxFromAmount, true)
+      const maxTokenAmount = this.maxFromAmount
+      const maxFiatAmount = this.maxFromAmountInFiat
+      const maxAmountForCurrentFieldMode =
+        this.fromAmountFieldMode === 'fiat' ? maxFiatAmount : maxTokenAmount
+
+      this.#setFromAmountAmount(maxAmountForCurrentFieldMode, true)
+      // Pin both representations to max regardless of active input mode.
+      this.fromAmount = maxTokenAmount
+      this.fromAmountInFiat = maxFiatAmount
```
PetromirDev: Hmm, I don't think that this is what Ivo meant in the issue - yes, we are setting isMax to true but it's not used anywhere 🤔 Sorry, but I don't understand how this is different than simply calling:
```
this.#setFromAmountAmount(maxAmountForCurrentFieldMode, true)
```
without:
```
// Pin both representations to max regardless of active input mode.
      this.fromAmount = maxTokenAmount
      this.fromAmountInFiat = maxFiatAmount
```
and without `this.isMax`

PetromirDev (changes requested):

src/controllers/swapAndBridge/swapAndBridge.ts
Maskln: Yes, you're right.
I missed to clean up form the previous tries where i used  — `isMax`. 
I’ve removed both and kept only the `#setFromAmountAmount(...)` flow.

PetromirDev (changes requested):
I'm not sure if we should go through with this one. I remember people complaining about "dust" being left over after sending and it's still the case

Image

Try sending max ADX 0xbd8551AeE378bca7c2964A7a2293B728680524B7 with 'usd' selected

PetromirDev (approved):

PR #2345, opened 2026-04-29
Security / dApp verification for SignMessage and SignAccountOp screens

jordan-enev:
Paired with: https://github.com/AmbireTech/ambire-app/pull/7052


src/controllers/dapps/dapps.ts
```diff
@@ -801,6 +802,106 @@ export class DappsController extends EventEmitter implements IDappsController {
...
+    ) =>
+      Array.from(new Set(dappVerificationData.filter(predicate).map((dapp) => dapp.name))).join(
+        ', '
+      )
+
+    // Appends dApp names to the message only when enabled, formatting "...requests." as "...requests: <dApp names>".
+    const withOptionalDappNames = (baseText: string, dappNames: string) => {
+      if (!includeDappNamesInText || !dappNames) return baseText
+      return `${baseText.replace(/\.$/, ':')} ${dappNames}`
+    }
```
PetromirDev: Kind of hackish, because it replaces the last dot and might break in the future.

src/controllers/dapps/dapps.ts
```diff
@@ -801,6 +802,106 @@ export class DappsController extends EventEmitter implements IDappsController {
...
+          dappNames
+        )
+      }
+    }
+
+    // 2) dApp verification failed / unknown
+    if (dappVerificationData.some((dapp) => dapp.status === 'FAILED_TO_GET' || !dapp.status)) {
+      const dappNames = getDappNamesByPredicate(
+        (dapp) => dapp.status === 'FAILED_TO_GET' || !dapp.status
+      )
```
PetromirDev: Instead of doing .some and then const dappNames =...
we can do it simpler:
```
const failedOrUnknownNames = getDappNamesByPredicate(
        (dapp) => dapp.status === 'FAILED_TO_GET' || !dapp.status
      )

if (failedOrUnknownNames.length)
```

src/controllers/signMessage/signMessage.ts
```diff
@@ -546,4 +552,34 @@ export class SignMessageController extends EventEmitter implements ISignMessageC
...
   }
+
+  #getDappVerificationBanner(): DappVerificationBanner | null {
+    if (!this.#dapps || !this.dapp?.url) return null
+
+    const banner = this.#dapps.getDappVerificationBanner([this.dapp.url.toLowerCase()], {
+      // SignMessage operates on a single dApp, and the request window already shows it,
+      // so repeating the dApp name in the banner text adds noise.
+      includeDappNamesInText: false
+    })
```
PetromirDev, line 567: Nice :)

PetromirDev (changes requested):
Let's add some positive/negative lower level unit tests and at least 1 integration test in sign message and sign account op

src/controllers/dapps/dapps.ts
superKalo: We can just skip doing this and use the name as is (which would be beautified anyways if the app is part of our default catalog: `name: dapp?.name || new URL(url).hostname`)

superKalo (approved):
Two suggestions:

1. Let's add unit test for `#getDappVerificationBanner` logic? So we make sure it doesn't break in future.

2. I'm not too happy with the wordings yet. They sound too soft IMO. I called @dmngdsm for help.

src/controllers/dapps/dapps.ts
jordan-enev: @PetromirDev, it's a local function, and if someone changes the message (baseText), they will most likely follow the current text format.

However, I've replaced the regex, and now the logic will always add the colon. If baseText ends with a dot or exclamation mark, it will be replaced. If it ends with anything else - no punctuation mark or something unusual (a real edge case) - it will still append a colon.

It's not perfect (since baseText could end with ";"), but that's a real edge case, and it's pretty clear what text format baseText should follow.

PR #2348, opened 2026-04-30
Feature: Squid router for Citrea

Oxbobby:
Change log:
* we are no longer requiring a min of 10$ to bridge from/to Ethereum as we are resilient and bridges work fine


Oxbobby:
Addressing AI comments:
1. Squid uses intents so a swap there is also correctly categorized as a bridge. We even display a banner for it
2. This is OK
3. This I also think is OK

src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -2590,18 +2591,6 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
       })
     }
 
-    // if we're bridging to ethereum, make the min from amount 10 usd
-    if (
-      isBridge &&
-      this.toChainId === 1 &&
-      this.fromAmountInFiat &&
-      Number(this.fromAmountInFiat) < 10
-    ) {
```
superKalo, line 2606: Please add this change to the PR descr 🙏

src/services/lifi/api.ts
```diff
@@ -285,6 +286,11 @@ export class LiFiAPI implements SwapProvider {
     this.isHealthy = null
   }
 
+  /** disable explicitly citrea for lifi */
+  areChainsSupported({ fromChainId, toChainId }: { fromChainId: number; toChainId: number }) {
+    return fromChainId !== CITREA_CHAIN_ID && toChainId !== CITREA_CHAIN_ID
+  }
```
superKalo, line 292: Why would we want to disable Citrea for LiFi (and for Socket)?

src/services/swapIntegrators/swapProviderParallelExecutor.ts
```diff
@@ -58,6 +58,13 @@ export class SwapProviderParallelExecutor {
     const supportedProviders = this.#providers.filter((provider) => {
       // If the request is not chainId specific, use all providers
       if (!uniqueChainIds.length) return true
+      if (reqMeta?.chainIds?.length === 2 && provider.areChainsSupported) {
+        return provider.areChainsSupported({
+          fromChainId: reqMeta.chainIds[0]!,
+          toChainId: reqMeta.chainIds[1]!
+        })
+      }
```
superKalo, line 66: Loosely related, I was wondering if it's only 1 provider that supports this chain(s) in the Parallel provider, can we easily write the messages singular?

Image

src/services/squid/api.ts
```diff
@@ -0,0 +1,495 @@
...
+  SwapProvider
+} from '../../interfaces/swapAndBridge'
+import {
+  addCustomTokensIfNeeded,
+  convertPortfolioTokenToSwapAndBridgeToToken,
+  getSlippage,
+  isNoFeeToken,
+  sortNativeTokenFirst
+} from '../../libs/swapAndBridge/swapAndBridge'
+import { ZERO_ADDRESS } from '../socket/constants'
```
superKalo: Minor, it's a bit strange to take this from socket constants, I'd use Ethers or View import instead (they both have ZeroAddress constant).

superKalo (approved):
Quite a lot of complexity, but I tested with bridges and swaps and couldn't find glitches.

src/services/lifi/api.ts
Oxbobby: I tested and I saw that we're making `quote` requests to lifi / socket even though they are providers that don't support the chain. That's why I made this to make sure no requests to them are made

src/services/swapIntegrators/swapProviderParallelExecutor.ts
Oxbobby: We can, yes. I'll do it and I'm resolving this

PR #2349, opened 2026-05-04
Fix/abstract chain

Oxbobby:
Goal: make the abstract chain work

Problems:
* blocks in abstract are returned without a signature, causing ethers to revert with a "missing r" parameter
* txn.origin during eth_call is always 0, causing our simulation/estimation to revert with incorrect spoof

Solutions:
* implement a fetchRawBlock for abstract only that fetches the block by doing directly a "eth_getBlockByNumber"
* use executeBySender() instead of execute() during estimations/simulations for the abstract chain only by state overriding the balance getter to have privileges


Oxbobby:
closing in favour of another pr for fixing abstract

PR #2354, opened 2026-05-05
Survey/ add retry mechanisms for relayer fetches

JIOjosBG:
When the req fails for fetching the survey or submitting the response, simply retry 2 times with 1s wait inbetween


PetromirDev (approved):
We agreed that this is a temporary solution for the release. A long term solution would be to fetch only once and have a retry button in the UI

PR #2363, opened 2026-05-08
Feature/bridge activity card

Oxbobby:
Changes:
* add an SubmittedAccountOpLike interface that accepts the minimum properties needed to display successfully a transaction in the activity tab
* add an externalAccountOps object in storage
* when a bridge completes, add that transaction as an externalAccountOp for that account
* mix internal & external account ops in the activity so they are visible to the user
* move debug_traceTransaction to its own declaration


src/controllers/activity/activity.ts
```diff
@@ -70,6 +86,11 @@ export interface InternalAccountsOps {
   [key: string]: { [key: string]: SubmittedAccountOp[] }
 }
 
+export interface ExternalAccountOps {
+  // account => network => SubmittedAccountOpLike[]
+  [key: string]: { [key: string]: SubmittedAccountOpLike[] }
+}
+
```
PetromirDev, line 92: You can rewrite the type as follows:
```
export interface ExternalAccountOps {
  [account: string]: { [network: string]: SubmittedAccountOpLike[] }
}
```
which removes the need for comments

src/controllers/activity/activity.ts
```diff
@@ -540,6 +590,102 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+    chainId,
+    txnId,
+    receipt,
+    callId
+  }: AddExternalAccountOpParams) {
+    await this.#initialLoadPromise
+
+    const network = this.#networks.networks.find((n) => n.chainId === chainId)
+    const provider = this.#providers.providers[chainId.toString()]
+    if (!network || !provider) return
```
PetromirDev: Emit a silent error

src/controllers/activity/activity.ts
```diff
@@ -540,6 +590,102 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+        )
+      })
+    } catch (error) {
+      submittedAccountOpLike.balanceChanges = undefined
+    }
+
+    this.#externalAccountOps = await this.#storage.get(
+      'externalAccountOps',
+      this.#externalAccountOps
+    )
```
PetromirDev: Imo we shouldn't read before writing. I don't think that we do it anywhere else - the latest value is the state.

src/controllers/activity/activity.ts
```diff
@@ -540,6 +590,102 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
 
+  async addExternalAccountOp({
+    accountAddr,
+    chainId,
+    txnId,
+    receipt,
+    callId
+  }: AddExternalAccountOpParams) {
+    await this.#initialLoadPromise
+
```
PetromirDev, line 627: We haven't done it for `addAccountOp`, I know, but do you think that we need a queue for this logic? To prevent race conditions in case the method is called twice in the same tick

src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -1855,10 +1856,32 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
...
+    await this.#initialLoadPromise
+
+    const chainId = activeRoute.route?.toChainId
+    if (!chainId) return
+
+    const provider = this.#providers.providers[chainId.toString()]
+    if (!provider) return
+
+    const receipt = await provider.getTransactionReceipt(txnId)
+    if (!receipt) return
```
PetromirDev: Let's log some silent errors in case of all return statements?

src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -1907,6 +1931,15 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
...
 
+      if (
+        routeStatusResult.txnId &&
+        (status === 'completed' || status === 'refunded') &&
+        activeRoute.route?.fromChainId !== activeRoute.route?.toChainId
+      ) {
+        // we shouldn't be awaiting this as it's OK to have it at a later stage
+        this.recordBridgeActivity(routeStatusResult.txnId, activeRoute).catch(console.error)
+      }
+
```
PetromirDev: We call `recordBridgeActivity` even if the status is `refunded`. Afaik, refunds are done by sending back the `fromToken` on the source chain (`fromChainId`). In case of status `refunded`, we will call `recordBridgeActivity`, which always works with `toChainId`.

PetromirDev (changes requested):

PR #2364, opened 2026-05-10
Feature/txn logs

Oxbobby:
Implement a logs controller that starts searching for transfer logs appearing on the blockchain. It tracks them for a limited time of 3 minutse on a trigger. The triggers are:
* when a transaction interacting with a dapp has been signed
* when an EIP-712 message from a dapp has been signed

The problem?  
Dapps like cowswap/1inch work with intents or orders. It means the user receives the funds in a future transaction. Happening a few seconds to minutes after the initial order has been given.  
The problem with that is that the wallet doesn't have a clue that such a thing is in place and:
* a manual portfolio refresh is needed to understand the balance changes
* sometimes a manual portfolio refresh is not enough as the transferred in token is not detected by velcro

The fix.  
Start searching the blockchain for transfer logs, record them as an externalAccountOp if found, learn the transferred in tokens from them & show an activity card for the transaction

Why only on a dapp interaction?  
Because inner swaps/bridges have this logic already implemented


.gitignore
```diff
@@ -37,6 +37,8 @@ Temporary Items
 ### Node ###
 # Logs
 logs
+!src/controllers/logs/
+!src/controllers/logs/logs.ts
```
superKalo: Ah, gotcha! Same problem if we have a lib with `logs.ts` down the road. But how about instead if we name the controller "TransfersScannerController"?

This way we will have:

- ActivityController (which receives discovered ops)
- TransfersScannerController (discovers ERC20 transfers via logs)

Sounds better to be than the generic LogsController (where logs could mean anything). Wdyt?

src/controllers/logs/logs.ts
```diff
@@ -0,0 +1,208 @@
...
+    await this.#portfolio.updateSelectedAccount(accAddr, [network])
+
+    return { nextFromBlock, txnIds }
+  }
+
+  startScanLogsLoop({ accAddr, chainId, fromBlock = 'latest' }: Omit<ScanLogsParams, 'toBlock'>) {
+    const chainIdString = chainId.toString()
+    this.#scanLoopId += 1
+    const scanLoopId = this.#scanLoopId
+    this.#activeScanLoopIdsByChain[chainIdString] = scanLoopId
```
superKalo: This stores only 1 active loop per chain.

1. What would happen if the user signs two dapp intents on the same network in these 3 minutes? Could this be a problem? Could it skip a beat (the "settlement" txn-а of the first loop)?

2. Same concern with account switching. If account A signs a dapp intent, then the user switches to account B and signs another dapp intent on the same network, account B's loop appears to cancel account A's loop?

src/controllers/main/main.ts
```diff
@@ -1036,6 +1047,22 @@ export class MainController extends EventEmitter implements IMainController {
       )
     }
 
+    // signing typed messages might trigger a txn
+    if (signedMessage.content.kind === 'typedMessage') {
```
superKalo, line 1055: Small question: do we know that every `typedMessage` reaching `#resolveSignMessage` here is always from a dapp request?

I wonder if we should check if `signedMessage.dapp` exists (is different than `null`). But if this path is only used for dapp message requests, then no need and all good.

src/controllers/logs/logs.ts
```diff
@@ -0,0 +1,208 @@
...
+type ScanLogsResult = {
+  nextFromBlock: number
+  txnIds: string[]
+}
+
+function topicAddress(address: string) {
+  return `0x${address.toLowerCase().replace(/^0x/, '').padStart(64, '0')}`
+}
+
+export class LogsController extends EventEmitter {
```
superKalo: When we finalize the discussions around this controller logic, I'd encourage adding unit tests for the core scan-loop behavior: recording matching transfer logs, learning tokens, retrying the same range after `getLogs` failures, and making sure overlapping scans don't accidentally cancel valid account/order scans.

.gitignore
Oxbobby: Sounds fair. On it

src/controllers/transfersScanner/transfersScanner.ts
```diff
@@ -0,0 +1,208 @@
...
+    this.#networks = networks
+    this.#portfolio = portfolio
+    this.#providers = providers
+  }
+
+  async scanLogs({
+    accAddr,
+    chainId,
+    fromBlock = 'latest'
+  }: ScanLogsParams): Promise<ScanLogsResult | null> {
```
superKalo, line 128: Could this scan pick up unrelated or scam token transfers during the 3-minute window?

Right now we record any ERC-20 `Transfer` log where the account is either `from` or `to`. If a dust/scam token is transferred to the account while the scan loop is active, it looks like we may record it as an `externalAccountOp` and also pass the token to `addTokensToBeLearned`?

Do we have any guardrails against getting unrelated/scam transfers here? Maybe we should consider some additional magic filtering before learning unknown tokens?

src/controllers/main/main.ts
Oxbobby: Yes, the signMessage controller as a whole is initialized only by dapps so I think we're safe here 👍

src/controllers/logs/logs.ts
Oxbobby: 1. Yes, it could. My concern with this is that we shouldn't easily allow multiple loops to run and that's why I've limited it to a max of 1 per chain. If a second one comes, it overrides the first, and we run the risk of skipping a beat with this code.

2. But your second point is fair and it should be at least one per account<->chain combination. I will edit this and let's talk out the first point if we can do a better solution

src/controllers/transfersScanner/transfersScanner.ts
Oxbobby: Currently, we don't have an easy solution for this except the already implemented ones: don't show the price for scam tokens, sending them to the bottom of the portfolio

src/controllers/logs/logs.ts
Oxbobby: I fixed both: we make sure the earliest block number is used in case of a collision; and I make the loop run once per chain-account combo. There's a unit tests for this as well

src/controllers/logs/logs.ts
Oxbobby: Added unit tests for various cases

PR #2366, opened 2026-05-11
Fix/abstract chain rpc calls

Oxbobby:
Change log:
* a new AmbireAccountAbstract.json file used for doing state overrides on the Abstract chain
* fix: gas price fetch on Abstract
* fix: balance changes fetch on Abstract
* fix: simulations on Abstract

We fix everything by allowing address 0 to be a valid spoof address


Oxbobby:
@Ivshti said he's OK with merging this

PR #2367, opened 2026-05-11
Update/ PR review config

PetromirDev:
### Reduces noise:
Actions were ran and skipped on every PR, which isn't needed
image

### Adds fallback models:
Kimi failed to respond in https://github.com/AmbireTech/ambire-common/pull/2358#issuecomment-4395571215 and PR agent used `gpt-5.4-mini` as fallback (default), which isn't adequate for our use cases.

### Updates system prompts:
Adds a couple of extra issues to system prompts.

### DIFF
--- a/.github/workflows/general-review.yml
+++ b/.github/workflows/general-review.yml
@@ -2,7 +2,7 @@ name: PR Agent - General Review
 
 on:
   pull_request:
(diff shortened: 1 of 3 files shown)


superKalo (approved):
Sounds good to me 👍

PR #2372, opened 2026-05-11
Fix / dApp not-present-in-catalog verification + integration/unit tests

jordan-enev:
- Fix: Verification for dApps not present in the default catalog was never reachable. Updated the logic to return the appropriate banner.
- Add: Integration tests for DappsController->getDappVerificationBanner().
- Add: Unit tests for SignAccountOpController and SignMessageController getDappVerificationBanner(). Here, we implemented unit tests because the main logic is already covered by the DappsController tests, and in both controllers we only test the branching logic.


superKalo (approved):
Great that we wrote tests, because they caught a real trouble 👏

PR #2377, opened 2026-05-13
Feature / Wallet Connect Mobile

sonytooo:
* add helper functions and props for the WC feature


superKalo (approved):

src/controllers/dapps/dapps.ts
```diff
@@ -551,6 +562,18 @@ export class DappsController extends EventEmitter implements IDappsController {
         }
       }
     })
+
+    // on disconnect clean up the WC sessions
+    if (ev === 'disconnect') {
+      dappSessions.forEach((dappSession) => {
+        if (
+          this.dappSessions[dappSession.sessionId] &&
+          this.dappSessions[dappSession.sessionId]?.wcTopic
```
superKalo: Minor, I think this can be simplified to just `this.dappSessions[dappSession.sessionId]?.wcTopic` (because the `?` does the trick if `this.dappSessions[dappSession.sessionId]` is missing.

src/controllers/dapps/dapps.ts
```diff
@@ -551,6 +562,18 @@ export class DappsController extends EventEmitter implements IDappsController {
...
+        ) {
+          this.deleteDappSession(dappSession.sessionId)
```
superKalo, line 570: Should we also `deleteDappSessionByWcTopic`?

I assume prob no need because this works with the generic disconnect flow.

sonytooo: deleteDappSessionByWcTopic this is just a helper function to remove temp wc sessions on connect so no need to call it here.
With WC we work with 2 sessions. We have a temp session while the app is not connected but still can make requests. Once connected we receive a new session from WC that we persist in dappSessions and after that we remove this temp session

PR #2379, opened 2026-05-13
Return the correct txnId to the dapp in the case of a 4337 txnId replacement

Oxbobby:
If there'a a 4337 txnId replacement, we ran the risk of returning an incorrect txnId to the dapp


superKalo (approved):
With the last iteration, looks strong & beton 👍

PR #2381, opened 2026-05-13
Feature/clear signing

Oxbobby:
## Summary

Adds ERC-7730 relayer-backed descriptor resolution and improves calldata/EIP-712 humanization logic.

## What changed

- Fetches ERC-7730 calldata and EIP-712 indexes through the relayer only.
- Adds in-memory caching for calldata indexes, EIP-712 indexes, and descriptors with stale-cache fallback.
- Adds stricter response validation for relayer registry payloads.
- Resolves Safe proxy singleton implementations dynamically for SafeTx and execTransaction flows.
- Supports SafeTx single-call and multisend decoding.
- Supports nested calldata humanization, including:
  - Safe multisend transactions
  - Safe execTransaction inner transactions
  - 1inch execute-with-permit inner swap calldata
- Improves fallback behavior for Safe-specific calls via existing Safe humanization.
- Adds/updates tests for registry caching, malformed relayer responses, Safe flows, Morpho multicall, 1inch EIP-712 orders, and nested calldata decoding.

## Testing

- `npm run jest -- --runInBand src/libs/humanizer/erc7730/registry.test.ts`
- `npm run jest -- --runInBand src/libs/humanizer/index.test.ts`
- Focused eslint checks on touched ERC-7730/humanizer files


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -768,19 +777,91 @@ export class SignAccountOpController extends EventEmitter implements ISignAccoun
...
+  #startHumanization() {
+    return this.startHumanization((humanizationId) => {
+      this.isHumanizing = true
+      this.humanization = []
+      this.humanizationId = humanizationId
+    })
+  }
+
+  #setFallbackHumanization(humanizationId: number) {
+    if (!this.isCurrentHumanization(humanizationId) || this.humanizationId !== humanizationId) {
```
JIOjosBG, line 799: Do we need this.humanizationId? Isn't this.isCurrentHumanization sufficient in all places

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -768,19 +777,91 @@ export class SignAccountOpController extends EventEmitter implements ISignAccoun
...
+    return true
+  }
+
+  #setErc7730Humanization(
+    humanizationId: number,
+    erc7730Descriptors: Awaited<ReturnType<typeof fetchErc7730DescriptorsForAccountOp>>
+  ) {
+    if (
+      !this.isCurrentHumanization(humanizationId) ||
+      this.humanizationId !== humanizationId ||
```
JIOjosBG, line 815: again i do not thunk we need `this.humanizationId !== humanizationId`

src/libs/humanizer/erc7730/registry.ts
```diff
@@ -0,0 +1,898 @@
...
+      return descriptor ? ([index, descriptor] as const) : null
+    })
+  )
+
+  return resolvedDescriptors.reduce<Record<number, Erc7730ResolvedDescriptor>>((acc, entry) => {
+    if (!entry) return acc
+
+    acc[entry[0]] = entry[1]
+    return acc
+  }, {})
```
JIOjosBG: here Object.fromEntries would have been easier to follow

src/libs/humanizer/erc7730/registry.ts
```diff
@@ -0,0 +1,898 @@
...
+}
+
+export const fetchErc7730DescriptorsForAccountOp = async (
+  accountOp: AccountOp,
+  options?: Erc7730RegistryOptions | BindedRelayerCall
+): Promise<Record<number, Erc7730ResolvedDescriptor>> => {
+  const resolvedDescriptors = await Promise.all(
+    accountOp.calls.map(async (call, index) => {
+      const descriptor = await fetchErc7730DescriptorForCall(call, accountOp.chainId, options)
+      return descriptor ? ([index, descriptor] as const) : null
```
JIOjosBG: instead of 'as const' you could have added the return type of the function to be [number, descripto] | null

src/libs/humanizer/erc7730/registry.ts
```diff
@@ -0,0 +1,898 @@
...
+  const descriptorPath = index[getRegistryKey(chainId, safeSingleton)]
+  if (!descriptorPath) return null
+
+  return fetchDescriptor(descriptorPath, callRelayer)
+}
+
+export const fetchErc7730DescriptorForCall = async (
+  call: Call,
+  chainId: AccountOp['chainId'],
+  options?: Erc7730RegistryOptions | BindedRelayerCall
```
JIOjosBG: why do you pass options or BindedRelayerCall? This seems unintuitive for an argument called options

JIOjosBG: can you place the callRelayer inside options?

src/libs/humanizer/erc7730/registry.ts
```diff
@@ -0,0 +1,898 @@
...
+): Promise<Erc7730ResolvedDescriptor | null> => {
+  if (!call.to || !isAddress(call.to)) return null
+
+  const registryOptions = normalizeRegistryOptions(options)
```
JIOjosBG: if you can pass the callRelayer to the options you will not need the `normalizeRegistryOptions` function at all

src/libs/humanizer/erc7730/registry.ts
```diff
@@ -0,0 +1,898 @@
...
+    }
+  } catch (error) {
+    console.error(error)
+    return builtInDescriptor
+  }
+}
+
+export const fetchErc7730DescriptorsForAccountOp = async (
+  accountOp: AccountOp,
+  options?: Erc7730RegistryOptions | BindedRelayerCall
```
JIOjosBG: can the binded relayer call be inside options or as separate optional  arg?

src/libs/humanizer/erc7730/registry.ts
```diff
@@ -0,0 +1,898 @@
...
+): Promise<Record<number, Erc7730ResolvedDescriptor>> => {
+  const resolvedDescriptors = await Promise.all(
+    accountOp.calls.map(async (call, index) => {
+      const descriptor = await fetchErc7730DescriptorForCall(call, accountOp.chainId, options)
```
JIOjosBG, line 857: in cases where i make a batch of revoke approves will this make one relayer fetch for each call?

JIOjosBG: also won't this leak what the user is interacting with? Should this have a privacy opt-out

JIOjosBG: it is also not clear to me if 7730 functionality will lead to many more network calls, should we persist the jsons in the storage somehow?

src/libs/humanizer/erc7730/types.ts
```diff
@@ -0,0 +1,97 @@
...
+  call: Call,
+  chainId: AccountOp['chainId'],
+  options?: Erc7730RegistryOptions
+) => Promise<Erc7730ResolvedDescriptor | null>
+
+export type FetchErc7730DescriptorForMessage = (
+  message: Message,
+  options?: Erc7730RegistryOptions
+) => Promise<Erc7730ResolvedDescriptor | null>
+
```
JIOjosBG: what are those types used for?

src/libs/accountOp/accountOp.ts
```diff
@@ -103,6 +103,8 @@ export interface AccountOp {
     speedUp?: {
       enabled: boolean
     }
+    /** Descriptor-backed humanization persisted so activity, history and Benzin show the same clear signing details. */
+    clearSigningHumanization?: unknown
```
JIOjosBG: why unknown?

src/libs/humanizer/erc7730/eip712.ts
```diff
@@ -0,0 +1,51 @@
+import { keccak256, toUtf8Bytes } from 'ethers'
+
+import { Erc7730TypedDataTypes } from './types'
+
+const ARRAY_SUFFIX_REGEX = /\[[^\]]*\]/g
```
JIOjosBG: no regex pls

src/libs/humanizer/erc7730/fetch.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  if (response?.success === undefined) return response as T
+
+  const { success, status, errorState, message, ...payload } = response
+  if (typeof payload !== 'object' || payload === null || Array.isArray(payload)) {
+    throw new Error(`Invalid ERC-7730 relayer resource response: ${path}`)
+  }
+
+  return payload as T
+}
```
JIOjosBG, line 20: 1. why have custom function for this, we already have a function for calling the relayer
2. casting to arbitrary type T is not safe

JIOjosBG: i see there is a `validate` function call later, but it will be easy to forget it if someone touches the code

src/libs/humanizer/erc7730/fetch.ts
```diff
@@ -0,0 +1,60 @@
...
+      message: `Timed out fetching ERC-7730 relayer resource: ${path}`
+    }
+  )
+  const payload = getRelayerPayload<T>(response, path)
+
+  validate(payload, path)
+
+  return payload
+}
+
```
JIOjosBG: we already have a function for fetching data for the relayer

src/libs/humanizer/erc7730/registry.ts
```diff
@@ -0,0 +1,938 @@
...
+): Erc7730RegistryOptions => {
+  if (typeof options === 'function') return { callRelayer: options }
+
+  return options || {}
+}
+
+const isPlainObject = (value: unknown): value is Record<string, unknown> =>
+  typeof value === 'object' && value !== null && !Array.isArray(value)
+
+const isRegistryKey = (key: string): boolean => /^eip155:\d+:0x[a-fA-F0-9]{40}$/.test(key)
```
JIOjosBG: pls no regex

src/libs/humanizer/erc7730/registry.ts
```diff
@@ -0,0 +1,938 @@
...
+
+const isHexHash = (value: string): boolean => /^0x[a-fA-F0-9]{64}$/.test(value)
```
JIOjosBG: again regex, you can use `isHex` from viem and separate length check

src/libs/humanizer/erc7730/registry.ts
```diff
@@ -0,0 +1,938 @@
...
+
+const throwInvalidRelayerResource = (path: string): never => {
+  throw new Error(`Invalid ERC-7730 relayer resource response: ${path}`)
+}
+
+const validateCalldataIndex = (payload: unknown, path: string): payload is Erc7730CalldataIndex => {
+  if (!isPlainObject(payload)) throwInvalidRelayerResource(path)
+
+  const index = payload as Record<string, unknown>
```
JIOjosBG: avoid casting

JIOjosBG (changes requested):
i have not reviewed all of the code, I will create separate PR resolving my own comments + I want to tinker further to understand the code better

src/controllers/humanization/humanization.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  protected startHumanization(onStart: (humanizationId: number) => void) {
+    const humanizationId = this.createHumanizationId()
+
+    onStart(humanizationId)
+    this.emitUpdate()
+
+    return humanizationId
+  }
```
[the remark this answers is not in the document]
JIOjosBG: actually since this controller is used in two places it will be cleaner to have this debuggable in one place. Do not change

src/controllers/signAccountOp/signAccountOp.ts
JIOjosBG: looks like it is leftover from refactoring

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -768,19 +777,91 @@ export class SignAccountOpController extends EventEmitter implements ISignAccoun
...
+    }
+
+    const erc7730Humanization = humanizeAccountOp(this.accountOp, { erc7730Descriptors })
+    if (
+      !hasErc7730Humanization(erc7730Humanization) ||
+      !this.isCurrentHumanization(humanizationId) ||
+      this.humanizationId !== humanizationId
+    ) {
+      return false
+    }
```
JIOjosBG, line 828: 1. you already have check for `this.isCurrentHumanization` aboev
2. why not set the humanization if `!hasErc7730Humanization`? I think you can set it always

JIOjosBG (changes requested):
feedback from claude

Image
  1 High        erc7730/humanize.ts:336                  return null in catch aborts the format-key loop; should be continue                            
  
2 Medium      erc7730/registry.ts:610                  Storage-slot address guard < 40 should be < 42; 40-char strings crash getAddress               
  
3 Medium      humanize.ts:816 vs registry.ts:526       getSafeTxCallsFromMessage duplicated with divergent guards (startsWith('0x') vs isHexString)   

  4 Low         erc7730/consts.ts:9                      Naming typo: SAFE_SINGLETOn_CACHE_TTL_MS — lowercase n in SINGLETON                        

  5 Low         requests/requests.ts:1280                console.log(e) should be console.error(e) — violates project rule, hides from monitoring       

  6 Low         erc7730/registry.ts:329                  Double withTimeout(4000ms) — outer wrapper's error message is unreachable; inner always fires first                     

  7 Latent      erc7730/fetch.ts:37                      validate() return value discarded — safe today (validators throw), breaks if any future validator returns false         

  8 Low         signMessage/signMessage.ts:307           Non-typedMessage path skips #startHumanization(), so isHumanizing is never true — inconsistent contract vs signAccountOp

  9 Cleanup     erc7730/humanize.ts:71 + registry.ts     Naming inconsistency: same type guard named isRecord vs isPlainObject; should be one shared util                        

  10 Convention registry.ts:3626, eip712.ts, humanize.ts Regex literals throughout the new module violate the AGENTS.md rule "NEVER use regex"

src/libs/humanizer/utils.ts
```diff
@@ -68,8 +74,47 @@ export function getChain(chainId: bigint): HumanizerVisualization {
...
+}
+
+export function getPersistedHumanization(meta?: {
+  clearSigningHumanization?: unknown
+}): IrCall[] | null {
+  const humanization = meta?.clearSigningHumanization
+  if (!Array.isArray(humanization)) return null
+
+  const persistedHumanization = humanization as IrCall[]
+  return hasErc7730Humanization(persistedHumanization) ? persistedHumanization : null
```
JIOjosBG: You do not need this function, it adds an extra layer for future debugging.

lets remove it and just use hasErc7730Humanization(persistedHumanization) in all places + comment

why make it persistent when 1 call has a 7730?

JIOjosBG (changes requested):

JIOjosBG:
i have not reached that part of the code, but from privacy standpoint, won't this leak the exact contract the user is interacting with? If so, we should either add opt-out or figure out how to download the whole thing in chunks at some point

src/libs/humanizer/erc7730/registry.ts
Oxbobby: yes, resolving this one

src/libs/humanizer/erc7730/registry.ts
Oxbobby: We have a memory cache for 24h set so if the data is in the cache, we won't be making extra calls.  
As for the revoke example - it will make one fetch and then it will reuse it from cache.  
Privacy opt out - you're fetching this info from the relayer so the relayer would be the only one to know. I think it's too early to think about an extension without the relayer atm. We can't put the registries/descriptors in storage as they might change or new ones might get introduced (they will, there are PRs open for new registries - and that's the goal, to expand)

src/libs/accountOp/accountOp.ts
Oxbobby: resolved

src/libs/humanizer/utils.ts
Oxbobby: resolved

src/libs/humanizer/erc7730/fetch.ts
Oxbobby: Yes. I optimized things a bit to have only 1 wrapper method, but it encapsulates validate, and I'm thinking of keeping this because of validate. But feel free to change this if you find a better solution

src/libs/humanizer/erc7730/fetch.ts
Oxbobby: let's keep the discussion for this here: https://github.com/AmbireTech/ambire-common/pull/2381#discussion_r3323867435

src/libs/humanizer/erc7730/types.ts
Oxbobby: there were 2 leftover types

src/controllers/signAccountOp/signAccountOp.ts
JIOjosBG: i might be wrong, leaving this as is

src/controllers/signAccountOp/signAccountOp.ts
JIOjosBG: i might be wrong

JIOjosBG:
> 2 Medium erc7730/registry.ts:610 Storage-slot address guard < 40 should be < 42; 40-char strings crash getAddress

This was hallucination, maybe we can add some web3 skill to claude. the slot is actually bigger

image

JIOjosBG (approved):
lets ship, but we need tests for humanization both in the signAccOp controller and signMessage controller for:
- full flow of data 
 - loading states
 - successfully using 7730
 - fallback to old humanizer
- using the cache
- not making multiple fetches for the same descriptor when, for example, there are 10 approve calls in the same batch

PR #2383, opened 2026-05-14
Types/ Make storage get and set strict and accurate; add missing storage keys

PetromirDev:
## How I found this:
1. Users reported that dismissing the defi banner doesn't persist - you see it again almost immediately
2. I reproduced it immediately
3. Logged the value of `dismissedBannerIds` and saw an array with length 0 that has a key inside??? (like an object)
4. Found out that we set the default value of `dismissedBannerIds`
5. Was confused why TS doesn't warn about this
6. Saw `any`...

## Why it's not that simple to fix this:
I wanted to get the following result:
1. A warning is shown if a defaultValue is not provided (can be undefined)
2. A warning is shown if a defaultValue is provided but is of the wrong type (can be StorageProps[K])
3. A warning is shown if a defaultValue is explicitly set to null (can be StorageProps[K] or null), but the same warning is not shown if a correct default value is provided

and also, a lot of storage keys were missing
## Changes:
- Fixes the typescript
- Adds a migration that fixes the array bug
- Fixes TS errors


sonytooo (approved):

superKalo (approved):
👏

superKalo:
Closing it, because these changes went in the release branch first in https://github.com/AmbireTech/ambire-common/pull/2387, then in `main`, then in `v2` ✅

PR #2388, opened 2026-05-15
Config / Grant PR-Agent read issues for ticket context in reviews

superKalo:
Adds `issues: read` to the PR-Agent general and security review workflows so ticket/issue context can be fetched when `require_ticket_analysis_review` is enabled in `.pr_agent.toml`.

PR-Agent resolves linked GitHub issues (PR description, branch name, `#123`) for ticket compliance analysis. Without `issues` scope, `get_issue()` can fail with 403 even in the same private repo.

See https://docs.pr-agent.ai/installation/github/

### DIFF
--- a/.github/workflows/general-review.yml
+++ b/.github/workflows/general-review.yml
@@ -17,6 +17,7 @@ jobs:
       contains(github.event.comment.body, '/review'))
     runs-on: ubuntu-latest
     permissions:
+      issues: read
       pull-requests: write
       contents: write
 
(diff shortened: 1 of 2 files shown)


superKalo:
This is not needed actually. That's because we don't have issues enabled in this repo and for the project, we use issues associated to another repo anyways.

So if we want to achieve this in -common, we should figure out another way for doing it, probably with fine-grained token that has read access to our main issues board.

PR #2389, opened 2026-05-15
Fix / do not auto update to token chain on from token change

gergana95:
Fixes:
- DO not update To token chain when from token is changed.
- Update the toTokenShortList on every from and to token change.
Closes https://github.com/AmbireTech/ambire-app/issues/7138


PetromirDev (changes requested):
Not sure if this is from this change, but `toTokenShortList` breaks after I change the fromToken to a token that is on a different network. How to reproduce:
1. Select a `fromToken` on Ethereum
2. Select a `toToken` on Base
3. Check the to token list (should contain many tokens)
4. Change the `fromToken` to be a token on Polygon
5. Check the to token list (will only contain 1 token)

Image

PetromirDev (approved):

PR #2392, opened 2026-05-19
Feature: Portfolio add low-value token helpers for dashboard collapse

Maskln:
paired with: https://github.com/AmbireTech/ambire-app/pull/7183

### Added
 - `src/libs/portfolio/lowValueTokens.ts` — shared helpers for collapsing low-value tokens on the dashboard:
   - `isLowValueToken` — treats a token as low-value when it has no USD price, or when holding value is below $0.01 (not unit price alone, so cheap tokens with a meaningful balance stay visible).
   - `shouldCollapseLowValueTokens` — enables collapse only when the account has ≥ 100 dashboard tokens.
shouldExcludeFromLowValueCollapse — keeps tokens out of collapse when they have a pending simulation balance change or are rewards tokens. 
   - `getDashboardTokenCount` — counts tokens for the collapse gate, excluding gas-tank and borrowed DeFi positions.
partitionLowValueTokens — splits tokens into visible vs low-value groups and returns the combined USD total for the collapsed group.
 - `src/libs/portfolio/lowValueTokens.test.ts` — unit tests for the above behavior.


Maskln:
resolved in this PR: https://github.com/AmbireTech/ambire-app/pull/7133

PR #2395, opened 2026-05-19
Bug / Potential hang with offchain-resolved names

superKalo:


superKalo:
Wrong clue. Wrote an unit test that profs - even if `reverseLookupEns` hangs, `withTimeout` should unstuck it in 15s.

PR #2406, opened 2026-05-20
Types/ Make `call.to` optional

PetromirDev:
Typescript will prevent issues such as https://github.com/AmbireTech/ambire-common/pull/2403/changes/9b0f405853c4d2c2a37b318cbd186dc63ab24c0d


src/libs/broadcast/broadcast.ts
```diff
@@ -147,15 +147,15 @@ export async function getTxnData(
     if (op.calls.length > 1) {
       const ambireAccount = new Interface(AmbireAccount.abi)
       return {
-        to: account.addr as Hex,
+        to: account.addr as Hex | undefined,
```
Oxbobby: this cannot be undefined

Oxbobby (changes requested):
one small change

PR #2412, opened 2026-05-22
Debug/ Extension performance with huge storage

PetromirDev:
## Note:
I'm away next week. The goal of this PR is to preserve what I've been debugging.

Related to https://github.com/AmbireTech/ambire-app/issues/7085 and https://github.com/AmbireTech/ambire-app/issues/7157

`Colleague` = name of the colleague

We debugged this with Colleague and here is the data we have:
- The issues started in https://github.com/AmbireTech/ambire-app/releases/tag/v6.7.0 (and can be reproduced with all subsequent versions)
- Going back to https://github.com/AmbireTech/ambire-app/releases/tag/v6.6.2 fixes it
- Removing the `#executeBalanceChanges`, introduced in https://github.com/AmbireTech/ambire-common/pull/2332/changes fixes it too
- His storage is ~250MB 

So, what's the reason? `#executeBalanceChanges` is not being awaited, so how is `withStatus('selectAccount'...` slowed down?

The call flow goes like this:
main.selectAccount -> accountsOpsStatusesInterval.restart({ runImmediately: true } -> updateAccountsOpsStatuses -> executeBalanceChanges -> storage.set() -> stringify

Stringify is sync and blocks the main thread for ~5 seconds on my end (and even more on Colleague's machine)

## How to generate a ton of accountOps to test this
Just execute this in the browser console in a new chrome profile (MUST FILL THE ACCOUNTS ARRAY FIRST!!)
[createOps.js](https://github.com/user-attachments/files/28149150/createOps.js)

## How to fix this:
executeBalanceChanges calls `persistAccountsOps` (which calls `storage.set()`) even if `balanceChangesTasks.length === 0`. We should fix this, which will fix `removeAccount` and `selectAccount` 

## To discuss:
Imo, there is a bigger problem here - yes, the extension has unlimited storage access, but if it blocks the main thread like this, then maybe we should consider fixing this. Yes, not many users will have this many account ops, but what if someone is farming daily, has many accounts and uses multiple networks? The extension will get slower with time, and we will never experience this while developing.

## What I still can't comprehend
For some reason, setting `selectedAccount` via the console takes 1.5s on his extension instance, while it takes 1ms on prod on the same machine. On my end, even with 350MB storage, selecting an account takes ~5ms.

This cannot be explained by `stringify`, because it's not called when using the console. Also, if `storage.set` takes this much time just because of the size, why is it so fast on my end, and why is the problem resolved with an older version and when the code is commented out??? :thinking: 

Note: nothing written in this description is based on assumptions. Everything is a fact that I've proven locally or on Colleague's machine

### DIFF
--- a/src/controllers/activity/activity.ts
+++ b/src/controllers/activity/activity.ts
@@ -493,6 +493,8 @@ export class ActivityController extends EventEmitter implements IActivityControl
 
     this.emitUpdate()
 
+    const backfillStart = Date.now()
+
     // find ops with no balance changes recorded and backfill them;
     // no need to console.log anything in the catch statement here
     // as error handling is handled in backfillAccountOpBalanceChangesAndPersist.
@@ -503,7 +505,14 @@ export class ActivityController extends EventEmitter implements IActivityControl
         op.balanceChanges === undefined
     )
     if (opsWithNoBalanceChanges.length)
-      this.backfillAccountOpBalanceChangesAndPersist(opsWithNoBalanceChanges).catch((e) => null)
+      this.backfillAccountOpBalanceChangesAndPersist(opsWithNoBalanceChanges)
+        .then(() => {
+          const backfillEnd = Date.now()
+          console.log(
+            `Debug: finished backfilling balance changes for account ${filters.account} in ${(backfillEnd - backfillStart) / 1000} seconds`
+          )
+        })
+        .catch((e) => null)
   }
 
   setDashboardBannersSeen(sessionId: string, accountAddr: string) {
@@ -559,7 +568,13 @@ export class ActivityController extends EventEmitter implements IActivityControl
   }
 
   private async persistAccountsOps() {
+    const start = Date.now()
     await this.#storage.set('accountsOps', this.#accountsOps)
+    console.log(
+      'Debug: called storage set for accountsOps: persistAccountsOps (took',
+      Date.now() - start,
+      'ms)'
+    )
     await this.syncFilteredAccountsOps()
     this.emitUpdate()
   }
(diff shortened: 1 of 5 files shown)


PetromirDev:
`What I still can't comprehend` was caused by either:
1. Chrome storage awaiting a LOCK from the huge activity write
2. parse/stringify blocking the CPU which then blocks storage writes

and not that Chrome storage is not optimized for such big datasets. We tested with Colleague and storage writes take 1-5ms on the same machine and extension instance. I'm closing this one.

superKalo:
Two observations for the "can't comprehend" part:

1. Why it's 300x slower on his machine than yours (same-ish data size): stringify on a 250MB object briefly holds two copies in memory at once: the original object and the giant text string being built - so RAM usage momentarily doubles to ~500MB+. On a machine with plenty of free RAM (yours), that's just CPU work - fast (5ms). On a RAM-tight machine, it doesn't fit comfortably, so the engine constantly pauses to free memory (garbage collection) and the OS maybe starts pushing memory to disk (swap) - and disk is thousands of times slower than RAM. Same task, but now it's "compute + constant cleanup + disk thrashing" - 1.5s. So the bottleneck isn't the data size itself, it's how much free RAM the specific machine has.

2. Op count hurts more than MB: `richJson` doesn't copy the blob in one go - it walks every single value and runs a check on it ("is this a bigint?"). So cost scales with the number of values, not the number of bytes. 250MB made of many tiny account ops (millions of fields) is far slower to stringify than 350MB made of fewer large objects - so a smaller-but-more-fragmented dataset can actually be slower.

Takeaway: the slowness is RAM-pressure + op-count, not raw size - exactly the class of problem that on theory goes away with per-record IndexedDB writes (no full-blob richJson on every tick) discussed in https://github.com/AmbireTech/ambire-app/issues/7264

PR #2414, opened 2026-05-26
Add: a 20% gas ovearhead when doing a MAX transfer to make sure RPC's don't reject the max transfer 

Oxbobby:
Closes: https://github.com/AmbireTech/ambire-app/issues/7223

### DIFF
--- a/src/libs/transfer/amount.ts
+++ b/src/libs/transfer/amount.ts
@@ -1,5 +1,16 @@
+const FEE_RESERVE_OVERHEAD_BPS = 2000n
+const BPS = 10000n
+
+const getFeeWithReserveOverhead = (fee: bigint): bigint => {
+  if (fee === 0n) return 0n
+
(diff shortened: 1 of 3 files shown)


src/libs/transfer/amount.ts
```diff
@@ -1,5 +1,16 @@
+const FEE_RESERVE_OVERHEAD_BPS = 2000n
+const BPS = 10000n
+
+const getFeeWithReserveOverhead = (fee: bigint): bigint => {
+  if (fee === 0n) return 0n
+
+  return (fee * (BPS + FEE_RESERVE_OVERHEAD_BPS) + BPS - 1n) / BPS
```
jordan-enev, line 8: I would only add a comment that here we do it for ceiling division `+ BPS - 1n`, i.e. to round up 3.6 to 4

jordan-enev (approved):

Oxbobby:
@superKalo ready for hotfix tomorrow

PR #2415, opened 2026-05-26
Fix: duplication (by filtering blacklisted tokens, e.g. EURe) from activity balance changes going forward

Oxbobby:
Problem: Eure was being shown twice in the activity cards as it's filtered from the portfolio (its blacklisted address), but not in activity. So we filter it in activity as well

Fixes: https://github.com/AmbireTech/ambire-app/issues/7222

### DIFF
--- a/src/controllers/activity/activity.ts
+++ b/src/controllers/activity/activity.ts
@@ -182,11 +182,14 @@ const getBalanceChangeTokenAddrsFromReceipts = async (
   accountOp: SubmittedAccountOp,
   receipts: TransactionReceipt[]
 ) => {
-  const foundTokens = (
-    await Promise.all(
-      receipts.map((receipt) => getTransferLogTokens(receipt.logs, accountOp.accountAddr))
-    )
-  ).flat()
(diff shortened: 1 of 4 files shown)


superKalo (approved):
Since we blacklist those on the Dashboard, the changes makes sense to me.

I'm only concerned if we have a valid use-case where we want to show in the ~~simulation~~ activity of the scam tokens (going in our out). Let's discuss that offline.

superKalo:
In https://github.com/AmbireTech/ambire-common/pull/2415/changes/dc47397872202795436e447eb83d21590504c9c0 I applied blacklist filter to all three `getTransferLogTokens` call sites in the activity controller so blacklisted tokens (e.g. EURe) are excluded from balance changes consistently.

superKalo:
Also, we discussed that for the passed activities - to keep things simple - we won't do a migration that cleans them up. Problem is not a big deal that would require another migration. So this deduplication logic would only click from new activities.

gergana95 (approved):

PR #2417, opened 2026-05-27
Fix/ allow the humanizer to decode txns with non-strict encoding

JIOjosBG:
The goal is to not allow humanizer bypassing like [the reported here](https://ambiretech.slack.com/archives/C0644227TAP/p1779545278204989)

## How to exploit 
Make a normal approve tx but replace a zero from the padding of the address argument with any other hex character. That would cause the ethers decoding to fail, resulting in no humanization for the approve

## The fix
replace the ethers decoding with viem decoding that allows humanization of the data

## How to test
Click sign on [this tx](https://transact.swiss-knife.xyz/send-tx?chainId=10&calldata=0x095ea7b30000a00000000000000000000b2c639c533813f4aa9d7837caf62653d097ff850000000000000000000000000000000000000000000000000000000000000001&to=0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85) which is has a zero padding byte changed in the beginning 
If there is humanization, it is fixed

### DIFF
--- a/src/libs/humanizer/index.test.ts
+++ b/src/libs/humanizer/index.test.ts
@@ -393,3 +393,108 @@ describe('with (Account | Key)[] arg', () => {
     compareHumanizerVisualizations(irCalls, expectedVisualizations)
   })
 })
+
+// Non-strict / dirty-bytes ABI encoding: the 12 leading zero bytes that pad a 20-byte
+// address to a 32-byte ABI slot (or the 31 zero bytes that pad a bool) are replaced with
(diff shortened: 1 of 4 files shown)


JIOjosBG:
re: the AI review
1. viem returning lowercase addresses. This is not true when decoding a transaction (tested)
image


It is very interesting that free cahtgpt and deepseek say the same, while grok and claude say it will be checksummed
2. missing per-call error handling - this has always been the case, i will fix in separate PR
3. leftover console log - good catch

PR #2419, opened 2026-05-27
Add safeguards for extreme gas fees and swap loss warnings

Maskln:
paired with: https://github.com/AmbireTech/ambire-app/pull/7231

## Summary
- Add safeguard thresholds and helpers for extreme gas fees (mainnet 20 GWEI, other chains $10) and swap losses above $10k.
- Extend `calculateAmountWarnings` with `estimatedLossUsd`, `severity` (`elevated` | `extreme`), and warning types in `consts/safeguards/swapAmountWarnings.ts`.
- Add `getExtremeGasFeeWarningState` for sign/estimation flows and unit tests under `libs/safeguards/`.


src/consts/safeguards/extremeGasFee.ts
```diff
@@ -0,0 +1,15 @@
+export const EXTREME_GAS_FEE_THRESHOLD_MAINNET_USD = 100
+export const EXTREME_GAS_FEE_THRESHOLD_DEFAULT_USD = 10
```
superKalo, line 1: @Oxbobby wydt about these, sound legit?

src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -626,16 +626,10 @@ export const calculateAmountWarnings = (
     // Can be negative if the output is higher
     // (possible during arbitrage swaps)
     const difference = inputValueInUsd - outputValueInUsd
+    if (difference <= 0) return null
```
superKalo: I think one of the tests caught a regression with this.

```
FAIL src/libs/swapAndBridge/swapAndBridge.test.ts
  ● swapAndBridge lib › calculateAmountWarnings › edge case: should return slippage impact warning even if outputValueInUsd > inputValueInUsd when minAmountOut is very low
```

Please check the arbitrage case (input $100, output $110, minAmountOut $20) - should still warn about dangerous slippage, right?

The new `if (difference <= 0) return null` guard short-circuits before the slippage check runs.

src/consts/safeguards/extremeGasFee.ts
```diff
@@ -0,0 +1,15 @@
...
+  return chainId === 1n
+    ? EXTREME_GAS_FEE_THRESHOLD_MAINNET_USD
+    : EXTREME_GAS_FEE_THRESHOLD_DEFAULT_USD
+}
+
+export function isExtremeGasFee(feeUsd: number, chainId: bigint): boolean {
+  if (!Number.isFinite(feeUsd) || feeUsd <= 0) return false
+
+  return feeUsd > getExtremeGasFeeThresholdUsd(chainId)
+}
```
superKalo: Minor note, both `consts/safeguards/extremeGasFee.ts` and `consts/safeguards/extremeSwapLoss.ts` are mostly functions, not constants. Consider moving the helper functions to `libs/safeguards/` and
keeping only the threshold values in `consts/`, similar to how `swapAmountWarnings.ts` in consts
is only types.

src/libs/safeguards/extremeGasFee.ts
```diff
@@ -0,0 +1,38 @@
...
+  )
+  const selectedFeeSpeed =
+    signAccountOpState.feeSpeeds[identifier]?.find(
+      (speed) => speed.type === signAccountOpState.selectedFeeSpeed
+    ) || signAccountOpState.feeSpeeds[identifier]?.find((speed) => speed.type === FeeSpeed.Fast)
+
+  if (!selectedFeeSpeed?.amountUsd) return null
+
+  const feeUsd = Number(selectedFeeSpeed.amountUsd)
+  if (!isExtremeGasFee(feeUsd, networkChainId)) return null
```
superKalo: The current check (`feeUsd > threshold`) has no context about what's being signed. A batched AccountOp with 5 expensive DeFi calls costing $120 in gas is reasonable; a single ERC-20 transfer costing $120 is not.

Calling @Oxbobby for an opinion here. I think we should consider adding a gas-to-value ratio check alongside the absolute threshold (e.g. warn only when fee exceeds X% of the total transaction value), or at minimum divide `feeUsd` by the number of user ops in the batch to get a per-operation cost before comparing to the threshold.

src/consts/safeguards/extremeGasFee.ts
Oxbobby: yes, they look fair

src/consts/safeguards/extremeSwapLoss.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function getExtremeSwapConfirmationPhrase(estimatedLossUsd: number): string {
+  const lossRounded = Math.floor(estimatedLossUsd)
+
+  return `I understand I will lose over ${lossRounded} dollars on this trade`
+}
+
+export function normalizeConfirmationPhraseInput(phrase: string): string {
+  return phrase.trim().replace(/\s+/g, ' ')
+}
```
Oxbobby: Let's remove the regex. A trim() is enough

superKalo (approved):
Great, what we discussed:

1) Please before merging make sure the new tests reflect tha latest conditions and do pass.

2) Please check if we have a test that covers the swap arbitrage case (where no warnings should be displayed, because the trade is actually net positive)

3) Minor: consider trimming the "continue" phrase (I've seen many times users typing with blank space at the end), this shouldn't be an issue I think.

PR #2421, opened 2026-05-28
When we select a token from the dashboard, change the network in swap&bridge to that network

Oxbobby:
Problem:  
I selected a token in the dashboard -> clicked -> swap & bridge for that token -> went to the swap & bridge screen, but the network of the token I selected differs from the chosen on in swap & bridge

Fix:  
Make it the same


superKalo (approved):
Makes sense 👍 

Thanks for adding a test for this 👏

PR #2427, opened 2026-05-29
Disallow wrong 712 domain chain

JIOjosBG:
Resolves: https://github.com/ambireTech/ambire-app/issues/7254
## Copies raby's and MM behavior:
- when chainId is missing, assume it should be the same as the current chain
- when chainId is 0 assume it is the same as the current hcain
- else when it is not correct throw error for the dapp to handle
- when correct do everything as expected 

## To test
try this with all 3 wallets
send712Request()
send712Request(0)
send712Request(RANDOM_CHAIN_ID)
send712Request(CURRENT_CHAIN_ID)

this is the function you can copy paste in the console:
```js
const send712Request = async (chainId) => {
  if (!window.ethereum) {
    throw new Error("No wallet found");
  }

  const [account] = await ethereum.request({
    method: "eth_requestAccounts",
  });

  const domain = {
    name: "MyDapp",
    version: "1",
    chainId,
    verifyingContract: "0x0000000000000000000000000000000000000000",
  };

  const types = {
    EIP712Domain: [
      { name: "name", type: "string" },
      { name: "version", type: "string" },
      { name: "chainId", type: "uint256" },
      { name: "verifyingContract", type: "address" },
    ],
    Message: [
      { name: "contents", type: "string" },
      { name: "timestamp", type: "uint256" },
    ],
  };

  const message = {
    contents: "Hello from EIP-712",
    timestamp: Math.floor(Date.now() / 1000),
  };

  const data = JSON.stringify({
    domain,
    types,
    primaryType: "Message",
    message,
  });

  const signature = await ethereum.request({
    method: "eth_signTypedData_v4",
    params: [account, data],
  });

  console.log("Address:", account);
  console.log("Signature:", signature);
}
```

### DIFF
--- a/src/controllers/requests/requests.test.ts
+++ b/src/controllers/requests/requests.test.ts
@@ -738,5 +738,45 @@ describe('RequestsController ', () => {
       expect(controller.userRequests.length).toBe(1)
       expect(controller.userRequests[0]!.kind).toBe('typedMessage')
     })
+
+    test('rejects when domain.chainId does not match the current network chainId', async () => {
+      const { controller } = await prepareTest(true)
+      const typedData = {
+        ...VALID_TYPED_DATA,
+        domain: { ...VALID_TYPED_DATA.domain, chainId: 999 }
+      }
+      await expect(buildSignTypedDataRequest(controller, typedData)).rejects.toThrow(
+        'The domain chainId (999) does not match the current network chainId (1)'
+      )
+    })
+
+    test('replaces domain.chainId with current network chainId when domain.chainId is 0', async () => {
+      const { controller } = await prepareTest(true)
+      const typedData = {
+        ...VALID_TYPED_DATA,
+        domain: { ...VALID_TYPED_DATA.domain, chainId: 0 }
+      }
+      await expect(buildSignTypedDataRequest(controller, typedData)).resolves.toBeUndefined()
+      expect(controller.userRequests.length).toBe(1)
+      const req = controller.userRequests[0]! as any
+      expect(req.meta.params.domain.chainId).toBe(1n)
+    })
+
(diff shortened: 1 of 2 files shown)


Oxbobby (changes requested):
This is dangerous and I'm leaving a comment to allow time to think before we merge.  

A lot of dapps are requesting messages on Ethereum even though the connected network is another. In many cases, this is for a plain text signature. But there could be cases for EIP-712 messages. I propose merging this only if we're 100% sure it won't cause problems along the way

JIOjosBG:
The behavior i implemented is taken from how Rabby and MM appear to behave.

Now i read their code and looks the same

- if not chainId - allow signing
- if chainId = 0 or 0n - allow signing
- if chainId present and not the same - disallow 

Rabby 
```
    if (jsonData.domain.chainId) {
      const chainItem = findChainByEnum(currentChain);
      if (
        !currentChain ||
        (chainItem && Number(jsonData.domain.chainId) !== chainItem.id)
      ) {
        throw ethErrors.rpc.invalidParams(
          'chainId should be same as current chainId'
        );
      }
    }
```

MM
```
    if (chainId) {
        if (typeof chainId === 'string') {
            chainId = parseInt(chainId, chainId.startsWith('0x') ? 16 : 10);
        }
        const activeChainId = parseInt(currentChainId, 16);
        if (Number.isNaN(activeChainId)) {
            throw new Error(`Cannot sign messages for chainId "${chainId}", because MetaMask is switching networks.`);
        }
        if (chainId !== activeChainId) {
            throw new Error(`Provided chainId "${chainId}" must match the active chainId "${activeChainId}"`);
        }
    }

```


Lets merge it now so it has a week in v2 and then a week in next
@Oxbobby wdyt

Oxbobby (approved):

PR #2428, opened 2026-05-30
Improve: activity refresh interval to make chains feel faster

Oxbobby:
Improve the activity refresh interval and allow that config to be set from the relayer to make fast chains really feel faster


Oxbobby:
AI comment is irrelevant, that's exactly how it's supposed to work

sonytooo (approved):

PR #2433, opened 2026-06-01
Fixes / Portfolio state clone

sonytooo:
* Remove redundant structuredClone of the portfolioAccountState in updateSelectedAccountPortfolio because it was causing global error on the mobile app when the portfolio state was containing Error instances (criticalError/errors on a NetworkState). WebView's JS engine on mobile cannot structuredClone Error objects and throws "Unable to deserialize data.", which surfaced as an opaque "Script error.". The other reason for the removal is that we do not mutate the portfolio state anymore in this function so no need to keep it there just to slow down the updates

### DIFF
--- a/src/controllers/selectedAccount/selectedAccount.ts
+++ b/src/controllers/selectedAccount/selectedAccount.ts
@@ -226,14 +226,7 @@ export class SelectedAccountController extends EventEmitter implements ISelected
   updateSelectedAccountPortfolio(skipUpdate?: boolean) {
     if (!this.#portfolio || !this.account) return
 
-    // Use cloneDeep instead of structuredClone because the portfolio state can
-    // contain Error instances (criticalError/errors on a NetworkState). The
-    // WebView's JS engine on mobile cannot structuredClone Error objects and
-    // throws "Unable to deserialize data.", which surfaced as an opaque
-    // "Script error." in the Global handler. cloneDeep handles Error and BigInt.
-    const portfolioAccountState = cloneDeep(
-      this.#portfolio.getAccountPortfolioState(this.account.addr)
-    ) as AccountState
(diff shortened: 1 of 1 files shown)


PetromirDev (approved):

src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -21,6 +23,7 @@ import {
   getDefiPositionsOnDisabledNetworksForTheSelectedAccount
 } from '../../libs/banners/banners'
 import { AssetType } from '../../libs/defiPositions/types'
+import { cloneDeep } from '../../libs/richJson/richJson'
```
superKalo: Import not used?

src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -1,5 +1,7 @@
 import { formatEther, getAddress, isAddress } from 'ethers'
 
+import { AccountState } from '@/libs/portfolio/interfaces'
```
superKalo: This is also import that's not being used.

superKalo (approved):

src/controllers/selectedAccount/selectedAccount.ts
superKalo: Fixed in https://github.com/AmbireTech/ambire-common/pull/2433/changes/c9b54a9c1389c51e93d1c6a3173715a28f64f389

src/controllers/selectedAccount/selectedAccount.ts
superKalo: Fixed in https://github.com/AmbireTech/ambire-common/pull/2433/changes/c9b54a9c1389c51e93d1c6a3173715a28f64f389

PR #2437, opened 2026-06-02
Performance optimizations

sonytooo:
**1. richJson parse fast-path (ambire-common)**
* parse() now skips the per-node reviver when neither $bigint nor a serialized Error marker is present, taking V8's native JSON.parse fast path. Previously a reviver was forced on every node, freezing the worker thread (on mobile especially) on large all-string blobs (e.g. the multi-MB phishing list). Covered by new tests for both the fast path and the bigint/error revive path.
* Improves: ~10× faster parse of large all-string payloads.


src/libs/richJson/richJson.test.ts
```diff
@@ -53,6 +53,24 @@ describe('bigintJson', () => {
...
 
+  it('parses all-string data without the reviver (fast path) identically', async () => {
+    const phishingLike = {
+      version: 12,
+      updatedAt: 1700000000000,
+      domains: ['evil.example', 'scam.example', 'Error-lookalike.example'],
+      addresses: ['0xdeadbeef', '0x000000000000000000000000000000000000dEaD']
+    }
+    expect(parse(stringify(phishingLike))).toEqual(phishingLike)
+  })
```
superKalo, line 64: This test does not exercise the fast path it's named after 🤔 'Error-lookalike.example' contains the "Error marker, so needsReviver is true and this input goes through the reviver path. Is this expected?

Better test would spy on JSON.parse and assert it's called with a single argument for marker-free input, and with a reviver when markers are present.

superKalo (approved):
Solid optimization 👏

PR #2440, opened 2026-06-03
Feature/ Scheduled portfolio updates

PetromirDev:
## Problem
If you do the following:

- Open AAVE to Supply
- AAVE requests an approval. Once the txn is confirmed, the nonce is incremented and we do a force update.
- The approval is confirmed so AAVE sends the supply txn
- The supply txn is confirmed, we do a force update, but the server returns the cached response, because the server has an internal limit to how often you can do force updates

## Solution
Instead of calling velcro immediately after a transaction is confirmed we schedule an update for after 1 minute. A portfolio interval runs every 20seconds and checks for scheduled updates. If there is one, it runs the update. 

## End result
Defi positions are automatically discovered after 1 minute of not doing transactions on a network.


src/controllers/portfolio/portfolio.ts
```diff
@@ -778,6 +858,10 @@ export class PortfolioController extends EventEmitter implements IPortfolioContr
         return
       }
 
+      // The update below doesn't bypass the server cache; schedule a cache-busting
+      // update for after the server has had time to index the confirmed transaction.
+      // See ScheduledUpdates for the full rationale.
+      this.#scheduleUpdate({ accountId: accountAddr, chainId, bypassServerSideCache: true })
```
Oxbobby: Why is `this.#scheduleUpdate` in discardSimulation? Because by presumption, this is always the last item we call after the portfolio finishes?

src/libs/defiPositions/defiPositions.ts
```diff
@@ -528,9 +528,14 @@ const getShouldBypassServerSideCache = (
...
   sessionIds: string[],
-  hasNonceChangedSinceLastUpdate: boolean
+  hasNonceChangedSinceLastUpdate: boolean,
+  hasScheduledUpdate: boolean
 ): boolean => {
-  // Always bypass cache if the nonce has changed
+  // A scheduled update will bypass the cache soon; don't burn the server-side bypass
+  // budget for automatic updates (the server has a per-account cooldown).
+  // Manual updates are exempt: the user explicitly requested fresh data.
+  if (hasScheduledUpdate && !isManualUpdate) return false
```
Oxbobby, line 537: There's a case to be made to remove `!isManualUpdate` from the check here. The user himself may not be able to judge correctly whether he's not being forceful. Even debank needs some time to index the new DeFi position.

If we have a hasScheduledUpdate and we know it's going to be up after a minute, maybe it's better to force the wait. Otherwise, the user refreshes too soon, the defi position is not there, the cache is set and he's done for a while

Oxbobby (changes requested):

src/controllers/portfolio/portfolio.ts
PetromirDev: Because discardSimulation is only called when a transaction made by the selected account is confirmed:
image

src/libs/defiPositions/defiPositions.ts
PetromirDev: I see your point. What I'm worried about is the case where a user makes like 5-6 txns in a row and the defi txn is one of the first. In that case the time to discover the position will be much bigger than usual and the user may reload manually.

PetromirDev: Tho I'm really 50/50 about this

src/controllers/portfolio/portfolio.ts
PetromirDev: The alternative is to make it a public method of the portfolio and call it explicitly before the removal of the simulation. Wdyt?

src/libs/defiPositions/defiPositions.ts
Oxbobby: discussed in slack: we're leaving it as it is

PR #2447, opened 2026-06-04
Fixes/clear signing cases

Oxbobby:
## Summary

- Add ERC-7730 humanization support for Safe reject transactions, including SafeTx typed messages and Safe `execTransaction` calls.
- Improve Safe transaction decoding by falling back to ABI head parsing when normal decoding cannot decode nested/multisend calldata.
- Detect Permit2 zero-amount approvals as revoke approvals and show `Revoke approval` wording.
- Improve ERC-7730 value handling for uint-encoded addresses, bigint address formatting, byte slices, and hex enum values.
- Normalize 1inch `fillOrder` humanization into clear swap rows: amount to send, minimum to receive, plus any non-token metadata rows.
- Harden Safe module `execTransaction` decoding and delegate-call warning handling.

## Testing

- Added/updated humanizer tests for Safe reject humanization, Permit2 revoke approvals, 1inch orders, uint byte slices, and Safe ERC-7730 fallback handling

### DIFF
--- a/src/libs/humanizer/modules/Safe/index.ts
+++ b/src/libs/humanizer/modules/Safe/index.ts
@@ -207,10 +207,34 @@ const SafeModule: HumanizerCallModule = (accOp: AccountOp, calls: IrCall[]): IrC
     [toFunctionSelector(execTransactionAbi[0])]: (call: HexIrCall): IrCall | undefined => {
       if (!call.to) return
       if (call.value) return
-      const { args } = decodeFunctionData({ abi: execTransactionAbi, data: call.data })
-      const [to, value, data, operation] = args
+      let args: unknown[]
 
-      const safeSpecificHumanization = getSafeHumanization(accOp.accountAddr, to, value, data)
+      try {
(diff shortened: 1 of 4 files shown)


src/libs/humanizer/modules/Safe/index.ts
```diff
@@ -207,10 +207,34 @@ const SafeModule: HumanizerCallModule = (accOp: AccountOp, calls: IrCall[]): IrC
...
+      if (typeof value !== 'bigint' && typeof value !== 'number' && typeof value !== 'string')
+        return
+      if (
+        typeof operation !== 'bigint' &&
+        typeof operation !== 'number' &&
+        typeof operation !== 'string'
+      )
+        return
+      const bigintValue = BigInt(value)
+      const bigintOperation = BigInt(operation)
```
JIOjosBG, line 230: you do not need to make all those checks, if viem succeeds it will return correct types.

you can:
```
try{
const args = decodeFunctionData({ abi: execTransactionAbi, data: call.data }).args
const [to, value, data, operation] = args
LOGIC HERE
}catch(e){
...
}
```

PR #2448, opened 2026-06-04
Refactor/ AI PR reviews and fix security reviews

PetromirDev:
## Changes: 

**Security reviews work now.** `/security-review` is not a command PR-Agent recognizes, so the previous workflow (which rewrote the event payload to trick the agent) never reliably triggered the review logic. The workflows now invoke
the PR-Agent CLI directly with an explicit `--pr_url ... review`, so trigger phrases (`/review`, `/security-review`, labels, manual dispatch) are handled purely by GitHub Actions conditions and PR-Agent never parses the comment. The shared logic lives in a reusable workflow, so both review flavors (and other repos later) use the same code with different models/instructions.

**Issue/ticket fetching disabled.** The repo has no GitHub issues, so PR-Agent's issue lookups always 404'd (and triggered a logging crash inside PR-Agent). Disabled via `require_ticket_analysis_review = false` and `extract_issue_from_branch = false`.

**Each review posts a new comment.** Previously reviews updated the existing comment (`persistent_comment = true`), which made the history hard to follow on PRs with many re-reviews. Now every run adds a fresh comment.

**Fixed the token cap that silently broke reviews.** PR-Agent defaults to `max_model_tokens = 32000`, which clipped diffs well below Kimi's actual context window; combined with reasoning-token exhaustion this caused runs that produced no review at all. The cap is now aligned with `custom_model_max_tokens` (256k). Also fixed `fallback_models` being set under the wrong config section, where it was silently ignored.

Supply-chain hardening included: PR-Agent runs from a Docker image pinned by immutable digest, actions are pinned by commit SHA, and the workflow token is scoped to `pull-requests: write` + `contents: read` (no push access).

## Known limitation

PR-Agent does not reliably propagate failures via exit codes (upstream issue The-PR-Agent/pr-agent#1053), so a failed run can still end "green" without posting a review. I haven't found a non-hacky fix for this yet - for now we'll monitor for missing review comments manually, and revisit if upstream improves error propagation or a clean solution turns up.

### DIFF
--- a/.github/workflows/general-review.yml
+++ b/.github/workflows/general-review.yml
@@ -20,9 +20,8 @@ jobs:
       pull-requests: write
       # This MUST be read!
       contents: read
-    uses: ./.github/workflows/pr-agent-review.yml
+    uses: ./.github/workflows/pr-agent.yml
     with:
-      review_name: General review
       model: openrouter/moonshotai/kimi-k2.6
       fallback_models: '["openrouter/anthropic/claude-sonnet-4.6"]'
       extra_instructions: |
--- a/.github/workflows/general-review.yml
+++ b/.github/workflows/general-review.yml
@@ -12,25 +12,36 @@ jobs:
     if: >-
       (github.event.action == 'labeled' && github.event.label.name == 'review') ||
       github.event_name == 'workflow_dispatch' ||
-      (github.event.sender.type != 'Bot' && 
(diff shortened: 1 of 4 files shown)


PetromirDev:
Dang, PR agent's review is dogshit compared to Copilot. Gotta work on that

superKalo (approved):

PR #2449, opened 2026-06-05
Configure swap slippage settings as they were far too strict

Oxbobby:
## Summary

Updates Swap & Bridge slippage defaults to be route-aware instead of shrinking tolerance based on trade size.

Previously, `getSlippage()` reduced slippage as the USD amount increased, which made large trades increasingly hard to execute and could even produce values below provider minimums. The new logic chooses defaults from pre-quote route intent:

- wrap / unwrap: `0.01%`
- same-chain stable pair: `0.2%`
- same-chain volatile pair: `0.5%`
- cross-chain stable pair: `0.5%`
- cross-chain volatile pair: `1%`

The helper also now handles provider-specific units:
- LI.FI receives decimal slippage, e.g. `0.005`
- Socket/Bungee and Squid receive percentage slippage, e.g. `0.5`

## Changes

- Refactored `getSlippage()` to accept route context instead of amount-based boundaries.
- Added provider-specific slippage formatting for LI.FI, Socket/Bungee, and Squid.
- Added Socket/Bungee quote slippage support.
- Added focused tests for route-aware defaults and provider unit conversion.

## Testing

- `npm run jest -- swapAndBridge.test.ts`
- `npx prettier --check src/libs/swapAndBridge/swapAndBridge.ts src/services/lifi/api.ts src/services/squid/api.ts src/services/socket/api.ts src/libs/swapAndBridge/swapAndBridge.test.ts`
- `yarn extension:type:check-new`

### DIFF
--- a/src/libs/swapAndBridge/swapAndBridge.ts
+++ b/src/libs/swapAndBridge/swapAndBridge.ts
@@ -14,7 +14,6 @@ import {
   BRIDGE_STATUS_INTERVAL,
   UPDATE_SWAP_AND_BRIDGE_QUOTE_INTERVAL
 } from '../../consts/intervals'
-import { getTokenUsdAmount } from '../../controllers/signAccountOp/helper'
 import { Account, AccountOnchainState } from '../../interfaces/account'
 import { Fetch } from '../../interfaces/fetch'
 import { Network } from '../../interfaces/network'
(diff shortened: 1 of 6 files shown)


src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -572,21 +571,63 @@ const isNoFeeToken = (chainId: number, tokenAddr: string) => {
...
+  'USDT',
+  'USDT.E',
+  'USDS',
+  'XDAI'
+])
+
+const SQUID_MIN_SLIPPAGE_PERCENT = 0.01
+
+const getIsStableToken = (token: { symbol: string }) =>
+  STABLE_TOKEN_SYMBOLS.has(token.symbol.trim().toUpperCase())
```
PetromirDev: Make it an array instead of a set and use .includes to handle .E, (Pos),... vairants. Also, shouldn't we handle `USD₮`?

src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -572,21 +571,63 @@ const isNoFeeToken = (chainId: number, tokenAddr: string) => {
...
+
+const formatProviderSlippage = (provider: SlippageProvider, slippagePercent: number) => {
+  const providerSlippage =
+    provider === 'lifi'
+      ? slippagePercent / 100
+      : Math.max(slippagePercent, SQUID_MIN_SLIPPAGE_PERCENT)
+
+  return providerSlippage.toString()
+}
```
PetromirDev, line 620: wb socket?

src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -572,21 +571,63 @@ const isNoFeeToken = (chainId: number, tokenAddr: string) => {
...
+  toAsset: SwapAndBridgeToToken
+  fromChainId: number
+  toChainId: number
+  provider: SlippageProvider
+  isWrapOrUnwrap: boolean
+}) => {
+  const isCrossChain = fromChainId !== toChainId
+  const isStablePair = getIsStableToken(fromAsset) && getIsStableToken(toAsset)
+
+  if (isWrapOrUnwrap) return formatProviderSlippage(provider, SQUID_MIN_SLIPPAGE_PERCENT)
```
PetromirDev, line 640: Why pass SQUID_MIN_SLIPPAGE_PERCENT here?

PetromirDev (changes requested):

PetromirDev:
Rerunning to test https://github.com/AmbireTech/ambire-common/pull/2451

/review

src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -572,21 +571,63 @@ const isNoFeeToken = (chainId: number, tokenAddr: string) => {
...
+  'TUSD',
+  'USDC',
+  'USDC.E',
+  'USDE',
+  'USDP',
+  'USDT',
+  'USDT.E',
+  'USDS',
+  'XDAI'
+])
```
superKalo: Other popular USD-stables missing: BUSD, GUSD, RLUSD, USD1.

Wydt to we expand the list with EUR-stables and GBP-stables?

superKalo: Also, how about the yield-bearing stables like sUSDS, sDAI, sUSDe, sfrxUSD, USDY, sGHO 👀 ?

superKalo (approved):

src/libs/swapAndBridge/swapAndBridge.ts
Oxbobby: Expanding the list with:

```
  'XDAI',
  'BUSD',
  'GUSD',
  'RLUSD',
  'USD1',
  'sUSDS',
  'sDAI',
  'sUSDe',
  'sfrxUSD',
  'USDY',
  'sGHO',
  'USD₮',
  'EURe',
  'EURc',
  'GBPe',
  ```

Let's think if we need something else. Also, this low slippage makes sense for usd <-> usd swap. I don't think it makes sense for usd -> eur swap. They are stables, but liquidity should not be deep enough for you to be able to execute such a trade so easily with such a low slippage.  
Adding them for now and let's think about this

File: src/libs/swapAndBridge/swapAndBridge.ts (hunk 1 above)
Comment (Oxbobby) [reply src/libs/swapAndBridge/swapAndBridge.ts]: added USD₮ and others to the list. I fixed the rest

File: src/libs/swapAndBridge/swapAndBridge.ts (hunk 3 above)
Comment (Oxbobby) [reply src/libs/swapAndBridge/swapAndBridge.ts:640]: The if condition was wrong - it should !== 'squid'. Other than that, Squid has a minimum of 0.01, or let's say a recommended minimum. That's why we use it

File: src/libs/swapAndBridge/swapAndBridge.ts (hunk 2 above)
Comment (Oxbobby) [reply src/libs/swapAndBridge/swapAndBridge.ts:620]: fixed

**Review by superKalo — APPROVED** [review]

File: src/libs/swapAndBridge/swapAndBridge.ts [hunk 5]
```diff
@@ -572,21 +573,75 @@ const isNoFeeToken = (chainId: number, tokenAddr: string) => {
...
+}: {
+  fromAsset: TokenResult
+  toAsset: SwapAndBridgeToToken
+  fromChainId: number
+  toChainId: number
+  provider: SwapProviderName
+  isWrapOrUnwrap: boolean
+}) => {
+  const isCrossChain = fromChainId !== toChainId
+  const isStablePair = getIsStableToken(fromAsset) && getIsStableToken(toAsset)
```
superKalo, line 638: Note: we discussed offline that if we decide to include non-USD stables here, this should check if both assets are stables against the same currency. Otherwise USDC-EURC would result in false positive here.

superKalo (changes requested):
I'm changing the status until we sync up how to move forward with this.

superKalo:
More context:

- 9 months ago we used hardcoded slippage `Number(fromAmountInUsd) <= 400 ? '0.010' : '0.005'`
- then we improved it in https://github.com/AmbireTech/ambire-common/pull/1737 for large trades - limiting the slippage for trades above $20k, making it in half each time the from amount exceed $20k

To be discussed: if we should try our service providers "auto" settings and handover to them these slippage adjustments.

superKalo:
Update: we decided to try the service provider "auto" as a next step (vs changing our algo) and re-evaluate then.

Oxbobby:
I'm closing this for now as we won't be merging it. Let's reevaluate again if the problem persists

PR #2457, opened 2026-06-08
Security / Suspicious dapps hosting detection

jordan-enev:
Add: Detect and warn on suspicious hosting platforms. Shared hostingplatforms (Google Sites, Vercel, etc. ) are commonly used for phishing because their trusted parent domain bypasses most phishing filters. This adds a SUSPICIOUS_HOSTING status that warns users when a dApp is hosted on such a platform - either directly or via an iframe embedded in a suspicious tab context.

Paired with and more info here https://github.com/AmbireTech/ambire-app/pull/7321.


superKalo (approved):
Well documented and well-thought 👏

src/controllers/dapps/dapps.ts
```diff
@@ -1120,34 +1138,72 @@ export class DappsController extends EventEmitter implements IDappsController {
...
+   *
+   * This detects phishing pages that host a legitimate dApp in an iframe: the legitimate
+   * dApp's own session looks clean, but the phishing page's session (e.g. sites.google.com)
+   * is in the same tab and has SUSPICIOUS_HOSTING status.
+   *
+   * Returns `undefined` if no dangerous co-session is found, or if the status cannot yet
+   * be determined (phishing DB not loaded and domain not in the static list).
+   */
+  #getTabContextStatus(session: Session): BlacklistedStatus | undefined {
+    for (const s of Object.values(this.dappSessions)) {
```
superKalo, line 1153: `#getTabContextStatus` reads every accumulated session for a `(tabId, windowId)`, but I just found out that maybe sessions are rarely cleaned up! Two things feed into this:

- No session cleanup on same-tab navigation - `#createDappSession` only adds, it never evicts prior sessions for the same tab (which is probably by design, but I'm just saying)
- Tab-close cleanup in the extension is currently broken: see `background.ts` and search for filters `sessionKeys.filter((k) => k.startsWith(`${tabId}-`))`, but session IDs are `${windowId}-${tabId}-${dappId}`, so the prefix is the windowId and the filter never matches, right? However, this seems to be pre-existing problem in v2 also - the `windowId` prefix was added to `getSessionId` after that cleanup was written, right? Please check.

src/controllers/dapps/dapps.ts
```diff
@@ -1120,34 +1138,72 @@ export class DappsController extends EventEmitter implements IDappsController {
...
+        s.tabId !== session.tabId ||
+        s.windowId !== session.windowId
+      ) {
+        continue
+      }
+      const status = this.#phishing.getDomainBlacklistedStatus(s.origin)
+      // Whether the co-session's domain is BLACKLISTED or SUSPICIOUS_HOSTING, the threat
+      // is the same for this session: dangerous hosting context. Always return SUSPICIOUS_HOSTING
+      // — BLACKLISTED belongs to the co-session's own domain, not to this dApp.
+      if (status === 'BLACKLISTED' || status === 'SUSPICIOUS_HOSTING') return 'SUSPICIOUS_HOSTING'
```
superKalo, line 1165: The rationale in the comment is legit (the blacklist belongs to the co-session's own domain, not this dApp), but just to confirm - sharing a tab with a *confirmed* phishing page is arguably more dangerous than a soft warning so in this case will be user see blacklisted warn or the suspicious hosting UI warn? IMO - blacklisted.

jordan-enev: here we will show "SUSPICIUOUS" warning + hold to proceed button

If we want to make it stricter, we can change it easily

src/controllers/dapps/dapps.ts
jordan-enev: from my observation - yes, we need to investigate & fix in a new issue

PR #2460, opened 2026-06-09
Config / Exclude tokens without a USD price from the Swap & Bridge "From" list

superKalo:
Tokens without a known USD price most prob can't be quoted reliably, so they're now filtered out of the Swap & Bridge "From" list.

### DIFF
--- a/src/libs/swapAndBridge/swapAndBridge.ts
+++ b/src/libs/swapAndBridge/swapAndBridge.ts
@@ -44,7 +44,7 @@ import { Call } from '../accountOp/types'
 import { AssetType } from '../defiPositions/types'
 import { PaymasterService } from '../erc7677/types'
 import { TokenResult } from '../portfolio'
-import { getTokenBalanceInUSD } from '../portfolio/helpers'
+import { getTokenBalanceInUSD, getTokenUsdPrice } from '../portfolio/helpers'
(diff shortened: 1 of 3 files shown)


src/libs/swapAndBridge/swapAndBridge.ts
```diff
@@ -247,6 +247,12 @@ export const getIsTokenEligibleForSwapAndBridge = (
     // and are not transferable/swappable by design.
     token.flags.defiTokenType !== AssetType.Borrow
 
+  // Tokens without a known USD price most prob can't be quoted reliably, so exclude them
+  // from the list when the caller opts in (e.g. the Swap & Bridge "form" tokens).
+  if (requirePrice && getTokenUsdPrice(token) <= 0) {
+    return false
+  }
+
```
PetromirDev, line 255: What if cena is down? Maybe we should check for a portfolio price error?

PetromirDev (changes requested):

src/libs/swapAndBridge/swapAndBridge.ts
superKalo: Agreed, I skipped the price req for chains that currently have a price fetch error in https://github.com/AmbireTech/ambire-common/pull/2460/changes/341c40c1d5d7292ff784f5e283c0ddfec2a7e2f0

PR #2468, opened 2026-06-10
Update/ Switch to kimi nitro to improve reliability

PetromirDev:
Done to avoid shitty providers like this one (caps output to 16k tokens)
image

Also should make reviews faster

https://openrouter.ai/docs/guides/routing/model-variants/nitro


PetromirDev:
The issue occurred in https://github.com/AmbireTech/ambire-common/pull/2465#issuecomment-4670632272

PR #2470, opened 2026-06-11
Fix/add native sends to approvals

Oxbobby:
Add detecting native value in approvals + add a warning for it

### DIFF
--- a/src/libs/humanizer/erc7730/humanize.ts
+++ b/src/libs/humanizer/erc7730/humanize.ts
@@ -36,6 +36,7 @@ import {
   getText,
   getToken,
   getWarning,
+  flattenHumanizerVisualizations,
   uintToAddress
 } from '../utils'
 import { SAFE_TX_PRIMARY_TYPE } from './consts'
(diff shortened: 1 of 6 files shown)


src/libs/humanizer/erc7730/humanize.ts
```diff
@@ -1219,6 +1220,38 @@ const dedupeWarnings = (warnings: HumanizerWarning[]): HumanizerWarning[] => {
...
+              value.value !== undefined &&
+              value.value > 0n
+          )
+      )
+  )
+
+  return hasNativeValue
+    ? [
+        getWarning(
+          `This transaction requires ${nativeAssetSymbol}`,
```
JIOjosBG: change the wording. This is incorrect, the transaction does not **require** native. It sends native

`This transaction will send native`

JIOjosBG (changes requested):

JIOjosBG:
The other problem with this approach is that it adds the "send native" row only for approve txns, while the old humanizer added it for all transactions except:
```js
[
            'Swap',
            'Bridge',
            'Swap/Bridge',
            'Supply',
            'Deposit',
            'Supply to vault',
            'Wrap'
          ]
```

I am fine with this going as a hotfix, but we need the logic working for all txns except for the listed actions.

image

Oxbobby:
@JIOjosBG I did the last piece by applying this logic:
If there's something in the erc-7730 humanization that's displaying the ETH, do not add another row. If there's not OR there is but the amounts differ, add it and add the warning to it. This catches all cases where ETH is participating as txn.value

JIOjosBG (approved):

PR #2472, opened 2026-06-11
Feature/ Retry mechanism to `signAccountOp.traceCall`

PetromirDev:
## How it may fail

1. App requests approval
2. Approval pending; app sends a transaction that depends on it
3. Transaction is being estimation; traceCall is ran for the txn
4. Estimation+traceCall fail; Estimation keeps retrying
5. Approval is confirmed; estimation retries, trace call doesn't

### DIFF
--- a/src/controllers/signAccountOp/signAccountOp.ts
+++ b/src/controllers/signAccountOp/signAccountOp.ts
@@ -1,5 +1,3 @@
-/* eslint-disable @typescript-eslint/brace-style */
-
 import {
   AbiCoder,
   formatEther,
@@ -403,7 +401,7 @@ export class SignAccountOpController
 
   signAndBroadcastPromise: Promise<void> | undefined
 
-  #traceCallTimeoutId: ReturnType<typeof setTimeout> | null = null
+  private traceCallTimeoutId: ReturnType<typeof setTimeout> | null = null
 
   #gasPriceInterval: IRecurringTimeout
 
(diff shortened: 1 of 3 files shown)


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -774,8 +772,31 @@ export class SignAccountOpController
...
+      if (
+        lastEstimationStatus === EstimationStatus.Error &&
+        this.estimation.status === EstimationStatus.Success &&
+        // Retry unless discovery already completed (Done) - not only when it Failed.
+        // Any discovery that was still pending while the estimation was failing is
+        // unreliable too: it may have queried the RPC before the underlying issue
+        // cleared, so its result can't be trusted. Only a Done discovery is safe to keep.
+        this.traceCallDiscoveryStatus !== TraceCallDiscoveryStatus.Done
+      ) {
+        this.traceCallDiscoveryStatus = TraceCallDiscoveryStatus.NotStarted
```
Oxbobby: Let's use setDiscoveryStatus() here although it's doing the same thing

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -774,8 +772,31 @@ export class SignAccountOpController
...
+      // An example is doing an approval and a transaction immediately after - the second
+      // transaction will fail before the approval txn is confirmed.
+      if (
+        lastEstimationStatus === EstimationStatus.Error &&
+        this.estimation.status === EstimationStatus.Success &&
+        // Retry unless discovery already completed (Done) - not only when it Failed.
+        // Any discovery that was still pending while the estimation was failing is
+        // unreliable too: it may have queried the RPC before the underlying issue
+        // cleared, so its result can't be trusted. Only a Done discovery is safe to keep.
+        this.traceCallDiscoveryStatus !== TraceCallDiscoveryStatus.Done
```
Oxbobby: I'm not sure about this last check. Even if debug_traceCall is reverting onchain, it will return a success and set the discovery status to Done. The estimation will result in an error, but debug_traceCall won't. That's because debugging transaction is a valid thing to do even if the transaction is a reverting one. This means this check will apply only when the RPC returns a 500 / 502 server error for debug trace call.

I think this check should be removed. Anytime Error -> Success happens, we can replay traceCall()

Oxbobby:
We have this code:
```
    const timeoutId = setTimeout(() => {
      this.setDiscoveryStatus(TraceCallDiscoveryStatus.SlowPendingResponse)
      this.calculateWarnings()
    }, 2000)
```

In extremely rare cases I think it's possible for this to mess up the logic as it just triggers automatically after 2 seconds upon a slow request. Those cases could be:
* add multiple new calls one after the other
* an error -> success estimation with a new call addition approximately 2s after the first trigger

We can add two protections to it:
```
if (this.traceCallDiscoveryStatus !== TraceCallDiscoveryStatus.InProgress || this.traceCallTimeoutId !== timeoutId) return
```

Let's add a comment why we're doing this as well: prevent multiple traceCall() race conditions

PetromirDev:
Implemented in https://github.com/AmbireTech/ambire-common/pull/2475 instead

PR #2476, opened 2026-06-12
Config / Improve asset blacklisting to avoid false positives

superKalo:
Blacklist patterns for token/NFT symbols and names were matched as space-padded substrings (e.g. "mint "), which only checks one boundary and wrongly hid legitimate assets like "papermint protocol".

This reworks the matching:

- `prepareBlacklistPatterns` normalizes the static + dynamic lists once. Space-padded entries are treated as lure words and de-duplicated (" mint", "mint ", " mint " all collapse to "mint"); bare entries ("https", "www.", a full symbol/name) keep the existing substring contract.
- Lure words are matched as whole words and only inside multi-word text, so a single-token symbol/name (WIN, WETH, WIN-AIRDROP) stays visible while "claim your airdrop now" is hidden.
- Matching tokenizes the text once and does a set lookup (no regex).

Confirmed spam from the logs (e.g. "ARBITRUM AIRDROP", "$PENDLE AIRDROP", domain-bearing NFT names) is still hidden. Added unit tests for the new behavior and the papermint-style regression.

### DIFF
--- a/src/libs/portfolio/blacklist.ts
+++ b/src/libs/portfolio/blacklist.ts
@@ -87,45 +87,130 @@ export const containsDomainLike = (text: string): boolean => {
   return false
 }
 
+const isWordChar = (char: string | undefined): boolean =>
+  !!char && ((char >= 'a' && char <= 'z') || (char >= '0' && char <= '9'))
+
+/**
+ * Splits text into its word tokens — maximal runs of [a-z0-9] — so a lure word
+ * can be matched as a WHOLE word rather than a substring. No regex, per repo
+ * rule, mirroring the char scan in `containsDomainLike`.
+ *
+ * "free mint!" → ["free", "mint"], so the lure "mint" matches "free mint" but
+ * never the legitimate "papermint". Text is assumed lowercased by the caller.
+ */
+const toWords = (text: string): string[] => {
+  const words: string[] = []
+  let current = ''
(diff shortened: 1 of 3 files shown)


src/libs/portfolio/blacklist.ts
```diff
@@ -87,45 +87,130 @@ export const containsDomainLike = (text: string): boolean => {
...
+  for (const raw of rawPatterns) {
+    const trimmed = raw.trim()
+    if (!trimmed) continue
+    const pattern = trimmed.toLowerCase()
+    // Surrounding whitespace marks a lure word; a bare pattern (or a padded
+    // multi-word phrase, which has no single whole-word token) stays a substring.
+    const isPaddedLureWord = trimmed !== raw && !pattern.includes(' ')
+    if (isPaddedLureWord) wordPatterns.add(pattern)
+    else substringPatterns.add(pattern)
+  }
```
PetromirDev, line 154: I don't really think that this is needed as we have full control over the API/blacklist so we should just edit it there.

PetromirDev (changes requested):
Tested in with a lot of accounts and all I can say is good job 👏 It works perfectly and I wasn't able to find an account with false positives.

superKalo:
I'll apply the feedback in other PR to not slow this improvements down any further.

PR #2478, opened 2026-06-12
Fix/ Defi apps positions not sorted

PetromirDev:
You can test with `0x0b8aa35c28b7c6ab18f11dc168f437a8a69fd4f8`


superKalo (approved):
Confirm, after this change - separate positions inside a position are sorted correctly.

PR #2479, opened 2026-06-12
Fix / SignMessage - dapp verification banner propagate update

jordan-enev:
Tomek reported that from time to time he sees the FAILED_TO_GET dapp verification banner for heyAura staging when signing.

My theory is that we don't propagate dapp controller updates to signMessage, so the initial banner value FAILED_TO_GET gets cached - simply because it was the very first value, set while the storage hadn't loaded yet in the dapps controller.

With this PR, I'm introducing 2 changes:
1. First, while the dapp controller is loading, the banner resolves to the LOADING state.
2. Second, propagate dapp controller updates to the signMessage controller, so it receives the correct banner state.

https://goodmorning-dev.slack.com/archives/C0644227TAP/p1781007768349939

### DIFF
--- a/src/controllers/dapps/dapps.test.ts
+++ b/src/controllers/dapps/dapps.test.ts
@@ -301,6 +301,40 @@ describe('DappsController', () => {
       }
     })
 
+    test('should return loading banner while the initial storage load is still pending', async () => {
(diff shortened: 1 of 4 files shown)


superKalo (approved):
Great finding 👏

PR #2491, opened 2026-06-16
Optimize / Storage migrations

sonytooo:
## Optimize storage migration sweep

`StorageController` ran 20 migrations on every boot, and since `get`/`set`/`remove` all block on that sweep, it held up app load (especially on mobile, where each storage call is a postMessage round-trip to native RN storage).

Each migration re-read `passedMigrations` and read its data keys *before* checking whether it had already run. In the steady state (returning user, everything already migrated) this meant **~41 reads + ~3 writes** on every single launch.

### Changes
- Read `passedMigrations` **once** into an in-memory `Set`; every migration now checks it **first** and returns before touching storage.
- `passedMigrations` is written only when a migration actually does work (`#markMigrationPassed`).
- Added missing guards to 4 migrations that previously ran (and 2 that wrote) on every boot.

No behavioral change - each migration performs the same transformation under the same conditions; only redundant I/O was removed.

### Result
Steady-state boot: **~44 storage requests  ->  1** (~97% fewer).

### Tests
Added `migration sweep performance` tests asserting a fully-migrated install does 1 read / 0 writes, and that the newly-guarded migrations don't re-run.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ambire-common",
-  "version": "2.99.0",
+  "version": "2.101.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "ambire-common",
-      "version": "2.99.0",
+      "version": "2.101.1",
       "dependencies": {
         "@ambire/signature-validator": "^1.5.0",
         "@lifi/types": "^17.7.1",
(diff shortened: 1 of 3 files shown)


PetromirDev (approved):
👏

superKalo (approved):

PR #2492, opened 2026-06-17
Feature/scan for safes

Oxbobby:
## Summary

Adds on-demand discovery of Safe accounts during account import and changes smart account discovery so smart-account keys are derived only when the user explicitly scans for linked smart accounts.

## Changes

- Add Safe discovery helpers that use Safe API Kit to:
  - find Safes by owner address
  - fetch Safe creation/config data
  - return import-ready Safe metadata
- Add `AccountPickerController.scanForSafeAccounts(ownerAddrs)` to discover Safes for selected owner accounts across enabled Safe-supported networks.
- Track separate loading/completion state for linked smart account scans and Safe scans.
- Make linked smart account discovery explicit/on-demand instead of deriving smart-account-only keys during initial page load.
- Preserve the initial account page as basic EOAs only until smart account discovery is requested, improving hardware wallet flows such as Ledger.
- Refactor Safe import lookup to reuse the new shared `getSafeImportInfo` helper.
- Treat already-imported Safe accounts as imported with the same keys.
- Update account picker tests for the new on-demand smart account derivation flow


Oxbobby:
This was implemented in a different way in another pr

PR #2496, opened 2026-06-19
 Fix account picker infinite loading during Ledger import

Maskln:
paired with: https://github.com/AmbireTech/ambire-app/pull/7408

## Summary
- Add `#setPageGeneration` to invalidate in-flight `setPage()` requests and clear `accountsLoading` on `reset()`
- Fix invalid page (`<= 0`) leaving the spinner active
- Apply `pageSize` from `setInitParams` earlier so the UI can become ready sooner
- Improve Ledger error normalization for benign SDK states (`none`, empty interaction)
- Add test for invalid page clearing `accountsLoading`

### DIFF
--- a/src/controllers/accountPicker/accountPicker.test.ts
+++ b/src/controllers/accountPicker/accountPicker.test.ts
@@ -125,6 +125,24 @@ describe('AccountPicker', () => {
     })
   })
 
+  test('should clear accountsLoading when an invalid page is requested', async () => {
+    const { controller } = await prepareTest()
+    const keyIterator = new KeyIterator(process.env.SEED)
+    controller.setInitParams({
+      keyIterator,
+      hdPathTemplate: BIP44_STANDARD_DERIVATION_TEMPLATE,
+      shouldGetAccountsUsedOnNetworks: false,
+      shouldSearchForLinkedAccounts: false,
+      shouldAddNextAccountAutomatically: false
+    })
+    await controller.init()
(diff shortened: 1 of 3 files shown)


src/controllers/accountPicker/accountPicker.test.ts
```diff
@@ -125,6 +125,24 @@ describe('AccountPicker', () => {
...
+      shouldSearchForLinkedAccounts: false,
+      shouldAddNextAccountAutomatically: false
+    })
+    await controller.init()
+    await controller.setPage({ page: 0 })
+
+    expect(controller.accountsLoading).toBe(false)
+    expect(controller.pageError).toBeTruthy()
+    expect(controller.page).toEqual(DEFAULT_PAGE)
+  })
```
superKalo, line 144: I don't see the context behind this test. `setPage({ page: 0 })` isn't reachable anyways - every caller in the UI already guards it (pagination, navigation, etc).

This either makes no sense, or covers a corner case that's not explained?

Meanwhile the change this PR actually exists for - the `#setPageGeneration` stale-request guard - has no test coverage.

So either swap this test for one that calls two `setPage` calls (assert the later page wins), or one where `reset()` interrupts an in-flight `setPage` and test if `accountsLoading` still lands on `false`.

Those would fail before the fix and actually protect the race, which would ultimately validate that the changes in this PR are actually fixing problem(s).

PR #2501, opened 2026-06-22
Fix / Slow unlock when jit is disabled

sonytooo:
Speed up keystore unlock by using WASM scrypt

**Description:**
Keystore unlock derives the key with scrypt. The browser has no native scrypt, so it runs in our own code — and in JavaScript that takes ~45s when the user disables JIT optimization for security (a popular browser setting). This switches the browser path to a WASM scrypt (hash-wasm), which stays fast (~0.3s) regardless of that setting, and falls back to noble JS if WASM is unavailable. Mobile keeps its native implementation.
Output is byte-identical to the previous scrypt, so existing keystores unlock unchanged — no migration.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ambire-common",
-  "version": "2.101.1",
+  "version": "2.102.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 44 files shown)


superKalo (approved):

PetromirDev (approved):
👏

PR #2502, opened 2026-06-23
Add: protection against incorrect utf8 message humanizations

Oxbobby:


src/libs/humanizer/index.ts
```diff
@@ -178,7 +184,16 @@ const humanizeMessage = (_message: Message, options?: HumanizeMessageOptions): I
...
+      humanizerTMModules
+        .map((m) => {
+          try {
+            return m(message)
+          } catch (error) {
+            if (!isInvalidUtf8MessageError(error)) console.error(error)
+            return {}
+          }
+        })
+        .filter((p) => p.fullVisualization?.length)[0] || {}
```
PetromirDev, line 190: Fix it in the opensea module. The bug can be reproduced very easily with the following request:


window.ethereum.request({
  method: 'personal_sign',
  params: [
'0x8a33edcfea6ce92596a1d3cc16ac06a5d1a7280593e80c7204f6fec560025af9', '0x5AC95b0cfc1CFC92bF0678b4fb45E4532023a5c7' <--- Replace with your account
]
})

Image

Otherwise having a module level try catch is okay, it's just that we don't need to ignore the error, but fix it.

PetromirDev (changes requested):

src/libs/humanizer/index.ts
Oxbobby: resolved

PetromirDev (approved):

PR #2503, opened 2026-06-23
Fix: persist account daap preferences

Oxbobby:


sonytooo (approved):

PetromirDev (changes requested):
The fix is perfect, nice catch 👏 Sadly, the tests are not covering the bug properly and pass even if you revert dapps.ts to the state before the changes. This is caused by two separate issues:
- Setting version `1.0.0` skips the dapp fetch request entirely
- sources-dapp.com is not a part of the mocked dapp list (mockDapps.ts), thus the dapps from the request don't collide with the one in storage

I have prepared a fix for both and can push it if you are okay (had to fix it anyway to prove that there is a problem)

Image

PetromirDev (approved):

PR #2505, opened 2026-06-24
Feature / Trending tokens

sonytooo:
* Add trending tokens to the DappsController. Later the dapps ctrl is planned to be renamed to the more generic name ExploreController to match to FE functionality


src/controllers/dapps/dapps.ts
```diff
@@ -158,6 +183,13 @@ export class DappsController extends EventEmitter implements IDappsController {
       5 * 60 * 1000 // 5min.
     )
 
+    // Trending tokens are refreshed server-side every 10 minutes; poll at the same cadence.
+    this.#updateTrendingTokensInterval = new RecurringTimeout(
+      () => this.continuouslyUpdateTrendingTokens(),
+      TRENDING_TOKENS_UPDATE_INTERVAL,
+      this.emitError.bind(this)
+    )
```
PetromirDev: Should we move it to continuous updates?

src/controllers/dapps/dapps.ts
```diff
@@ -511,6 +546,72 @@ export class DappsController extends EventEmitter implements IDappsController {
...
+      return
+    }
+
+    this.#continuouslyUpdateTrendingTokensPromise = this.#continuouslyUpdateTrendingTokens()
+      .catch((err) => {
+        this.#updateTrendingTokensInterval.updateTimeout({
+          timeout: TRENDING_TOKENS_FAILED_UPDATE_INTERVAL
+        })
+        throw err
+      })
```
PetromirDev: Why do we rethrow the error here and not emit an error?

src/consts/intervals.ts
```diff
@@ -12,6 +12,8 @@ export const BLACKLIST_UPDATE_INTERVAL = 8 * 60 * 60 * 1000 // 8 hrs
 export const PHISHING_INACTIVE_UPDATE_INTERVAL = 6 * 60 * 60 * 1000 // 6 hrs
 export const PHISHING_ACTIVE_UPDATE_INTERVAL = 15 * 60 * 1000 // 15 minutes
 export const PHISHING_FAILED_TO_GET_UPDATE_INTERVAL = 600000 // 10 minutes
+export const TRENDING_TOKENS_UPDATE_INTERVAL = 10 * 60 * 1000 // 10 minutes
```
PetromirDev: Isn't this way too often? There are 30 tokens, yes, but we are updating the tokens regardless of whether the extension is active or not. Imo, we shoud run the interval while the extension is active and simply update on popup open if the last update is more than X minutes ago. If the interval runs less often then it will be okay to run it always (e.g., every 4+ hours)

PetromirDev (changes requested):

JIOjosBG (changes requested):
i updated properties on the backend to reduce request size, you can merge this here https://github.com/AmbireTech/ambire-common/pull/2529

src/controllers/continuousUpdates/continuousUpdates.ts
```diff
@@ -229,6 +268,46 @@ export class ContinuousUpdatesController extends EventEmitter {
...
+        })
+      }
+    } catch (err) {
+      // Back off to the fast failed-retry cadence, then rethrow so RecurringTimeout's onError
+      // handler reports it (with level 'silent', i.e. no user-facing toast).
+      this.#updateTrendingTokensInterval.updateTimeout({
+        timeout: TRENDING_TOKENS_FAILED_UPDATE_INTERVAL
+      })
+      throw err
+    }
```
PetromirDev, line 320: There are no max retries - in theory it could keep trying every minute for as long as the API is down/there is some error. Imo we should not retry more than 5-10 times. Also cover this in the test.

src/controllers/dapps/dapps.test.ts
```diff
@@ -42,6 +96,14 @@ const prepareTest = async (
         }
       }
 
+      if (url === TRENDING_TOKENS_URL) {
+        return {
+          ok: true,
+          status: 200,
+          json: async () => mockTrending
+        }
+      }
```
PetromirDev, line 105: Let's have one integration test where we don't mock the API. It has happened before that the API starts returning a structure that the UI doesn't expect/cannot handle. It's better to have most tests mock it but let's have one that calls the API.

src/interfaces/dapp.ts
```diff
@@ -64,6 +64,63 @@ export interface RecentDappEntry {
...
+  symbol: string
+  market_cap_rank: number | null
+  image?: { thumb?: string; small?: string; large?: string }
+  // Primary CoinGecko asset platform (e.g. 'ethereum') plus the per-platform contract addresses
+  // and decimals. Used to reuse the portfolio token-details components and match held balances.
+  // Null/absent for native coins without an on-chain contract (e.g. BNB).
+  asset_platform_id?: string | null
+  contract_address?: string
+  platforms?: { [platform: string]: string }
+  decimals?: { [platform: string]: number }
```
PetromirDev: Please make the comments public

PetromirDev (changes requested):

PetromirDev (approved):

PR #2506, opened 2026-06-24
Feature/colibri

Oxbobby:
## Summary

- Add a `VerificationController` that initializes Colibri-backed providers for enabled networks and tracks verifier status per chain.
- Add Colibri network support fields (`isColibriEnabled`, `colibriProverUrl`) plus default prover URLs for Ethereum, Sepolia, and Gnosis.
- Verify ENS resolutions through Colibri when a verifier is ready, store successful domains in `verifiedDomainsStatus`, and surface mismatch errors.
- Verify portfolio balances on Colibri-enabled chains by comparing normal RPC portfolio results against Colibri-verified results.
- Mark portfolio verification as `warning` when Colibri fails or balances differ, and `success` when the verified result matches.
- Skip Colibri portfolio verification when a simulation is present.
- Add focused tests for ENS verification mismatch/success, verified ENS transfer flow, and portfolio verification mismatch/success.


src/controllers/portfolio/portfolio.ts
```diff
@@ -995,6 +1030,181 @@ export class PortfolioController
...
+    } catch (error: any) {
+      this.emitError({
+        level: 'silent',
+        message: `Error while verifying portfolio through Colibri on ${network.name} (${network.chainId}).`,
+        error
+      })
+
+      return {
+        status: 'warning',
+        error: error?.message || 'Colibri could not verify portfolio balances'
```
jordan-enev: Every time I enable Colibri on Ethereum and go to the Dashboard, there's a balance warning. After a hard balance reload, it gets fixed.

src/controllers/domains/domains.ts
```diff
@@ -142,12 +215,34 @@ export class DomainsController extends EventEmitter implements IDomainsControlle
     }
 
     this.resolveDomainsStatus[domain] = 'LOADING'
+    delete this.resolveDomainsErrors[domain]
```
jordan-enev, line 352: `verifiedDomainsStatus[domain]` (lines below) is only ever set to 'VERIFIED', never cleared. In ambire-app this field alone drives the "Verified by Colibri" badge on the recipient - so a stale flag shows a false trust signal.

**How it breaks:**
1. resolveDomain('alice.eth') - Colibri verifies - verifiedDomainsStatus = 'VERIFIED' ✅
2. Cache expires; Colibri goes down (or alice.eth re-points to a new address)
3. resolveDomain('alice.eth') - getReadyProvider() null - returns false (no throw) - new UNVERIFIED address stored, but flag stays 'VERIFIED' ❌ badge still shows

**Fix:**
Clear it on each new resolution:

```
delete this.resolveDomainsErrors[domain]
delete this.verifiedDomainsStatus[domain]
```

src/controllers/domains/domains.ts
```diff
@@ -142,12 +215,34 @@ export class DomainsController extends EventEmitter implements IDomainsControlle
...
-      this.resolveDomainsStatus[domain] = undefined
+      try {
+        if (this.domainToAddresses[domain]?.address) {
+          const isEnsVerifiedByColibri = await this.#verifyEnsResolution({
+            providerChainId,
+            domain,
+            address: this.domainToAddresses[domain]!.address!,
+            isNamoshiDomain
+          })
+          if (isEnsVerifiedByColibri) this.verifiedDomainsStatus[domain] = 'VERIFIED'
```
jordan-enev, line 365: Why don't we handle "FAILED" as well? From UX & Security, the user may not understand that the Colibry validation failed.

jordan-enev (approved):

src/controllers/portfolio/portfolio.ts
```diff
@@ -995,6 +1030,181 @@ export class PortfolioController
     return this.#portfolioLibs.get(key)!
   }
 
+  initializeVerificationPortfolioLibIfReady(
```
jordan-enev: Since `PortfolioController` is already quite heavy, I suggest moving the portfolio verification logic into `VerificationController` (Colibry). Currently, the latter is only responsible for creating and syncing network providers, which makes it a good candidate to handle portfolio verification as well. We could apply the same approach to ENS verification.

I've created a draft PR to illustrate one possible implementation. Keep in mind that it's only an example and is meant to demonstrate the idea: https://github.com/AmbireTech/ambire-common/pull/2517

src/controllers/portfolio/portfolio.ts
Oxbobby: I think it depends on the portfolio size. Bigger portfolio like Vitalik's need extra time to complete

src/controllers/domains/domains.ts
Oxbobby: We're doing this in `#verifyEnsResolution` by throwing an error:
```
throw new Error(
  `ENS resolution mismatch for ${domain}: RPC returned ${address}, Colibri returned ${verifiedResult.address}`
)
```
Then in catch, we do `this.#setResolveDomainFailure(domain, e)`. 
Which does `this.resolveDomainsStatus[domain] = 'FAILED'`

src/controllers/domains/domains.ts
Oxbobby: great catch, thank you 🙏

src/services/provider/colibri.ts
```diff
@@ -0,0 +1,61 @@
...
+  getDefaultColibriProverUrl,
+  isColibriProviderAvailable
+} from '../../libs/networks/colibri'
+
+const assertSupportedBlockTag = (method: string, params: any[] | Record<string, any>) => {
+  if (!Array.isArray(params)) return
+  if (method !== 'eth_call' && method !== 'eth_estimateGas') return
+  if (params[1] !== 'pending') return
+
+  throw new Error('Colibri verifier does not support pending block tag requests')
```
simon-jentzsch, line 14: with Version 2.0.0 (will be released in 2 days) colibri will automaticly use the rpc directly for blockTag "pending", since pending can not be verified. So there would be no need to throw an exception as long as not verifiying pending is ok.

src/services/provider/colibri.ts
```diff
@@ -0,0 +1,61 @@
...
+class ColibriRpcProvider extends JsonRpcProvider {
+  #colibri: Colibri
+
+  constructor(network: NetworkInterface, proverUrl: string, options?: JsonRpcApiProviderOptions) {
+    super(network.selectedRpcUrl, Network.from(Number(network.chainId)), options)
+
+    this.#colibri = new Colibri({
+      chainId: Number(network.chainId),
+      rpcs: [network.selectedRpcUrl],
+      prover: [proverUrl],
```
simon-jentzsch: In order to make sure colibri is able to verify even if the prover is not available ( removing the single point of failure ), I would suggest to use an array which will be use the other urls as fallback if the first is not available. if you don't set the prover explicitly colibri will use defaults based on the chainId, which already use multiple provers from different datacenters as fallback.

src/services/provider/colibri.ts
```diff
@@ -0,0 +1,61 @@
...
+      zk_proof: true
```
simon-jentzsch, line 33: colibri supports different prover_modes. I would suggest to use the hybrid-mode: (See https://corpus-core.gitbook.io/specification-colibri-stateless/developer-guide/bindings/javascript-typescript#configuration)

prover_mode: 'hybrid'

This will make colibri to only use the prover only for cacheable Blockproofs and get all other data from the rpc-provider. Same security, but it makes sure, that those Blockheaders are well cached ( even through cloudflare ) so you won't get a bottleneck on the prover side even for many request.

package.json
```diff
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@ambire/signature-validator": "^1.5.0",
+    "@corpus-core/colibri-stateless": "^1.1.30",
```
simon-jentzsch, line 19: in 2 days the Version 2.0.0 will be released. It contains some breaking changes for the way the zk-proof is created ( using SP1 v6 instead of v5). So please make sure to use this Version ( also with important bugfixes) right from the start.

src/libs/networks/colibri.ts
```diff
@@ -0,0 +1,15 @@
+import { Network } from '../../interfaces/network'
+
+export const COLIBRI_PROVER_URLS_BY_CHAIN_ID: Record<string, string> = {
+  '1': 'https://mainnet1.colibri-proof.tech',
```
simon-jentzsch: I would suggest to use an array and the same defaults as colibri-stateless ( https://github.com/corpus-core/colibri-stateless/blob/dev/bindings/emscripten/src/chains.ts ) since they also contain fallbacks in order to remove the single point of failure.

simon-jentzsch (commented):
Here are some suggestions from the main contributer from colibri-stateless. Hope this helps.

src/controllers/portfolio/portfolio.ts
Oxbobby: I harden this to retry a couple of times before declaring failure

src/controllers/portfolio/portfolio.ts
Oxbobby: Merged and moved the logic there

src/services/provider/colibri.ts
Oxbobby: We changed the logic so no pending requests go to colibri. I think that should be fine?

package.json
Oxbobby: No problem. We just have a policy of waiting no less than 2 weeks after packages get released before merging. This means this PR will get scheduled for merging after a month or so.  
If that is the recommended approach, let's go for it

src/libs/networks/colibri.ts
Oxbobby: Done

src/services/provider/colibri.ts
Oxbobby: Decided to stick with 'remote' for now as our RPC I think cannot reliable return eth_getProofs

src/services/provider/colibri.ts
Oxbobby: Done

PR #2507, opened 2026-06-24
Feature/ Fetch ENS expiry data for the selected account and add additional transfer ENS validation

PetromirDev:
## To be done:
- Possibly add an affiliate address (cc @Ivshti  so we don't forget)

Implements https://github.com/AmbireTech/ambire-app/issues/6853#issuecomment-4478478774

## Changes:
- The obvious in domains/the ens service/transfer to implement the comment
- `sentToHistory` in the activity controller. We need it anyway for ENS so I figured we can also use it instead of traversing every account op
- `recipientDomain` to `Call` - I tested many different solutions in the past two days and trust me, this one is the most reliable because there are many edge cases - the user sends to an address that has ENS, but hasn't typed the ENS; the user sends to an ENS, but then starts a new transfer before the txn is confirmed; the user sends to an ENS, but then adds to batch. We MUST save the ENS to storage ONLY if the user has typed it in manually, the ENS resolves to a different address and the user has typed THE SAME name again. (e.g., yesterday nick.eth -> 1; today nick.eth -> 2) as it would be confusing if the user has never sent to the user by typing in the ENS and we warn him that it changed

## Relevant info and docs:
- https://docs.ens.domains/contracts/ensv2/overview
- https://docs.ens.domains/web/ensv2-readiness#preparing-for-ensv2
- https://docs.ens.domains/wrapper/overview
- https://docs.ens.domains/registry/eth
- ens.js expiry implementation ([core logic](https://github.com/ensdomains/ensjs/blob/1b90b888/packages/ensjs/src/functions/public/getExpiry.ts) and [custom hook](https://github.com/ensdomains/ens-app-v3/blob/main/src/hooks/ensjs/public/useExpiry.ts))

## Useful apps:
- Compare expiry data with https://app.ens.domains/
- Find recently expiring ENS domains https://ens.tools/find/expiry (under the tab "Grace period")

## How to test:
1. Dashboard banner - open https://ens.tools/find/expiry, pick an ENS, import it and you will see it
2. Transfer warning. Use the following script, reload the extension and then input the name in the send form. It's important that you do so manually and don't pick the account from the address dropdown as that will fill the field with the address and not display the warning (as it's not needed)
  ```
  const ENS_NAME    = 'vitalik.eth'
  const OLD_ADDRESS = '0x000000000000000000000000000000000000dEaD' 
  
  ;(async () => {
    const KEY = 'sentToHistory'
    const raw = (await chrome.storage.local.get(KEY))[KEY]
    const history = raw ? JSON.parse(raw) : { domains: {}, recipients: {} }
    history.domains ??= {}
  
    history.domains[ENS_NAME.toLowerCase().trim()] = {
      address: OLD_ADDRESS,
      sentAt: Date.now()
    }
  
    await chrome.storage.local.set({ [KEY]: JSON.stringify(history) })
    console.log('Seeded sentToHistory.domains:', ENS_NAME, history.domains[ENS_NAME.toLowerCase().trim()])
  })()
  ```


src/libs/accountOp/submittedAccountOp.ts
```diff
@@ -390,15 +390,29 @@ const transferIface = new Interface(['function transfer(address,uint256)'])
...
  * @param whitelist Optional list of addresses to filter the results.
  */
-export function getAccountOpRecipients(op: SubmittedAccountOp, whitelist?: string[]): string[] {
+export function getAccountOpRecipients(
+  op: SubmittedAccountOp,
+  whitelist?: string[]
+): {
+  address: string
+  domain?: string
+}[] {
```
superKalo, line 399: `getAccountOpRecipients` now returns `{ address, domain }[]`, which I think could break some unit tests, let's run the unit tests before merging this one (they haven't run because this PR is towards another feature branch).

src/services/ensDomains/ensDomains.ts
```diff
@@ -52,34 +68,39 @@ async function resolveENSDomain({
...
-        : NAMOSHI_UNIVERSAL_RESOLVER
-    })
+  const [address, avatar, expiry] = await Promise.all([
+    client.getEnsAddress({ name: normalizedDomainName, universalResolverAddress }),
+    client.getEnsAvatar({ name: normalizedDomainName, universalResolverAddress }),
+    // Namoshi domains live on a different chain without the ENS registrar/NameWrapper, so they have
+    // no ENS expiry to read.
+    options?.isNamoshiDomain
+      ? Promise.resolve(null)
+      : getEnsExpiry(provider, {
```
superKalo, line 94: This adds an extra eth_call on every forward resolve (i.e. while typing a name in the send form). The reverse path is gated by `shouldRefetchEnsExpiry`, but this one isn't. Just confirming `resolveDomain` doesn't fire per keystroke?

superKalo: Hm, I think this is a problem, try typing `superkalo.ethththththththththt` manually.

src/controllers/domains/domains.ts
```diff
@@ -20,6 +31,30 @@ import EventEmitter from '../eventEmitter/eventEmitter'
 // 15 minutes
 export const PERSIST_DOMAIN_FOR_IN_MS = 15 * 60 * 1000
 export const PERSIST_DOMAIN_FOR_FAILED_LOOKUP_IN_MS = 5 * 60 * 1000 // 5 minutes
+// Once a name is within the warn window, re-poll its expiry at most this often to catch a renewal.
+export const PERSIST_EXPIRY_FOR_IF_CLOSE_TO_DEADLINE_IN_MS = 1 * 60 * 60 * 1000
+
+/**
+ * Decides whether to (re)fetch a name's ENS expiry. A registration expiry can only be extended, never
+ * shortened, so once fetched and far from the deadline the cached value is safe and we don't refetch.
```
superKalo: Are u sure? AI said:

> This relies on "expiry only grows". True for `.eth` 2LDs (registrar), but a parent can shorten a subname's expiry via the NameWrapper. If someone's primary name is a subname, we could miss the "expiring" banner. Edge case, just flagging.

src/controllers/domains/domains.ts
```diff
@@ -108,7 +143,9 @@ export class DomainsController extends EventEmitter implements IDomainsControlle
       Object.entries(domainsFromStorage).filter(([addressInDomains, data]) => {
         if (!data) return false
 
-        // @TODO: Filter out expired entries
+        const isExpired = data.ensExpiry && data.ensExpiry.gracePeriodEndsAt < Date.now()
```
superKalo, line 157: When an ENS entry is expired, the whole entry is removed. If the same address also has a `namoshi` name, that gets dropped too. Edge case?

PetromirDev: Yes, thanks

src/controllers/domains/domains.ts
PetromirDev: Yup, didn't know about that. What options do we have - to refetch every X hours if it's a subname?

src/libs/accountOp/submittedAccountOp.ts
PetromirDev: Okay

src/services/ensDomains/ensDomains.ts
PetromirDev: Will check it

PetromirDev: This doesn't worry me as much because the requests are batched and I don't see a clear alternative. Otherwise we would have to fetch the ENS name first and then the avatar and expiry if it's a valid ens name. This will increase the time it takes to load IMO as most users will be writing valid ENS names, not invalid ones (as us who are testing). Also, fetching is debounced on the UI

superKalo (approved):

PR #2510, opened 2026-06-29
Socket v3 implementation

Oxbobby:
## Summary

Migrates the Socket/Bungee swap provider integration to Socket Swap V3 for new quotes, while keeping the legacy Bungee API available for existing active routes.

## Changes

- Added a new `socketv3` service based on the existing `socket` provider structure.
- Uses Socket Swap V3 endpoints for new quote flows:
  - `/v3/swap/quote?userOps=tx`
  - `/v3/swap/status?quoteId=...`
  - `/v3/swap/supported-chains`
  - `/v3/swap/tokens/list`
  - `/v3/swap/tokens/search`
- Keeps the old `socket`/Bungee provider route-only so users with pending legacy routes can still resolve them.
- Wires new Socket quotes through `providerId: socketv3`.
- Updates route UI/provider handling so Socket v3 routes render as Socket/Bungee instead of falling back to Li.Fi branding.
- Fixes Socket v3 bridge status parsing:
  - prefers `status` over `statusCode`
  - treats `statusCode: FULFILLED` as completed
  - uses the destination tx hash when available

### DIFF
--- a/src/services/socketv3/api.ts
+++ b/src/services/socketv3/api.ts
@@ -1,4 +1,5 @@
 import { getAddress } from 'ethers'
+import { ethAddress, zeroAddress } from 'viem'
 
 import SwapAndBridgeProviderApiError from '../../classes/SwapAndBridgeProviderApiError'
 import { CustomResponse, Fetch, RequestInitWithCustomHeaders } from '../../interfaces/fetch'
@@ -23,9 +24,7 @@ import { CITREA_CHAIN_ID } from '../squid/constants'
 import {
   AMBIRE_FEE_TAKER_ADDRESSES,
   ETH_ON_OPTIMISM_LEGACY_ADDRESS,
-  FEE_PERCENT,
-  NULL_ADDRESS,
-  ZERO_ADDRESS
+  FEE_PERCENT
 } from './constants'
 
 type SocketV3Protocol = {
@@ -130,7 +129,7 @@ type SocketV3StatusResponse = {
 }
 
 const convertZeroAddressToNullAddressIfNeeded = (addr: string) =>
-  addr === ZERO_ADDRESS ? NULL_ADDRESS : addr
+  addr === zeroAddress ? ethAddress : addr
 
 const normalizeIncomingSocketTokenAddress = (address: string) =>
   // incoming token addresses from Socket are all lowercased
@@ -168,7 +167,8 @@ const getRouteProtocol = (route: SocketV3Route): SocketV3Protocol => {
 const getNativeValue = (route: SocketV3Route) => {
   try {
     return BigInt(route.txData?.object.value || '0')
-  } catch {
+  } catch (e) {
+    console.error(e)
     return 0n
   }
 }
@@ -361,7 +361,7 @@ export class SocketV3API implements SwapProvider {
     // One is with the `ZERO_ADDRESS` and one with `NULL_ADDRESS`, both for ETH.
     // Strip out the one with the `ZERO_ADDRESS` to be consistent with the rest.
     if (toChainId === 1)
-      tokens = tokens.filter((token: SocketAPIToken) => token.address !== ZERO_ADDRESS)
+      tokens = tokens.filter((token: SocketAPIToken) => token.address !== zeroAddress)
 
     tokens = tokens.map(normalizeIncomingSocketToken)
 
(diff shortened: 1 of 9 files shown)


src/services/socketv3/constants.ts
```diff
@@ -0,0 +1,75 @@
...
+export const SOCKET_EXPLORER_URL = 'https://www.socketscan.io'
+
+export const PROTOCOLS_WITH_CONTRACT_FEE_IN_NATIVE = [
+  'stargate',
+  'stargate-v2',
+  'arbitrum-bridge',
+  'zksync-native',
+  'Stargate V2',
+  'Stargate'
+]
```
PetromirDev: None of these constants are actually used. The only places that uses them imports from socket/ (not the v3 folder).

src/services/socketv3/constants.ts
```diff
@@ -0,0 +1,75 @@
+import { SwapAndBridgeToToken } from '../../interfaces/swapAndBridge'
+
+export const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
+// Some services (like Socket) use the null token address to represent the
+// native token as the ZERO_ADDRESS is not standard for it.
+export const NULL_ADDRESS = '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee'
```
PetromirDev: We can use zeroAddress and ethAddress from viem

src/services/socketv3/api.ts
```diff
@@ -0,0 +1,607 @@
...
+    icon: details?.logoURI || ''
+  }
+}
+
+const getNativeValue = (route: SocketV3Route) => {
+  try {
+    return BigInt(route.txData?.object.value || '0')
+  } catch {
+    return 0n
+  }
```
PetromirDev, line 173: should we log an error here?

src/services/socketv3/api.ts
```diff
@@ -0,0 +1,607 @@
...
+    const response = await this.#handleResponse<SocketV3StatusResponse>({
+      fetchPromise: this.#fetch(url, { headers: this.#headers }),
+      errorPrefix: 'Unable to get the route status. Please check back later to proceed.'
+    })
+
+    const status = (response.status || response.statusCode || '').toUpperCase()
+    const statusCode = (response.statusCode || '').toUpperCase()
+    if (status === 'COMPLETED' || status === 'FULFILLED' || statusCode === 'FULFILLED')
+      return { status: 'completed', txnId: getStatusTxnId(response, txHash) }
+    if (status === 'REFUNDED')
```
PetromirDev: There doesn't seem to be a `status` FULFILLED https://docs.socket.tech/integrate/integration-guides/socket-api#step-4-poll-status

src/services/socketv3/api.ts
```diff
@@ -0,0 +1,607 @@
...
+        userTxs,
+        userAddress,
+        isOnlySwapRoute: fromChainId === route.output.token.chainId,
+        currentUserTxIndex: 0,
+        serviceTime: route.estimatedTime ?? 1,
+        fromChainId,
+        toChainId: route.output.token.chainId,
+        inputValueInUsd: response.input.valueInUsd,
+        toToken: {
+          address: route.output.token.address,
```
PetromirDev: normalize the address

src/services/swapIntegrators/swapProviderParallelExecutor.ts
```diff
@@ -22,16 +22,24 @@ export class SwapProviderParallelExecutor {
 
   #providers: SwapProvider[]
 
+  /**
+   * Used for legacy providers to prevent active routes from bricking
+   * while also disabling these from new quotes
+   */
+  #routeProviders: SwapProvider[]
+
```
PetromirDev: Nope, just get rid of it entirely. There is no point in keeping the legacy provider, because:
1. No new quotes can be retrieved from the legacy provider (this is expected)
2. But also activeRoutes are not fetched from storage, so there is absolutely 0 reason to keep the old provider

(Read the comment in swapAndBridge:629):
```
    // FIXME: Temporarily omit getting prev activeRoutes from storage, because of
    // old records with different (unexpected) structure causing crashes.
    // this.activeRoutes = await this.#storage.get('swapAndBridgeActiveRoutes', [])

    // FIXME: Figure out a mechanism to clean up these routes in storage,
    // otherwise this is a potential storage leak (although we have unlimited storage permission).
    // also, just in case protection: filter out ready routes as we don't have
    // retry mechanism or follow up transaction handling anymore. Which means
    // ready routes in the storage are just leftover routes.
    // Same is true for completed, failed and refunded routes - they are just
    // leftover routes in storage
    // const filterOutStatuses = ['ready', 'completed', 'failed', 'refunded']
    // this.activeRoutes = this.activeRoutes.filter((r) => !filterOutStatuses.includes(r.routeStatus))
```

PetromirDev (changes requested):

src/services/socketv3/constants.ts
Oxbobby: I removed most of them from the old socket provider consts and imported them from the new one. ZeroAddr and NullAddr are spread out so it's better to do most of them in a separate PR

PR #2513, opened 2026-06-30
Update / indexed db account ops i and phishing integration

gergana95:
Issue is https://github.com/AmbireTech/ambire-app/issues/7264
Related ambire-app PR https://github.com/AmbireTech/ambire-app/pull/7697

# Move transaction history and phishing lists to IndexedDB

Makes IndexedDB the source of truth for `accountsOps` (transaction history) and the phishing blocklist, with a one-time migration from the legacy `chrome.storage.local` blobs.

> **Before running this branch, delete your local `ambire` IndexedDB** (DevTools → Application → IndexedDB). `dbVersion` stays at 1, so a database created by an earlier build lacks the current indexes — reads throw `NotFoundError` and history renders **empty, silently**.

## Why

`accountsOps` was a single key-value blob, so every new transaction rewrote the entire history. That cost grows without bound — a heavy account is tens of MB re-serialised per write, and the whole blob is read on every service-worker wake-up.

IndexedDB gives row-level writes and a bounded startup read: all pending ops, plus the 20 most recent finalized ops per chain **for the selected account only**. Pagination then loads one page per rendered chain on demand.

Mobile is unaffected — no IndexedDB, so it keeps the key-value backend. Which adapter is live is recorded in `activityStorageBackend`.

## What's in it

- **Row-level writes.** One row per op, keyed `[accountAddr, chainId, id]`.
- **Bounded startup read** — `loadStartupOps(finalizedFor)`, scoped to the selected account.
- **Paginated reads.** `ensureRecentLoaded` fetches `(fromPage + 1) × itemsPerPage` per *rendered* chain, so a disabled network can't consume the page budget or inflate the total.
- **Indexed duplicate guard.** A `multiEntry` `by-txn-id` index makes the external-op check a point lookup reading **no records** — flat in history size, and it covers per-call txnIds (the MultipleTxns shape) that a plain `op.txnId` index cannot see.
- **Recipient index.** `sentToHistory.recipients` answers "have I sent here before?" and the address-poisoning check with zero backend reads, backfilled once by `#indexSentToHistoryFromAccountsOps` in `StorageController`.
- **Phishing snapshot** moved to its own store via `PhishingPersistence`, with a degrading read so a failure can't leave an empty blocklist for the session.

## Failure paths, and how they're covered

Each needs a specific call to fail on demand, so none are reachable from the console. All were verified by removing the fix and confirming the matching test fails — not by a green suite.

| Fix | Needs | Test |
|---|---|---|
| Key-value fallback when migration fails | `migrateFromStorage` to throw | a failed migration keeps writes out of IDB |
| `#load` never rejects | `storage.get('activityStorageBackend')` to throw | init survives the post-load history checks throwing |
| Stranded history is reported | backend recorded as `idb`, IDB absent | reports stranded history when IDB is missing |
| Duplicate guard fails closed | `hasOpWithTxnId` to throw | a failed lookup skips the op |
| `updateOps` aborts on a malformed op | an op with no timestamp | a malformed op aborts the batch |
| Merge keeps in-memory-only ops | `putSingleOp` to fail | an op that failed to persist still survives |

## Follow-ups

- **Phishing store is read and written whole.** The store holds one document under `snapshot` with the full `domains` and `addresses` arrays. `#load()` does `new Set(phishing.domains)` over all of it on every service-worker wake-up, and `save()` rewrites the entire document for a delta that touched one domain. Moving the blob to IDB changed where it lives, not how it's accessed. Not a simple fix: `getDomainBlacklistedStatus` is synchronous and `DappsController` calls it from inside `.filter()` predicates, so the sets must stay in memory and the read is whole regardless of layout. Tracked separately — see issue #___.
- Decide legacy-key retention — bounded (delete once the migration has proven itself over a release).
- Telemetry before wide rollout.

Testing cases: 
1. Existing history survives (highest priority)
Make 3–4 txns on 2+ networks on the current release → load this branch → all history present, right order. Reload 3×, no duplicates.

2. New transaction
Send on Base with Activity open → appears Pending → flips to Confirmed on its own.

3. Pagination (a real bug was here)
Account with >25 txns:

Page to 2 and 3 → every page full, none blank
Disable a network you have history on, reload, page again → pages still full, that network gone from list and total
Dashboard "load more" to the end → stops cleanly
4. Counts
Account with >20 txns → count reflects full history, not ~20. New txn increments by 1.

5. Account removal
Remove an account with >20 txns, re-import → full history back, counts right.

6. Phishing (security)
Blocklisted domain blocked → reload extension 3× → still blocked.

7. Poisoning warnings (security)

Known recipient → no "first time" warning
New address → warning appears
Domain that now resolves elsewhere → shows the correct previous address
8. No duplicates
Receive an external transfer → appears once. Send on a scanned network → appears once. Use an account with >25 txns.

9. IndexedDB unavailable
Firefox private window → history still works, no errors.

More [here](https://github.com/AmbireTech/ambire-common/pull/2513#issuecomment-5634416109) from Petromir


src/services/storage/baseIdbStore.ts
```diff
@@ -0,0 +1,163 @@
...
+        resolve()
+      }
+
+      request.onupgradeneeded = (event) => {
+        const db = (event.target as IDBOpenDBRequest).result
+        if (!db.objectStoreNames.contains(this.config.storeName)) {
+          db.createObjectStore(this.config.storeName, { keyPath: this.config.keyPath })
+          console.log(`[BaseIdbStore] Created store "${this.config.storeName}"`)
+        }
+      }
```
PetromirDev: Do we need a custom wrapper when we can use https://www.npmjs.com/package/idb?

src/services/storage/activityIdb.ts
```diff
@@ -0,0 +1,387 @@
...
+function deserializeOp(serialized: string): SubmittedAccountOp {
+  return JSON.parse(serialized, (key, value) => {
+    if (!BIGINT_FIELDS.has(key) || typeof value !== 'string') return value
+    try {
+      return BigInt(value)
+    } catch {
+      return value
+    }
+  }) as SubmittedAccountOp
+}
```
PetromirDev: parse and stringify is not needed - indexed db uses structuredClone so it can handle bigints natively

src/services/storage/activityIdb.ts
```diff
@@ -0,0 +1,387 @@
...
+  constructor() {
+    super({
+      dbName: 'ambire',
+      storeName: 'accountsOps',
+      keyPath: ['accountAddr', 'chainId', 'id'],
+      dbVersion: 1
+    })
+  }
+
+  /**
```
PetromirDev: We need a global database instance that will handle the migrations of ALL controllers. Currently, if we initialize a second database for another controller, migrations will not work properly as one controller will migrate, bump the version and break the other.

src/services/storage/activityIdb.ts
```diff
@@ -0,0 +1,387 @@
+import { IActivityIdbStorage, InternalAccountsOps } from '../../interfaces/activity'
+import { SubmittedAccountOp, SubmittedAccountOpLike } from '../../libs/accountOp/submittedAccountOp'
+import { AccountOpStatus } from '../../libs/accountOp/types'
+import { BaseIdbStore } from './baseIdbStore'
```
PetromirDev: The goal was to stop loading/parsing everything. But:

- `loadStartupOps` does `getAll()` over the entire store (all accounts, all chains, all ops) then sorts and filters in the controller. There are **no secondary indexes** - only the primary key, which is ordered `[accountAddr, chainId, id]`, not by timestamp. So "20 most recent per group" and "all pending" cannot be queried efficiently and you fall back to reading the whole store at boot. Add an index on `[accountAddr, chainId, timestamp]` and `openCursor(..., 'prev')` per group to take 20, plus an index on `status` for pending.
- `addAccountOp` passes the full trimmed (account,chain) array to `putOpsForAccountAndChain`, which does range-delete + up-to-1000 `put`s (`#writeRecordToStore`, `:330`). And `persistAccountsOps` rewrites **every account x every chain** in memory on every status update / balance-change persist.

PetromirDev (changes requested):

src/services/storage/activityIdb.ts
```diff
@@ -0,0 +1,399 @@
...
+      )
+    }
+
+    // Sort each group descending by timestamp
+    for (const chainMap of Object.values(result)) {
+      for (const ops of Object.values(chainMap)) {
+        ops.sort((a, b) => b.timestamp - a.timestamp)
+      }
+    }
+
```
PetromirDev: This loads all pending ops, which is fine IMO, but also 20 ops per account:network pair, which can add up to many account ops if the user uses multiple networks with multiple accounts

src/controllers/activity/activity.ts
```diff
@@ -1469,8 +1628,22 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+    if (this.#activityIdb) {
+      try {
+        await this.#activityIdb.deleteAccount(address)
+      } catch (error) {
+        console.error(
+          'ActivityController: Failed to delete from IDB, falling back to storage',
+          error
+        )
+        await this.#storage.set('accountsOps', this.#accountsOps)
+        await this.#storage.set('signedMessages', this.#signedMessages)
```
PetromirDev: Why set to storage on failure? Also we are setting like 20 account ops per account:network instead of the full history, which will delete ops silently

PetromirDev (changes requested):
I love the idea of using indexeddb but don't think that investing much time into a POC is that viable. IMO you cannot compare inserting a single account op without serializing it to serializing up to 1000 ops per network for each account and then persisting that with `chrome.storage.local`. Let's have the three of us discuss this (@gergana95, @superKalo and me) and ship this. I have some considerations with the current state of the POC, mainly about migrations and not having an abstraction over storage. I won't share them yet because I know that this is a POC but want to discuss them and proceed with this one quicker as I think that it will be a great improvement.

src/services/storage/phishingIdb.ts
```diff
@@ -0,0 +1,157 @@
...
+// ─────────────────────────────────────────────────────────────────────────────
+
+// The 'phishing' store holds a single document keyed by this constant.
+// All reads and writes target this one record.
+//
+// NOTE: the store is deliberately NOT in AMBIRE_IDB_SCHEMA. Adding it would require a
+// dbVersion bump, and a shipped bump cannot be rolled back — so it is not worth
+// carrying for a store nothing reads yet. Add it in the same change that wires
+// PhishingController to this backend, not before.
+const STORE_NAME = 'phishing'
```
PetromirDev: "Adding it would require a
// dbVersion bump, and a shipped bump cannot be rolled back" - it is not shipped yet so I see no reason to not add it there

src/interfaces/activity.ts
```diff
@@ -1,5 +1,101 @@
...
+  /**
+   * One-time migration from legacy chrome.storage.local to IDB.
+   * Called with callbacks so the interface stays decoupled from IStorageController.
+   * IDB backend: migrates if empty; storage backend: no-op.
+   *
+   * isEmpty() and migrateFromStorage() are NOT part of this interface — they are
+   * implementation details of how ActivityIdbStorage decides whether to migrate,
+   * not something callers need polymorphically. Adding them here would force
+   * ActivityKeyValueStorage to carry dead stub methods it never uses.
+   */
```
PetromirDev, line 26: Comments are way too verbose and hard to read. Please simplify and shorten long comments.

src/controllers/activity/activity.ts
```diff
@@ -730,9 +1166,52 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+    // evicted the oldest row to stay under MAX_IDB_GROUP_SIZE, making the net change 0.
+    //
+    // ACCEPTED LAG: this lands after the emitUpdate above, so a banner gated on
+    // minTxnsTotal/maxTxnsTotal sees the new total only on the NEXT update. Refreshing
+    // before emitUpdate would block the UI on an IDB read, and emitting a second update
+    // just for the count would re-serialize the whole controller state per broadcast.
+    // The thresholds are coarse buckets, so one update of staleness does not change which
+    // bucket an account lands in for any realistic value.
+    await this.#refreshTotalOpsCount(accountAddr)
+
```
PetromirDev: Comments are way too verbose and hard to read. Please simplify and shorten long comments.

src/controllers/activity/activity.ts
```diff
@@ -730,9 +1166,52 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+    // Recounted after the write rather than incremented, because putSingleOp may have
+    // evicted the oldest row to stay under MAX_IDB_GROUP_SIZE, making the net change 0.
+    //
+    // ACCEPTED LAG: this lands after the emitUpdate above, so a banner gated on
+    // minTxnsTotal/maxTxnsTotal sees the new total only on the NEXT update. Refreshing
+    // before emitUpdate would block the UI on an IDB read, and emitting a second update
+    // just for the count would re-serialize the whole controller state per broadcast.
+    // The thresholds are coarse buckets, so one update of staleness does not change which
+    // bucket an account lands in for any realistic value.
+    await this.#refreshTotalOpsCount(accountAddr)
```
PetromirDev: We could store the count in memory, update the in memory counter before the emit and expose it instead of reading from the database.

src/controllers/activity/activity.ts
```diff
@@ -356,6 +406,311 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+    } catch (error) {
+      this.emitError({
+        level: 'silent',
+        message: 'Part of your transaction history could not be loaded.',
+        error:
+          error instanceof Error
+            ? error
+            : new Error('ActivityController: failed to expand full history from IDB')
+      })
+    }
```
PetromirDev: 300+ rows of IDB/storage code has leaked into the activity controller, which makes it harder to read. Let's figure out a way to have this abstracted inside of the storage class. A possible idea is to have a class called `PersistentStorage`, pass idb+storage to it and have it handle all of this logic internally. Best case scenario is to have the activity simply call different methods instead of adding all of this new logic in.

PetromirDev: Make the class accept different storage services. Based on the service it does something internally. This way we can have an adapter for `expo-sqlite` on mobile, idb on desktop and storage for most other controllers.

PetromirDev (changes requested):
Also, please update the description of the PR

src/services/storage/activityIdb.ts
PetromirDev: Once again - is there a specific reason to fetch the ops of other accounts on startup?

src/interfaces/activity.ts
```diff
@@ -1,5 +1,87 @@
...
+    accountAddr: string,
+    chainId: bigint | string
+  ): Promise<SubmittedAccountOp[] | undefined>
+
+  /** Write ops for one (account, chainId) pair. */
+  putOpsForAccountAndChain(
+    accountAddr: string,
+    chainId: bigint | string,
+    ops: (SubmittedAccountOp | SubmittedAccountOpLike)[]
+  ): Promise<void>
```
PetromirDev: Never used (only by tests)

src/services/storage/accountOpsPersistence.ts
```diff
@@ -0,0 +1,395 @@
...
+
+  /**
+   * Expand the given accounts from the startup window to their full history, for callers
+   * that must reason over every past op rather than the recent slice.
+   */
+  async ensureFullHistory(accountAddrs: string[]): Promise<void> {
+    if (!this.loadsPartially) return
+
+    await Promise.all(accountAddrs.map((addr) => this.#expandAccount(addr)))
+  }
```
PetromirDev, line 106: Imo ensureFullHistory and expandAccount are not needed. We should just migrate to `sentToHistory` entirely and delete the heavy logic in `hasAccountOpsSentTo`. This will make this much simpler and faster

src/controllers/activity/activity.ts
```diff
@@ -730,9 +795,22 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
     this.emitUpdate()
+
+    // Persistence LAST, per the storage rule in controllers/AGENTS.md — the key-value
+    // adapter rewrites the whole blob, so awaiting it earlier would block the UI on
+    // serializing the entire history.
+    //
+    // Do not "fix" this by persisting first: syncFilteredAccountsOps() above may expand this
+    // group from the backend before the op is written, and the merge inside
+    // AccountOpsPersistence is what keeps the memory-only op.
+    await this.#persistence.addOp(accountAddr, chainId, accountOp, trimmedId)
```
PetromirDev: Let's have `trim` return the item it popped instead of pre-computing the id here. Then this
becomes `const evicted = trim(group)` and we pass `evicted?.id` down to `addOp`. The
controller doesn't need to know about the IDB group cap or repeat the 1000.

src/services/storage/activityIdb.ts
```diff
@@ -0,0 +1,609 @@
...
+ *
+ * Exported because ActivityController needs it to decide whether an in-memory
+ * group is still the startup window or has already been expanded from IDB.
+ */
+export const STARTUP_RECENT_OPS_LIMIT = 20
+// Hard cap on the number of IDB rows per (account, chainId) group.
+// The in-memory cap is enforced by ActivityController; this guards against IDB
+// accumulating more rows than the in-memory limit (e.g. after a startup that
+// loaded only the 20-op subset before the limit was enforced).
+const MAX_IDB_GROUP_SIZE = 1000
```
PetromirDev: 1000 is in three places now - here, the `trim` default in activity.ts and the `>= 1000`
check in `addAccountOp`. Let's export one `MAX_OPS_PER_GROUP` and use it in all three.

src/services/storage/activityIdb.ts
```diff
@@ -0,0 +1,609 @@
...
+import { AMBIRE_IDB_SCHEMA } from './idbSchema'
+
+/**
+ * Finalized ops loaded per (account, chainId) at startup. Pending ops are always
+ * loaded in full on top of this.
+ *
+ * Exported because ActivityController needs it to decide whether an in-memory
+ * group is still the startup window or has already been expanded from IDB.
+ */
+export const STARTUP_RECENT_OPS_LIMIT = 20
```
PetromirDev: The comment says the controller needs this to tell whether a group is still the startup
window, but nothing outside this file imports it - that moved to `#fullyLoadedGroups`.
Make it private and drop the comment.

src/services/storage/activityIdb.ts
```diff
@@ -0,0 +1,609 @@
...
+
+        if (percentUsed > 80) {
+          console.warn(
+            `[ActivityIdbStorage] ${this.#storeName} quota usage high (${percentUsed.toFixed(1)}%)`
+          )
+        }
+      })
+      .catch(() => {})
+  }
+
```
PetromirDev: We have `unlimitedStorage` in the manifest, so `estimate()` reports a synthetic quota and
`percentUsed > 80` will basically never be true. So this is an async call plus a log on
every single write, in the SW, for a warning we can't reach. Let's drop it

src/services/storage/idbDatabase.ts
```diff
@@ -0,0 +1,189 @@
...
+ *
+ * Startup order, the schema/data migration distinction, and the rules for writing a
+ * migration handler are documented in ./README.md. Read it before changing this file.
+ */
+
+import { IDBPDatabase, IDBPTransaction, openDB } from 'idb'
+
+import { AMBIRE_IDB_SCHEMA, IdbStoreDef } from './idbSchema'
+
+export type AmbireIdbDatabase = IDBPDatabase<any>
```
PetromirDev: `IDBPDatabase<any>` means no checking at all on store names or row shapes, and it is the type
that gets threaded through `MainController` and `ActivityController`. `AMBIRE_IDB_SCHEMA`
already holds everything needed to declare a `DBSchema`

PetromirDev (changes requested):
The implementation is good and will soon be ready for testing. Imo we will need to do some refactoring to reduce code repetition, but this will be after migrating the phishing controller and especially after implementing a database solution for the mobile app. I don't think that there is a reason to do these refactorings beforehand as we don't know exactly what the future implementations will need. That's why I think that we should polish this one, ensure data integrity + smooth migration and just ship it.

src/controllers/activity/activity.ts
```diff
@@ -720,19 +752,90 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+      this.emitError({
+        level: 'silent',
+        message: 'Your past recipients could not be indexed.',
+        error:
+          error instanceof Error
+            ? error
+            : new Error('ActivityController: sentToHistory backfill failed')
+      })
+    }
+  }
```
PetromirDev, line 823: This should be handled in the storage controller where we have all other migrations

src/services/storage/phishingIdb.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  async loadSnapshot(): Promise<PhishingSnapshot> {
+    const row = await this.#db.get(this.#storeName, SNAPSHOT_KEY)
+    if (!row) return { ...DEFAULT_PHISHING_SNAPSHOT }
+    const { id: _id, ...snapshot } = row
+    return snapshot
+  }
+
+  async saveSnapshot(data: PhishingSnapshot): Promise<void> {
+    await this.#db.put(this.#storeName, { id: SNAPSHOT_KEY, ...data })
```
PetromirDev, line 75: The phishing store is one document under `snapshot` holding the whole `domains` and
`addresses` arrays, and `#load()` still does `new Set(phishing.domains)` on all of it. So
we moved the blob but kept reading it whole - and `save()` still rewrites the entire
document for a delta that touched one domain.

src/services/storage/accountOpsPersistence.ts
```diff
@@ -0,0 +1,337 @@
...
+      false
+    )
+  }
+
+  #setMigratedFlag(value: boolean): Promise<void> {
+    return (this.#storage.set as (key: string, value: boolean) => Promise<void>)(
+      'activityIdbMigrated',
+      value
+    )
+  }
```
PetromirDev, line 263: `activityIdbMigrated` is written in two places and read nowhere. Also, I'm thinking of something else - instead of storing migrated/not migrated I think we should store the active adapter as migrated is too vague. Also, the type casts are a bit weird, we don't have this pattern elsewhere

src/controllers/activity/activity.ts
```diff
@@ -465,22 +472,44 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+
+    // So pagination can page past the startup window, which holds finalized ops for the
+    // selected account only. With no chain filter every enabled chain has to be expanded, or a
+    // just-switched-to account would render its pending ops and nothing else.
+    await Promise.all(
+      (filters.chainId ? [filters.chainId] : enabledNetworkChainIds).map((chainId) =>
+        this.#persistence.ensureGroupLoaded(filters.account, chainId)
+      )
+    )
+    internalAccountOpsByChain = this.#accountsOps[filters.account] || internalAccountOpsByChain
```
PetromirDev, line 486: Why load all ops at once when we can only load the requested items?

src/controllers/activity/activity.ts
```diff
@@ -1521,12 +1630,21 @@ export class ActivityController extends EventEmitter implements IActivityControl
 
     delete this.#accountsOps[address]
     delete this.#signedMessages[address]
+    // Recipients are keyed per account, so they go with it. `domains` is
+    // deliberately left alone — it is a global index shared across accounts
+    // (see SentToHistory in ./types.ts).
+    delete this.#sentToHistory.recipients[address]
```
PetromirDev, line 1637: let's not delete this for now. It's cheap, yet valuable information

PetromirDev (changes requested):
I will review it once again tomorrow. Let's get @superKalo or @sonytooo on board as well. The criteria I have while reviewing:
- Having idb allows us to load only the information subset we need (e.g., 1 domain/20 ops) and we should utilize this
- Storage solutions should be pluggable - the implementation should allow us to add expo sqlite on mobile in the future
- Controllers shouldn't need a lot of changes and adapter-specific code should be abstracted away

Most of this is already implemented, I'm just sharing so we are on the same page and don't provide contradicting feedback.

src/services/storage/phishingIdb.ts
PetromirDev: This may not be 100% possible as it would require an async method.

src/controllers/activity/activity.ts
```diff
@@ -374,67 +416,42 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
-        found: true,
-        lastTransactionDate: new Date(lastSentAt),
-        addressPoisoningMatch: null
-      }
+    // Deliberately not a scan of #accountsOps: recipients outlive MAX_OPS_PER_GROUP eviction,
+    // so the map is the more complete source. See README.md, "The recipient index".
+    const accounts = accountId ? [accountId] : Object.keys(this.#sentToHistory.recipients)
 
-    const accounts = accountId ? [accountId] : Object.keys(this.#accountsOps)
-    let found = false
```
PetromirDev, line 389: I know this is legacy logic but it doesn't make much sense. This is called in transfer where the user must have an account. Make it so it returns nothing if !accountId

src/controllers/activity/activity.ts
```diff
@@ -720,43 +791,42 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+    //
+    // Do not "fix" this by persisting first: syncFilteredAccountsOps() above may expand this
+    // group from the backend before the op is written, and the merge inside
+    // AccountOpsPersistence is what keeps the memory-only op.
+    await this.#persistence.addOp(accountAddr, chainId, accountOp, evicted?.id)
+
+    // sentToHistory is a small durable index and always lives in key-value storage,
+    // never IDB. Persisting it here is what lets the recipient fast path in
+    // hasAccountOpsSentTo survive a service worker restart — without this the map
+    // is rebuilt empty on every wake-up and every recipient looks new again.
```
PetromirDev, line 828: These comments are a bit too much. Let's shorten them to 1 line each

src/controllers/activity/activity.ts
```diff
@@ -806,8 +876,24 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+
+      // Read failed, so absence proves nothing. Treated as a duplicate: a stored duplicate is
+      // permanent, while a skipped op is re-offered on the scanner's next pass.
+      if (!storedInternalOps) return true
+
+      // Kept alongside the stored read because IDB keys are exact while this lookup is
+      // case-insensitive — the same account addressed with different casing is stored under a
+      // key the read above cannot find.
+      const inMemoryInternalOps = getAccountOpsForAccountAndChain(
         this.#accountsOps,
```
PetromirDev, line 897: Seems inefficient to pull everything out of indexedb to check for a few ids. Let's discuss this offline

src/controllers/activity/activity.ts
```diff
@@ -1521,12 +1609,21 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+    // sentToHistory is deliberately left intact, both the global `domains` map and this
+    // account's `recipients`. It is a few bytes per recipient, and keeping it means a
+    // re-imported account still recognises the addresses it has sent to before rather than
+    // treating every one as first-time — which is what drives the poisoning warning.
 
     await this.syncFilteredAccountsOps()
     await this.syncSignedMessages()
 
-    await this.#storage.set('accountsOps', this.#accountsOps)
+    // signedMessages and sentToHistory are always persisted to storage (not IDB).
```
PetromirDev, line 1603: This is not needed

src/controllers/activity/activityIdbMigration.test.ts
```diff
@@ -0,0 +1,1082 @@
...
+    expect(controller.emittedErrors).toHaveLength(0)
+  })
+
+  test('stays usable and silent when IDB is missing after a migration already completed', async () => {
+    // Previous session migrated into IDB. This session failed to open IDB, so the
+    // history exists but is unreachable. There is deliberately no user-facing surfacing
+    // for this — what must hold is that the controller still loads and reads the
+    // retained legacy blob instead of throwing or starting a divergent one.
+    await (storage.set as (key: string, value: boolean) => Promise<void>)(
+      'activityIdbMigrated',
```
PetromirDev, line 308: The key is gone - nothing writes or reads `activityIdbMigrated` any more, it is `activityStorageBackend` now. The cast is only there because the key is not in `StorageProps`, so this line sets nothing and the test does not set up the state its name describes. Let's set `activityStorageBackend` to `idb` instead - note that then `#recordActiveBackend` reports the stranded history and the `toHaveLength(0)` below fails, which is the behaviour we actually want to be asserting.

src/controllers/activity/activityIdbMigration.test.ts
```diff
@@ -0,0 +1,1082 @@
...
+    // IDB must have ops for this to bite: the flag writer returns early on an
+    // empty store, so without seeding, the flag is never read and the test would pass
+    // whether or not the guard exists.
+    await new ActivityIdbStorage(db).putMultiple([
+      { accountAddr: ACC, chainId: CHAIN_1, ops: [makeOp('some-op', 1000) as any] }
+    ])
+
+    const failing: IStorageController = Object.create(storage)
+    failing.get = (async (key: string, defaultValue?: any) => {
+      if (key === 'activityIdbMigrated') throw new Error('flag read failed')
```
PetromirDev, line 536: Same dead key here - `#load` never reads `activityIdbMigrated`, so this injected failure never fires and the test passes for a different reason than its name. The description lists it as the only coverage for the post-load guard. Let's point it at `activityStorageBackend`.

src/controllers/activity/activity.ts
```diff
@@ -509,7 +558,26 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
     }
 
-    const result = paginate(filteredItems, pagination.fromPage, pagination.itemsPerPage)
+    // Counted over the chains this call renders, so a disabled chain cannot inflate the total.
+    // An identifiedBy filter narrows to a single transaction and has no stored equivalent to
+    // count, so there the loaded items ARE the total.
+    const storedTotal = filters.identifiedBy
+      ? 0
+      : await this.#persistence.countOps(filters.account, chainIdsToRender)
+    const externalTotal = filters.identifiedBy
```
PetromirDev, line 567: `externalTotal` counts every external op on the chains we render, but `filteredItems` drops the ones whose `txnId` matches an internal op at the merge above. So `itemsTotal` and `maxPages` come out higher than what can actually render - the settings pagination offers empty pages and `hasMoreTxnToLoad` on the dashboard stays true with nothing left to load. Let's count the externals after the dedup filter rather than before it.

src/controllers/activity/activity.ts
```diff
@@ -465,22 +482,54 @@ export class ActivityController extends EventEmitter implements IActivityControl
...
+    // Load only what the requested page can need, rather than the whole history.
+    //
+    // The page is a union of stored ops, all pending ops (already in memory from startup) and
+    // all external ops (always in memory). Fetching the newest (fromPage + 1) x itemsPerPage
+    // stored ops of every rendered chain is sufficient: any item in the union's first N is
+    // either pending, external, or among the newest N of its own chain.
+    //
+    // Per chain and not account-wide: an account-wide fetch spends part of its budget on
+    // disabled networks, which are dropped below, and would render a short page.
+    const pageEnd = (pagination.fromPage + 1) * pagination.itemsPerPage
```
PetromirDev: `filterAccountsOps` runs from `syncFilteredAccountsOps` on every status poll, balance backfill and add/remove, not only when the page changes. Each run is a `getRecentOps` cursor walk of `pageEnd` steps plus a `count()` per enabled chain, per open session - 15 networks on page 5 is around 750 cursor round trips per poll. Let's skip both when `filters` and `pagination` are unchanged since the last call for that session.

src/controllers/activity/activity.ts
```diff
@@ -902,7 +989,7 @@ export class ActivityController extends EventEmitter implements IActivityControl
       submittedAccountOpLike.balanceChanges = undefined
     }
 
-    if (hasExistingAccountOpWithTxnId()) return
+    if (await hasExistingAccountOpWithTxnId()) return
```
PetromirDev, line 992: Adding to my comment above on the read itself - this is the second call, so every external op the scanner offers costs two full group reads out of indexeddb, not one. Worth folding into the same discussion.

src/services/storage/accountOpsPersistence.ts
```diff
@@ -0,0 +1,359 @@
...
+
+      return counts.reduce((total, count) => total + count, 0)
+    } catch (error) {
+      this.#report('Your transaction history could not be counted.', error, 'count ops')
+
+      return 0
+    }
+  }
+
+  async #refreshCount(accountAddr: string): Promise<void> {
```
PetromirDev, line 344: The note on `addOp` says it has to recount because `putSingleOp` may have evicted - but `putSingleOp` already knows whether it deleted a row, so it could return that and we adjust the counter by the delta. That keeps the number in memory and off the read path, which is what I was after originally.

src/services/storage/accountOpsPersistence.ts
```diff
@@ -0,0 +1,359 @@
...
+
+  /**
+   * Merge fetched rows into the cache, keeping the CACHED object on an id collision.
+   *
+   * A merge and not a replace, because the cache can hold ops the backend does not have yet
+   * (a just-broadcast op lands in memory before the write) and objects that in-flight work
+   * still mutates (status updates mutate across provider awaits). Replacing would drop the
+   * first and detach the second.
+   */
+  #mergeIntoCache(accountAddr: string, chainId: string, fetched: SubmittedAccountOp[]): void {
```
PetromirDev: The merge does not re-apply `MAX_OPS_PER_GROUP`, so paging deep grows the group past 1000 and the next `addAccountOp` only pops one back off. That makes the "enforced twice and the two MUST agree" note on the constant untrue. Let's slice to the cap after the sort.

src/services/storage/activityIdb.ts
```diff
@@ -0,0 +1,595 @@
...
+    const count = await tx.objectStore(this.#storeName).count()
+    return count === 0
+  }
+
+  // ──────────────────────────────────────────────────────────────────────────────
+  // Private helpers
+  // ──────────────────────────────────────────────────────────────────────────────
+
+  #writeRecordToStore(
+    store: any,
```
PetromirDev: `store: any` puts back the hole we typed away everywhere else - this is the one write path that is not checked against `AmbireIdbSchema`. Let's type it as the store off the transaction.

src/controllers/AGENTS.md
```diff
@@ -86,7 +86,56 @@ Most controllers have `initialLoadPromise` that resolves when the controller fin
...
+- Do not delete the legacy key as part of a data migration while the IDB path is still new. Keep it as a safety-net copy and record a completion flag (e.g. `activityIdbMigrated`) instead. The controller must _read_ from that copy when the migration fails, otherwise it shows an empty state for the whole session while the data sits intact one key away. The flag itself has no consumer today — `ActivityController` writes it (`#recordMigrationCompleted` in `AccountOpsPersistence`) because it can only be recorded while IDB works, and a session that cannot open IDB can no longer tell "never had transactions" from "history is in IDB and unreachable". There is deliberately no banner or error for that state: it is detected inside `#load()` on service-worker startup, where errors reach nothing because no window is open, and a banner for a condition the user cannot act on was judged not worth the noise.
```
PetromirDev, line 129: This still documents `activityIdbMigrated` and `#recordMigrationCompleted`, both renamed, and says the flag has no consumer - `#recordActiveBackend` reads it now to report stranded history. Let's update it to `activityStorageBackend`.

src/services/storage/README.md
```diff
@@ -0,0 +1,246 @@
...
+   noted so nobody assumes the in-memory workaround extends to either.
+
+## The startup window, and who has to care
+
+`loadStartupOps()` returns, per (account, chain): **all pending ops** plus the **20 most
+recent finalized** ones. So in-memory group lengths are *not* totals.
+
+Two mechanisms exist because of that:
+
+- expansion markers — a per-`(account, chain)` flag marking groups expanded to full history
```
PetromirDev, line 171: Expansion markers went away with `ensureFullHistory`, there is no per group flag any more. Let's drop this bullet.

src/services/storage/phishingIdb.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  async loadSnapshot(): Promise<PhishingSnapshot> {
+    const row = await this.#db.get(this.#storeName, SNAPSHOT_KEY)
+    if (!row) return { ...DEFAULT_PHISHING_SNAPSHOT }
+    const { id: _id, ...snapshot } = row
+    return snapshot
+  }
+
+  async saveSnapshot(data: PhishingSnapshot): Promise<void> {
```
PetromirDev: Following up on my own comment on the store - the sync argument in the README only defends the read, it does not defend this. `updatePhishing` already receives a delta from `get_update?version=N` and applies it op by op to the sets, then throws that away here and rewrites both arrays whole. A row per domain would let this write only the changed rows while `#load` still does one `getAll()` into the same in memory Set, so `getDomainBlacklistedStatus` stays sync.

src/interfaces/activity.ts
```diff
@@ -1,5 +1,74 @@
...
+  ensureMigrated(
+    getStoredOps: () => Promise<InternalAccountsOps>,
+    removeStoredOps: () => Promise<void>
+  ): Promise<void>
+
+  /**
+   * Pending ops for every account, plus the finalized window for `finalizedFor` only —
+   * finalized ops are only rendered for the account being viewed. Key-value reads everything.
+   */
+  loadStartupOps(finalizedFor?: string): Promise<InternalAccountsOps>
```
PetromirDev: One consumer of this window was missed - `findByIdentifiedBy` (activity.ts:1696) still answers out of `#accountsOps`, which on idb is now 20 finalized per chain plus the pending ones. `ambire_getAccountOpStatus` throws `invalid identifier passed` on a miss, so a dapp asking about a transaction that has aged out of the window gets an rpc error where it used to get a status. This is the in memory consumer audit the new `AGENTS.md` section warns about - let's have it fall back to a backend read when the cache misses.

src/controllers/activity/activity.ts
PetromirDev: the comment i mena

src/controllers/AGENTS.md
```diff
@@ -86,7 +86,56 @@ Most controllers have `initialLoadPromise` that resolves when the controller fin
...
+6. **If the IDB backend loads only a subset at startup, audit every in-memory consumer.** This is the easiest way to introduce a silent bug. `ActivityController`'s startup read returns only the 20 most recent finalized ops per chain (plus all pending ones), which quietly weakened address-poisoning detection. Anything that reasons over the _full_ history must not scan the cache — give it a **separate durable index** instead, kept up to date on write and backfilled once from existing data. `sentToHistory.recipients` is the reference: expanding the cache on demand worked but reloaded the whole history into memory on a user-facing path, so it was replaced. Note that such an index outlives the rows it was derived from, which is usually desirable (a recipient evicted from history still raises a lookalike warning) but means "no rows" no longer implies "no index entries".
```
PetromirDev, line 124: Take this out of AGENTS.md as it's very specific and not needed when working with controllers on a daily basis

src/services/storage/phishingIdb.ts
PetromirDev: Let's open an issue for this for later as to not slow down the PR further..

src/controllers/AGENTS.md
```diff
@@ -86,7 +86,56 @@ Most controllers have `initialLoadPromise` that resolves when the controller fin
...
+- **A `dbVersion` bump is effectively one-way.** Once a user's database has been upgraded, a build pinned to the older version cannot open it — `openDB` rejects with `VersionError`, `openAmbireIdb()` fails, and every controller silently falls back to key-value storage. For `ActivityController` that means the retained legacy blob becomes the source of truth again, silently missing everything written since the migration. Treat version bumps as unrollbackable: ship one on its own, not bundled with unrelated risk, and only when something actually reads the new structure.
+- `AccountOpsPersistence` and `PhishingPersistence` are the two worked examples of the whole wiring. Copy the phishing one unless your data needs a partial startup read — it is the smaller shape.
+
```
PetromirDev, line 136: Can be simplified to:
- A `StorageController` migration cannot reach a key that already moved to IDB - it would rewrite the dead legacy copy. Use an `idbDatabase.ts` migration handler instead.
- Keep the legacy key as a fallback while the IDB path is new, and record which backend is active (`activityStorageBackend`, written by `#recordActiveBackend`).
- That copy is frozen at migration time. It is a recovery floor, never a source of truth, and decide up front when it gets deleted.
- `isEmpty()` cannot tell "never migrated" from "wiped, then partially repopulated". Accepted, not worth fixing.
- Bulk writes must drop malformed legacy rows with a warning, not throw mid-batch. A partial commit makes `isEmpty()` false and disables the retry forever.
- A `dbVersion` bump is one-way - an older build gets `VersionError` and falls back to key-value. Ship one alone, and land any store you already know you need before release.
- Copy `PhishingPersistence` unless you need a partial startup read, it is the smaller shape.

PetromirDev:
Some test case ideas:

Cap of 1000 per account/chain
- the cap is per chain, two chains of one account hold 1000 each

Counts with a partly loaded history
- an account with ops on two chains reports the sum
- the count stays at 1000 after an addAccountOp at the cap
- the count is right when the startup read loaded less than is stored, across two chains

Internal + external totals

- external only, itemsTotal equals the external count
- mixed, itemsTotal equals internal plus external
- an external op duplicating an internal one is counted once
- external ops on a disabled chain are in neither items nor itemsTotal
- with an identifiedBy filter itemsTotal is just the matched items

Address casing
- external ops keyed lowercase show up for a checksummed filters.account, in items and in the total

sonytooo (changes requested):
PR review: https://github.com/AmbireTech/ambire-common/pull/2513/changes/3cea51a32ef56f3dcaa36888cda05d7d41d76ceb

PR #2517, opened 2026-07-02
Feature / Colibri - Extract Portfolio verification into its own controller

jordan-enev:
WIP:  Extract Portfolio verification into its own controller to reduce the complexity of the main PortfolioController.


Oxbobby:
I'm merging this and will fix issues in the main branch

PR #2518, opened 2026-07-02
Feature / support Chrome side panel overlay view and in-panel request handling

Maskln:
paired with: https://github.com/AmbireTech/ambire-app/pull/7448

## Summary

Adds controller-layer support for running the extension in Chrome’s side panel instead of a separate popup request window.

- Introduce `side-panel` as an extension overlay view type alongside `popup`, and update `UiController` to treat popup/side-panel as a single overlay slot (swap type in place when the user toggles modes).
- Update `RequestsController` to detect an open side panel and skip opening/focusing a request window; action requests are rendered in-place in the panel via `currentUserRequest`.
- Add `dismissActiveRequest()` with the same queue semantics as closing the request window (calls stay queued, non-calls are rejected per existing rules).
- Extract shared `#onActiveRequestDismissed()` / `#handleRequestWindowClose()` logic so popup close and side-panel dismiss behave consistently.
- Wire side-panel awareness through `dapps`, `transfer`, and `main` controllers where overlay view type matters.

## Test plan

- [ ] With side panel open, trigger a dapp connect / sign message / sign transaction request — no popup window opens; request becomes `currentUserRequest`.
- [ ] Dismiss an active request from the side panel — calls are queued with notification; other request types are rejected as before.
- [ ] Close the side panel while requests are pending — behavior matches popup overlay teardown.
- [ ] Toggle side panel mode off — extension falls back to popup request window flow.
- [ ] Regression: popup-only mode still opens, focuses, and closes request windows correctly.


src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -511,6 +511,19 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
     })
 
     this.#ui.uiEvent.on('removeView', (view: View) => {
+      if (view.type === 'side-panel') {
```
PetromirDev: isSidePanelView

src/controllers/transfer/transfer.ts
```diff
@@ -291,7 +291,8 @@ export class TransferController extends EventEmitter implements ITransferControl
     const isSameMode = this.isTopUp === nextIsTopUp
     const hasNoSearchParams = Object.keys(searchParams || {}).length === 0
 
-    const shouldKeepExistingForm = isFormInitialized && isSameMode && hasNoSearchParams
+    const shouldKeepExistingForm =
+      isFormInitialized && isSameMode && hasNoSearchParams && view.type !== 'side-panel'
```
PetromirDev: !isSidePanelView

src/controllers/requests/requests.ts
```diff
@@ -594,10 +594,24 @@ export class RequestsController extends EventEmitter implements IRequestsControl
     await this.closeRequestWindow()
   }
 
+  // When the wallet runs in the Chrome side panel and the panel is open, the action
+  // requests are rendered inside the panel itself (over the dashboard) instead of a
+  // separate request window. A side-panel view only exists in `views` while the panel
+  // is open, so its presence is the signal to skip the window entirely.
+  get #isSidePanelOpen() {
+    return this.#ui.views.some((view) => view.type === 'side-panel')
```
PetromirDev: isSidePanelView

PetromirDev (changes requested):

Oxbobby (approved):

sonytooo (changes requested):
https://github.com/user-attachments/assets/25b95e7c-ba4b-441d-92b6-69cef0c679cd

2 bugs:
* on reject or batch the SignAccountOp request should be removed or closed but should not close the panel. It should navigate to the Dashboard as a default route. The whole idea of a side panel is to not get closed automatically
* on consecutive open we should not focus the pending requests but see the dashboard with the pending banners and it is up to the user what to open

sonytooo (changes requested):
Image trending tokens detailed view footer broken styles

sonytooo (changes requested):
Image broken footer styles

sonytooo (changes requested):
Image 

2 issues:
* from side panel if i see a button like address book manage or some other that requires opening an ambire settings tab the side panel automatically navigates to dashboard. I think the the side panel should remain on the same route
* when all accounts are removed the side panel is just a while square so talk it out with Ivan and Miro how to solve this and what to show in there

sonytooo (changes requested):
https://github.com/user-attachments/assets/25bc5143-4078-45d1-b3f4-cc9a75922c8a

bug while the side panel is opened. Actually on every other onboarding flow there is an issue that prevents me from continuing to the account picker or the the account personalize screen

Image

sonytooo (changes requested):
and while the side panel is opened and i add a view-only account the dashboard in a side panel does not work at all stuck on spinner

Screenshot 2026-08-05 at 17 16 05

and after restart I see a sign message request from another account already deleted (the only account that is added is the view-only one)
Screenshot 2026-08-05 at 17 19 13

src/controllers/requests/requests.ts
```diff
@@ -612,10 +612,71 @@ export class RequestsController extends EventEmitter implements IRequestsControl
...
+  // persisted mode flag (survives brief view/port races) and fall back to an
+  // already-open side-panel view.
+  get #isSidePanelOpen() {
+    return this.#ui.views.some((view) => isSidePanelView(view))
+  }
+
+  async #shouldHandleRequestsInSidePanel() {
+    if (this.#isSidePanelOpen) return true
+
+    try {
```
sonytooo: try catch here is redundant when we have a default value to return

sonytooo (approved):

PR #2522, opened 2026-07-03
add: GNS (.gwei) names support

lucadonnoh:
## What

Adds support for resolving **GNS (Gwei Name Service) `.gwei` names**, following the exact pattern Namoshi established: same ENS code path, different universal resolver contract.

GNS is an ownerless, ENS-compatible naming system on Ethereum ([gwei.domains](https://gwei.domains), [source](https://github.com/donnoh-eth/gwei-names)). Names are ERC-721 tokens keyed by EIP-137 namehash, registration fees are burned, and the contracts have no owner or admin.

## How it plugs in

GNS ships a universal resolver implementing the same interface viem and `EnsGetter` already speak (`resolveWithGateways` / `reverseWithGateways`):

- [`GnsUniversalResolver`](https://etherscan.io/address/0xD658131FFB6D732335d37f199374289F1b31564F): `0xD658131FFB6D732335d37f199374289F1b31564F`, verified, stateless, no owner.
- Deployed at the **same address on Ethereum mainnet and Sepolia**, so `defaultNetworksMode: 'testnet'` works with no extra wiring.
- `.gwei` names live on the Ethereum provider already used for ENS. No new network, no new RPC provider.

## Changes

- `services/ensDomains`: `GNS_UNIVERSAL_RESOLVER` const, `getIsGweiDomain`, `isGweiDomain` option on `resolveENSDomain` / `reverseLookupEns` / `getEnsAvatar`.
- `controllers/domains`: `.gwei` branch in `resolveDomain`, a third parallel lookup in `#reverseLookup` (batched through the existing deployless `EnsGetter`, which takes the resolver address as a parameter, so no contract changes), `gwei` field in the domains cache.
- `interfaces/domains`: `gwei` field on `Domains` entries, `'gwei'` in `resolvedAddressType`.
- Tests: four new cases in `domains.test.ts` resolving a real mainnet name (`donnoh.gwei`) in both directions, plus the unregistered-name and no-name cases. Existing spy-count tests updated because each reverse refresh now performs two Ethereum lookups (ENS and GNS).

## Behavior notes

- The GNS resolver signals "not found" with empty return values rather than reverts, so viem maps missing/expired names to `null` and `EnsGetter` treats them as `hasName: false`. No new error handling needed anywhere.
- Reverse records are forward-verified on-chain by GNS itself (a name that no longer resolves to the queried address is not returned).
- GNS is fully on-chain: no CCIP-read, gateways are ignored.

## Verification

- `npm run jest -- src/controllers/domains/domains.test.ts`: 24/24 (the new tests hit the live mainnet deployment).
- `npm run jest -- src/services/ensDomains/ensDomains.test.ts`: 4/4.
- `npm run type:check`: no new errors (210 pre-existing on `v2`, identical with and without this change).
- `npx eslint` on the touched files: clean.

A companion PR adds the UI side (icon, input validation labels, account search) in the extension repo.

### DIFF
--- a/src/controllers/AGENTS.md
+++ b/src/controllers/AGENTS.md
@@ -116,7 +116,7 @@ ALWAYS update this list when creating a new controller, and provide a one-senten
 - **ContractNamesController** – Resolves human-readable names for smart-contract addresses via the relayer.
 - **DappsController** – Manages dApp connections, sessions, verification status, and the dApp catalog.
 - **DebugController** – Toggles per-controller debug logging at runtime (developer tool); persists toggles and hydrates the `debugLogger` module.
-- **DomainsController** – Resolves and caches ENS and Namoshi names (and avatars) for addresses.
+- **DomainsController** – Resolves and caches ENS, Namoshi and GNS names (and avatars) for addresses.
 - **EmailVaultController** – Handles email-based recovery, magic-link flows, and vault secret management.
 - **FeatureFlagsController** – Toggles application features at runtime for roll-outs and A/B testing.
 - **EstimationController** – Estimates gas, fees, and payment options for smart-account transactions.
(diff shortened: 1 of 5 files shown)


PetromirDev:
Implemented in https://github.com/AmbireTech/ambire-common/pull/2532

PR #2523, opened 2026-07-06
Fix: safe accounts status bridges were always in a pending state

Oxbobby:
This was because, with safe accounts, we don't have a "one click broadcast" in the swap & bridge mode


superKalo (approved):
I'd appreciate if @sonytooo takes a look too, he knows this logic inside out.

sonytooo (approved):

PR #2524, opened 2026-07-06
Feature/safe sponsorship

Oxbobby:
## Summary

Enables Gas Tank sponsorship for Safe accounts by adding Safe-specific user operation support, paymaster signature handling, bundler estimation, fee calculation, and deployment/config support for the Safe sender flow.

## Changes

- Add Safe user operation flow through `SAFE_SENDER`.
- Generate Safe-specific userOp calldata, keeping the Gas Tank fee call outside the Safe transaction.
- Support paymaster-returned Safe signatures.
- Add Safe bundler state override using `SafeNoSignatureValidation`.
- Allow Safe accounts to broadcast through the bundler when paying from Gas Tank.
- Adjust Safe gas estimation with bundler overhead.
- Update fee calculation with Safe-specific overheads.
- Disable unsupported Safe/native account-paid fee options while keeping them visible.
- Restrict Gas Tank deposit eligibility for unsupported tokens/networks.
- Add Safe sender deployment script and expanded Hardhat network config.
- Add/update tests for Safe userOp calldata and Safe fee overhead.


Oxbobby:
I reviewed the AI comments and they are not applicable

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -2802,6 +2823,25 @@ export class SignAccountOpController
       const paymasterData = response as PaymasterSuccessReponse
       localOp.paymaster = paymasterData.paymaster
       localOp.paymasterData = paymasterData.paymasterData
+
+      // if it's a safe account, add the SAFE_SIGNER signature
+      if (this.account.safeCreation) {
+        if (!paymasterData.signature) {
+          const error = 'Gas tank is currently inavailable'
```
PetromirDev: unavailable

src/libs/erc7677/types.ts
```diff
@@ -30,6 +30,8 @@ export type PaymasterEstimationData = {
 export interface PaymasterData {
   paymaster: Hex
   paymasterData: Hex
+  // safe userOps come with a SAFE_SENDER signature
```
PetromirDev: make it public

src/libs/account/Safe.ts
```diff
@@ -155,14 +188,12 @@ export class Safe extends BaseAccount {
   }
 
   getNonceId(): string {
-    // the Safe will move only its own smart account nonce as we don't have 4337
+    // the Safe will move only its own smart account nonce
     return `${this.accountState.nonce.toString()}`
```
PetromirDev, line 196: Should we change this to:
`${this.accountState.nonce.toString()}-${this.accountState.erc4337Nonce.toString()}`

src/libs/fees/fees.ts
```diff
@@ -1,11 +1,16 @@
...
 
-export function increaseFee(amount: bigint, broadcaster: FeeBroadcaster = 'relayer'): bigint {
-  if (broadcaster === 'paymaster') return amount + amount / 10n
+export function increaseFee(amount: bigint, isAccountSafe: boolean, network: FeeNetwork): bigint {
+  if (isAccountSafe) {
+    if (network.chainId === 1n) return amount + amount / 4n
 
-  return amount + amount / 20n
+    return amount + amount / 2n
+  }
```
PetromirDev, line 11: Isn't this a 25%/50% increase?

src/libs/safe/safe.ts
```diff
@@ -204,7 +204,7 @@ export function getSafeBroadcastTxn(
       safeTxn.gasPrice,
       safeTxn.gasToken,
       safeTxn.refundReceiver,
-      op.signature
+      op.signature || getSigForCalculations()
```
PetromirDev: The fallback will not work if the signature is `0x`, is that okay?

Image

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -1070,21 +1082,12 @@ export class SignAccountOpController
...
-        } else if (payOptionsPaidByEOA.length) {
-          const selected = this.#getDefaultFeeOption(payOptionsPaidByEOA)!
-          this.feeTokenResult = selected.token
-          this.#paidBy = selected.paidBy
-        }
+        const selected = this.#getDefaultFeeOption(
+          payOptionsPaidByUsOrGasTank,
+          payOptionsPaidByEOA
+        )!
+        this.feeTokenResult = selected.token
```
PetromirDev: selected could be undefined, which will make selected.token throw an error

PetromirDev (changes requested):

src/libs/account/Safe.ts
Oxbobby: I didn't change it on purpose as `accountState.erc4337Nonce` would be the SAFE_SENDER nonce. It's a bit strange to include that as the account nonce.  
Also, `accountState.nonce` always increments upon a successful transaction. So all the logic that's depending on this will work as intended

src/libs/fees/fees.ts
Oxbobby: Yes. The gas tank for safe accounts is going to be a premium feature. The percentages are correct - 25% on Ethereum, 50% on L2.  
While percentage wise it's a big increase, I broadcast over 30 txns in total across networks (only a few on Ethereum) and I managed to do so with 20 cents or so. When there aren't gas spikes, users don't notice the change as they are paying below a cent. When there are, they will need to pay the fee if they want to use the gas tank.  
We also stopped the gas tank cashback.

src/libs/safe/safe.ts
Oxbobby: It would, but it's a good improvement regardless. I'm changing it to:
```
op.signature && op.signature !== '0x' ? op.signature : getSigForCalculations()
```

src/controllers/signAccountOp/signAccountOp.ts
Oxbobby: Changing to:
```
const selected = this.#getDefaultFeeOption(payOptionsPaidByUsOrGasTank, payOptionsPaidByEOA)
if (selected) {
  this.feeTokenResult = selected.token
  this.#paidBy = selected.paidBy
} else {
  console.error(
    'Failed to set a default selected option. Perhaps native is not always returned for the account?'
  )
}
```

PR #2530, opened 2026-07-09
Feature/cowswap humanization module

Oxbobby:
Add message and transaction modules for Cowswap to improve humanization

Screenshot 2026-07-09 at 17 01 48

### DIFF
--- a/src/libs/humanizer/messageModules/cowSwapModule.ts
+++ b/src/libs/humanizer/messageModules/cowSwapModule.ts
@@ -1,4 +1,4 @@
-import { isAddress, zeroAddress } from 'viem'
+import { isAddress, isHex, zeroAddress } from 'viem'
 
 import { Message } from '../../../interfaces/userRequest'
 import { HumanizerTypedMessageModule, HumanizerVisualization } from '../interfaces'
@@ -6,25 +6,24 @@ import { getAction, getAddressVisualization, getDeadline, getLabel, getToken } f
 
 const COW_SWAP_SETTLEMENT_ADDRESS = '0x9008d19f58aabd9ed0d60971565aa8510560ab41'
 
-const toBigInt = (value: unknown): bigint | null => {
-  try {
-    if (typeof value === 'bigint' || typeof value === 'number' || typeof value === 'string') {
-      return BigInt(value)
-    }
-
-    if (
-      value &&
-      typeof value === 'object' &&
-      '$bigint' in value &&
-      typeof value.$bigint === 'string'
-    ) {
-      return BigInt(value.$bigint)
-    }
-  } catch {
-    return null
+const getOrderUidVisualization = (
+  orderUid: string,
+  accountAddr: string
+): HumanizerVisualization[] => {
+  if (!isHex(orderUid) || orderUid.length !== 114) {
+    return [getLabel(`with order ID ${orderUid.slice(0, 8)}...${orderUid.slice(-6)}`)]
   }
 
-  return null
+  const owner = `0x${orderUid.slice(66, 106)}`
+  const validTo = BigInt(`0x${orderUid.slice(106)}`)
+
+  return [
+    getLabel(`with order ID ${orderUid.slice(0, 8)}...${orderUid.slice(-6)}`),
+    ...(owner.toLowerCase() !== accountAddr.toLowerCase()
+      ? [getLabel('owned by'), getAddressVisualization(owner)]
+      : []),
+    getDeadline(validTo)
+  ]
 }
 
 export const cowSwapModule: HumanizerTypedMessageModule = (message: Message) => {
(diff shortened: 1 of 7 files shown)


src/libs/humanizer/messageModules/cowSwapModule.ts
```diff
@@ -0,0 +1,105 @@
...
+      typeof value.$bigint === 'string'
+    ) {
+      return BigInt(value.$bigint)
+    }
+  } catch {
+    return null
+  }
+
+  return null
+}
```
JIOjosBG: i am not sure you need this function, when will you actually need this instead of BigInt? Do we really get values like `{ $bigint: ... }`?

src/libs/humanizer/messageModules/cowSwapModule.ts
```diff
@@ -0,0 +1,105 @@
...
+    validTo: rawValidTo,
+    kind,
+    receiver,
+    feeAmount: rawFeeAmount
+  } = tm.message
+  const sellAmount = toBigInt(rawSellAmount)
+  const buyAmount = toBigInt(rawBuyAmount)
+  const validTo = toBigInt(rawValidTo)
+  const feeAmount = toBigInt(rawFeeAmount) || 0n
+  const chainId = toBigInt(tm.domain.chainId) || message.chainId
```
JIOjosBG: if the function is not needed you can replace those with BigInt

src/libs/humanizer/messageModules/cowSwapModule.ts
```diff
@@ -0,0 +1,105 @@
...
+        ]
+
+  if (feeAmount > 0n) {
+    fullVisualization.push(getLabel('including fee'), getToken(sellToken, feeAmount, chainId))
+  }
+
+  if (
+    typeof receiver === 'string' &&
+    isAddress(receiver) &&
+    receiver.toLowerCase() !== zeroAddress &&
```
JIOjosBG, line 110: Are you sure that when that value is ZeroAddress it is ignored and it does not burn the output tokens?

src/libs/humanizer/modules/CowSwap/cowSwapModule.ts
```diff
@@ -0,0 +1,162 @@
...
+      return [getAction('Swap'), ...getOrderVisualization(accountOp, tokens, order)]
+    },
+    [toFunctionSelector(settleAbi[0])]: (call) => {
+      const { args } = decodeFunctionData({ abi: settleAbi, data: call.data })
+      const [tokens, , trades] = args as unknown as [
+        readonly string[],
+        readonly bigint[],
+        readonly CowSwapOrder[],
+        unknown
+      ]
```
JIOjosBG: do not cast, viem is able to detect the type generally, 

in this case it might no be able becuase you have defined the abi above with string and not with json. Use json for abis that result in decoded arguments with unknown types

I had the same case here: https://github.com/ambireTech/ambire-common/blob/v2/src/libs/humanizer/modules/OpenSea/index.ts#L23-L131

src/libs/humanizer/modules/CowSwap/cowSwapModule.ts
```diff
@@ -0,0 +1,162 @@
...
+const getOrderUidDeadline = (orderUid: string): HumanizerVisualization | null => {
+  if (!isHex(orderUid) || orderUid.length !== 114) return null
+
+  return getDeadline(BigInt(`0x${orderUid.slice(-8)}`))
+}
+
+const getOrderUidVisualization = (orderUid: string): HumanizerVisualization[] =>
+  [getLabel(`with order ID ${getShortOrderUid(orderUid)}`), getOrderUidDeadline(orderUid)].filter(
+    Boolean
+  ) as HumanizerVisualization[]
```
JIOjosBG: ~remove the cast, by altering the filter.
Try this, you might need to change it a bit
`.filter((v: HumanizerVisualization | null): v is HumanizerVisualization => Boolean(v))`~

JIOjosBG: actually you can merge those 3 funuction `getShortOrderUid` `getOrderUidDeadline` and `getOrderUidVisualization` and have optional return
```
if(deadline data) return [label,deadline]
else return [label]
```

JIOjosBG (changes requested):

src/libs/humanizer/modules/CowSwap/cowSwapModule.ts
JIOjosBG: i resolved and committed the change in this PR

src/libs/humanizer/modules/CowSwap/cowSwapModule.ts
JIOjosBG: I resolved that, you can check it out, i looks just a little uglier but avoids the type cast and allows viem to detect the type

src/libs/humanizer/messageModules/cowSwapModule.ts
Oxbobby: removing

src/libs/humanizer/messageModules/cowSwapModule.ts
Oxbobby: Yes. 1inch also had this case with erc-7730 humanization where their receiver was zero address. It's hell confusing but it means you are the receiver

src/libs/humanizer/modules/CowSwap/cowSwapModule.ts
Oxbobby: nice 👍

JIOjosBG (approved):

PR #2532, opened 2026-07-09
Refactor/ Domains controller and implement gwei names

PetromirDev:
Reworks domain resolution so the wallet can resolve names from more than just ENS. Instead of ENS/Namoshi being hardcoded, each name service is now a small self-contained resolver, and adding a new ENS-compatible service is a few lines. Also adds GNS (.gwei) on top of this.

## Changes:

- Introduced a pluggable resolver layer - ENS-compatible services (ENS, Namoshi, GNS) share one resolution path and only differ by their universal resolver, chain and TLD
- Added GNS (.gwei) resolution
- Added GNS verification with Colibri
- Reworked the domains cache to store resolved names, avatar and expiry per service rather than ENS-specific fields
- Fixed: Forward resolution was being cached (introduced in https://github.com/AmbireTech/ambire-common/commit/9bc4f888351f13ce4765f4d32885ae94b092e212)


src/services/nameResolvers/resolvers/GnsResolver.ts
```diff
@@ -0,0 +1,23 @@
...
+      id: 'gns',
+      label: 'GNS',
+      universalResolver: GNS_UNIVERSAL_RESOLVER,
+      chainId: ETHEREUM_CHAIN_ID,
+      featureFlag: 'gnsDomains'
+    })
+  }
+
+  matches(domain: string): boolean {
+    return domain.endsWith('.gwei')
```
gergana95, line 21: This resolves donnoh.gwei but not DONNOH.GWEI. This may be correct, just checkin

gergana95 (approved):
Tested and works great, did not find any issues. Great work 🙌

PR #2534, opened 2026-07-10
Feature: opt out of erc-4337

Oxbobby:
## Summary

- Replaces custom gas price sampling from full block transaction data with viem fee estimation.
- Uses `eth_feeHistory` for EIP-1559 priority fee guidance, with fallback to viem fee estimation if fee history fails.
- Adds the `erc4337` feature flag, enabled by default.
- Threads the ERC-4337 flag through account creation, estimation, gas price fetching, paymaster setup, transfers, swaps, and dapp requests.
- Disables ERC-4337-dependent behavior when the flag is off:
  - bundler gas price fetching
  - bundler estimation support for EOA7702, V2, and Safe accounts
  - V2 internal/gas tank fee options
  - swap sponsorship
  - paymaster service
  - EOA transaction recording used for Gas Tank
- Adds `enableErc4337AndReestimate()` so the UI can re-enable smart account fee options and refresh estimation.
- Adds/updates tests for gas price fetching, account fee options, and swap sponsorship gating.


src/consts/featureFlags.ts
```diff
@@ -5,6 +5,13 @@ export interface FeatureFlags {
   testnetMode: boolean
   tokenAndDefiAutoDiscovery: boolean
   apiForFunctionSelectors: boolean
+  /**
+   * Allow the user to opt out of erc4337 which will automatically
+   * disable paying gas in different tokens & gas tank.
+   * For Ambire v2 accounts, it will also disalbe ETH payments (the user
```
PetromirDev: typo - disable

src/libs/paymaster/paymaster.ts
```diff
@@ -82,14 +82,22 @@ export class Paymaster extends AbstractPaymaster {
...
   // this is a temporary solution where the live relayer doesn't have
   // a chain id paymaster route open yet as it's not merged
   ambirePaymasterUrl: string | undefined
 
-  constructor(relayerUrl: string, fetch: Fetch, errorCallback: Function) {
+  constructor(
+    relayerUrl: string,
+    fetch: Fetch,
+    errorCallback: Function,
+    getIsErc4337Enabled: () => boolean = () => true
```
PetromirDev: Even though this class is only used in one place, I'd avoid defaulting `getIsErc4337Enabled` to true. If we forget it's set that way, we could end up making requests we didn't intend to.

src/libs/paymaster/paymaster.ts
```diff
@@ -104,6 +112,11 @@ export class Paymaster extends AbstractPaymaster {
     this.provider = provider
     this.ambirePaymasterUrl = `/v2/paymaster/${this.network.chainId}/request`
 
+    if (!this.#getIsErc4337Enabled()) {
+      this.type = 'None'
+      return
+    }
```
PetromirDev: Actually, are the changes in this file even needed, considering the short circuit in estimateBundler:138

src/services/paymaster/PaymasterFactory.ts
```diff
@@ -20,10 +20,18 @@ export class PaymasterFactory {
...
   errorCallback: Function | undefined = undefined
 
-  init(relayerUrl: string, fetch: Fetch, errorCallback: Function) {
+  getIsErc4337Enabled: () => boolean = () => true
+
+  init(
+    relayerUrl: string,
+    fetch: Fetch,
+    errorCallback: Function,
+    getIsErc4337Enabled: () => boolean = () => true
```
PetromirDev: same comment regarding the default

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -425,12 +430,18 @@ export class SignAccountOpController
...
     this.feeTokenPreference = this.#signAccountOpPreference.feeTokenPreference
     this.#externalSignerControllers = externalSignerControllers
     this.account = account
     const accountState = accounts.accountStates[account.addr]![network.chainId.toString()]! // ! is safe as otherwise, nothing will work
-    this.baseAccount = getBaseAccount(account, accountState, network)
+    this.baseAccount = getBaseAccount(
+      account,
+      accountState,
+      network,
+      this.#featureFlags.isFeatureEnabled('erc4337')
```
PetromirDev, line 449: .bind(this) or use an anonymous function

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -473,14 +484,21 @@ export class SignAccountOpController
...
+    this.gasPrice = new GasPriceController(
+      network,
+      provider,
+      this.baseAccount,
+      () => ({
+        estimation: this.estimation,
+        readyToSign: this.readyToSign,
+        stopRefetching: this.#stopRefetching
+      }),
+      () => this.#featureFlags.isFeatureEnabled('erc4337')
```
PetromirDev: do we need to pass this and can't we rely on this.#baseAccount.supportsBundlerEstimation()?

src/libs/account/V2.ts
```diff
@@ -45,14 +45,18 @@ export class V2 extends BaseAccount {
...
   }
 
   getAvailableFeeOptions(
     estimation: FullEstimationSummary,
     feePaymentOptions: FeePaymentOption[],
     op: AccountOp
   ): FeePaymentOption[] {
+    if (!this.isErc4337Enabled) {
+      return feePaymentOptions.filter((opt) => opt.paidBy !== this.account.addr)
+    }
```
PetromirDev, line 61: We are not checking if opt.availableAmount > 0n

src/libs/gasPrice/gasPrice.ts
```diff
@@ -36,285 +54,141 @@ export interface Gas1559Recommendation {
...
 
-async function fetchBlock(
-  provider: Provider,
-  network: Network,
-  blockTag: string | number
-): Promise<GasPriceBlock | null> {
-  if (network.chainId === 2741n) return fetchRawBlock(provider, blockTag)
-  return await provider.getBlock(blockTag, true)
+async function getLegacyViemFees(client: PublicClient) {
+  return client.estimateFeesPerGas({ chain: null, type: 'legacy' })
```
PetromirDev, line 106: Just fyi that this will make the estimatedBaseFee 20% higher because of baseFeeMultiplier https://viem.sh/docs/chains/fees#feesbasefeemultiplier

src/libs/gasPrice/gasPrice.ts
```diff
@@ -1,30 +1,48 @@
...
+    baseFeeAddBps: 200n,
+    priorityFeeMinAddBps: 2500n,
+    priorityFeeMaxAddBps: 5000n
+  },
+  {
+    name: 'ape',
+    baseFeeAddBps: 300n,
+    priorityFeeMinAddBps: 5000n,
+    priorityFeeMaxAddBps: 10000n
+  }
```
PetromirDev, line 40: baseFeeAddBps went from 0/500/1000/1500 to 0/100/200/300, so the base fee buffers are now 0/1/2/3%. Since gasPriceToBundlerFormat sums base fee and tip, slow broadcasts at exactly the predicted next block base fee with nothing on top.

src/libs/gasPrice/gasPrice.ts
```diff
@@ -36,285 +54,141 @@ export interface Gas1559Recommendation {
...
-  return lastBlock
+async function getLegacyGasPriceRecommendations(
+  client: PublicClient,
+  network: Network
+): Promise<GasPriceRecommendation[]> {
+  const { gasPrice } = await getLegacyViemFees(client)
+  const minGasPrice = increaseByPercent(
+    gasPrice > MIN_GAS_PRICE ? gasPrice : MIN_GAS_PRICE,
+    network.feeOptions.feeIncrease
+  )
```
PetromirDev: feeIncrease wasn't read anywhere before this branch. Combined with viem's 1.2 multiplier, Polygon slow becomes eth_gasPrice * 1.32, and all four speeds sit within 3% of each other where they used to span the last block's gas prices. Intentional?

PetromirDev (changes requested):

src/libs/paymaster/paymaster.ts
Oxbobby: We don't need them. Nice catch, I will delete them

src/controllers/signAccountOp/signAccountOp.ts
Oxbobby: we can rely on it, yes, nice catch

src/controllers/signAccountOp/signAccountOp.ts
Oxbobby: I don't think we need it here, no?

src/libs/gasPrice/gasPrice.ts
Oxbobby: With nothing on top from our logic, yes. There's the 20% fee increase that comes from the viem calculations that you discovered

src/libs/gasPrice/gasPrice.ts
Oxbobby: Nice 👍

src/libs/gasPrice/gasPrice.ts
Oxbobby: Let's scrap it. It's true that `feeIncrease` is from the old days where we didn't make a good job of this.
Edit: deleted all mentions for `feeIncrease` in the project

PR #2540, opened 2026-07-14
Fix / Sign message replacement race

sonytooo:
Resolves: https://github.com/AmbireTech/ambire-app/issues/7496

**ISSUE:**
A same-kind/same-account request could call SignMessageController.init() while an approved signing operation was awaiting getSigner(), swapping the installed message under the running op. The post-await validity check only verified the controller was initialized and held some message, so the stale op resumed, signed the replacement payload, and resolved it under the previous approval.

**FIX:**
Add a #signingGeneration counter bumped on every init()/reset(). #sign() captures it before its first await and re-checks after each async step; on a mismatch the operation aborts instead of signing the replacement.

Add a regression test covering the getSigner() window.


superKalo (approved):

superKalo:
Please point it towards the release branch (v2.103) 🙏

superKalo:
Note: because this brought everything from v2 branch, I had to reset the release/v2.103 branch and cherry-pick the commit from this branch in release/v2.103 branch in https://github.com/AmbireTech/ambire-common/pull/2540/changes/2d84e0d30c890403b8e58bd21e526956ad45ea78

PR #2541, opened 2026-07-14
Fix / Reject signing calls appended after the user reviewed the op

sonytooo:
Resolves: https://github.com/AmbireTech/ambire-app/issues/7491

**ISSUE:** 
A malicious dApp can append a call to a calls request after the user
has reviewed it but before the Sign click is processed. All calls for the same
account+chain share one mutable SignAccountOpController whose calls array stays
mutable until signing starts, and the Sign click carries only the stable
fromRequestId — so a race-winning append gets signed and broadcast unreviewed

**FIX:** 
Bind the click to the accountOp the user actually saw. The UI now sends the
reviewed accountOp id (regenerated on every calls change) with the Sign dispatch,
and the background aborts signing when it no longer matches the live op, letting
the user re-review the updated calls.

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -1010,8 +1010,15 @@ export class MainController extends EventEmitter implements IMainController {
     if (op.meta?.swapTxn) this.swapAndBridge.removeActiveRoute(op.meta.swapTxn.activeRouteId)
   }
 
-  async handleSignAndBroadcastAccountOp(type: SignAccountOpType, fromRequestId: string | number) {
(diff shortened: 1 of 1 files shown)


superKalo (approved):

superKalo:
Note: because prev PR brought everything from v2 branch, I had to reset the release/v2.103 branch and cherry-pick the commit from this branch in release/v2.103 branch in https://github.com/AmbireTech/ambire-common/pull/2527/changes/91b754c2f00e51e3024682abe192e6775e9c5530

PR #2547, opened 2026-07-16
Feature / Safe accounts on mobile

sonytooo:
## Summary
Cross-device support for Safe (multisig) message co-signing: propagates the requesting dapp's identity to other owners, and refreshes pending Safe requests when the view regains focus.

## Changes
- **Dapp metadata across devices** — when the first owner signs a Safe message, the requesting dapp's `{name, url}` is stamped into the Safe Transaction Service `origin` field. On the co-signing device the message is rebuilt with that metadata (`dappName`/`dappUrl` on the request meta), so the signer sees which dapp requested the signature instead of an empty screen.
- **Pending requests refresh** — added a generic `viewFocus` UI event (`SET_VIEW_FOCUS`) so the UI only reports focus and the controllers decide when to refresh Safe pending requests.

## Notes
- `origin` is capped at 200 chars, so only name + url are carried; if it doesn't fit, origin is skipped so the message proposal never fails (the icon is resolved app-side from the dapp catalog).
- Reading is defensive — a non-JSON `origin` set by another wallet is treated as the name.
- Added unit tests for the origin build/parse helpers.

### DIFF
--- a/src/controllers/main/main.ts
+++ b/src/controllers/main/main.ts
@@ -210,6 +210,8 @@ export class MainController extends EventEmitter implements IMainController {
 
   #continuousUpdates: ContinuousUpdatesController | undefined
 
+  #fetchSafeTxnsPromise: Promise<void> | undefined
+
   safe: ISafeController
 
   get continuousUpdates() {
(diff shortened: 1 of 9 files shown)


src/controllers/main/main.ts
```diff
@@ -715,13 +717,17 @@ export class MainController extends EventEmitter implements IMainController {
...
-        this.fetchSafeTxns().catch((e) => e) // we catch the error inside
       }
     })
 
     this.ui.uiEvent.on('addView', async (view: View) => {
       if (view.type === 'popup') await this.onPopupOpen(view.id)
     })
+
+    this.ui.uiEvent.on('viewFocus', () => {
+      console.log('in')
```
PetromirDev: delete

src/libs/safe/safe.ts
```diff
@@ -165,13 +166,55 @@ export async function addMessage(
...
+    if (parsed && typeof parsed === 'object') {
+      const name = typeof parsed.name === 'string' ? parsed.name : undefined
+      const url = typeof parsed.url === 'string' ? parsed.url : undefined
+      return { name, url }
+    }
+  } catch {
+    // origin may be a plain, non-JSON string set by another wallet; treat it as the name
+    return { name: origin }
+  }
+  return {}
```
PetromirDev: Move these to helpers

PetromirDev (changes requested):

PR #2548, opened 2026-07-20
Add: implement eth_simulatev1

Oxbobby:
Add: implement eth_simulatev1 as a third backup option for token discovery should debug_traceCall and eth_createAccessList fail

### DIFF
--- a/src/controllers/signAccountOp/signAccountOp.ts
+++ b/src/controllers/signAccountOp/signAccountOp.ts
@@ -2210,9 +2210,6 @@ export class SignAccountOpController
         return
       }
 
-      console.log('shouldUseAccessList', shouldUseAccessList)
-      console.log('accessListFailed', accessListFailed)
-      console.log('debugTraceCallFailed', debugTraceCallFailed)
       if ((!shouldUseAccessList || accessListFailed) && debugTraceCallFailed) {
         console.log('Debug: using eth_simulateV1 for asset discovery')
         const { tokens, nfts } = await ethSimulateV1(
--- a/src/controllers/signAccountOp/signAccountOp.ts
+++ b/src/controllers/signAccountOp/signAccountOp.ts
@@ -129,6 +129,7 @@ import {
 import { isPermit2Interaction } from '../../libs/simulation/detectPermit2Interaction'
 import { getGasUsed } from '../../libs/singleton/singleton'
 import { createAccessListCall, getShouldUseAccessListCall } from '../../libs/tracer/accessListCall'
+import { ethSimulateV1 } from '../../libs/tracer/ethSimulatev1'
 import { UserOperation } from '../../libs/userOperation/types'
 import {
   getActivatorCall,
@@ -2150,6 +2151,7 @@ export class SignAccountOpController
       const stateOverride = getStateOverride(this.account, this.accountOp, state)
       const shouldUseAccessList = getShouldUseAccessListCall(this.account, !!stateOverride)
       let accessListFailed = false
+      let debugTraceCallFailed = false
 
       if (shouldUseAccessList) {
         console.log('Debug: using eth_createAccessList for asset discovery')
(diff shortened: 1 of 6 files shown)


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -2181,7 +2183,36 @@ export class SignAccountOpController
...
+          const { tokens, nfts } = await debugTraceCall(
+            this.baseAccount,
+            this.accountOp,
+            this.#network,
+            state,
+            stateOverride
+          )
+          erc20s = tokens
+          erc721s = nfts
+        } catch (e: any) {
```
PetromirDev, line 2193: Just thinking whether there is a way to optimize this. For example, if traceCall and accessList never work on robinhood, then we still have to call and await the first two methods, right? Not really sure how it would look like in terms of code, but it may be kind of cool to have the last successful method call, keyed by what this logic depends on. Probably something along the lines of `provider:accountType:isBatch`. Store the last successful and try that first, but still go through the rest in case it fails. I don't see downsides if we can make it work in terms of code.

PetromirDev (changes requested):

src/controllers/signAccountOp/signAccountOp.ts
Oxbobby: I agree and I was thinking to do this as a network setting. Like disable this and this for this network but enable it for another

Your solution sounds better, though. We can keep the last success in memory, without storage, and retry firstly from there depending on the account type

We should do this in a separate PR though. I will prepare an issue

PetromirDev (approved):

src/controllers/signAccountOp/signAccountOp.ts
superKalo: Sounds great indeed! Two things:

1) @Oxbobby wydt for having this logic with expiration (like 1 month) so that in 1 month the logic retries again the default path (debug_traceCall, eth_createAccessList and finally eth_simulatev1) because the RPC/chain might have added support for `debug_traceCall`?

2) Side note - this is causing Sentry noise on the fallback path (even without the changes in this PR). So each failed method now emits a `level: 'silent'` error even when a later fallback succeeds. On a network without `debug_traceCall` support - every discovery will emit 1–2 silent errors despite ending in `Done`.

  So we can maybe (in another PR) consider demoting the intermediate failures to a plain debug/console log and only calling `emitError` in the outer `catch` when all three methods fail - otherwise we'd get constant Sentry reports for which we should not care.

src/libs/tracer/ethSimulatev1.ts
```diff
@@ -0,0 +1,180 @@
...
+        stateOverrides: any
+        calls: { to?: string; value: string; data: string; from: string }[]
+      }[]
+      validation: false
+    },
+    'latest'
+  ]
+} | null {
+  const account = baseAcc.getAccount()
+  const shouldUseNativeBundle = op.calls.length > 1
```
superKalo: Minor: Please add a one-line comment stating the reason. I think it's because for multi-call ops this path only sets a `balance` override and ignores `overrideData` (the Safe priv / 7702 code override). The tests show this 👀

superKalo (approved):

src/controllers/signAccountOp/signAccountOp.ts
Oxbobby: 1. As you mentioned in slack, the in memory solution will handle that as we do releases quite often
2. Agreed, I will check and fix this

src/libs/tracer/ethSimulatev1.ts
Oxbobby: It should use the native bundle all the time, not only in certain scenarios. It was a bug - we did a state override on a safe account that wasn't a batch and discovery didn't work

PR #2550, opened 2026-07-20
Add: uniswap API for the robinhood chain; fix socket fee collector

Oxbobby:


superKalo (approved):

superKalo:
Note: with these changes, we now are always charging a fee (on any supported chain - even if new come in) across all swap & bridge service providers (for the cases where we charge a fee).

PS: For LiFi we don't pass a recipient address and rely on LiFi's per-chain integrator configuration instead.

PR #2551, opened 2026-07-20
Humanizer/ self-contained errors

JIOjosBG:
## Change 1
Before this fix - if a module throws an error for one call, all other calls are not parsed by that module
Now - each call and module might throw exception and that specific call will not be parsed by that specific module, every other module-call pair is parsed properly

## Change 2
Before - account ops with global `.safeTx` with operation=1 with non whitelisted multicall contract used tot get warning on their first call 
Now - no call-level warning is added if the whole accOp is with operation 1 in that case - there will be separate PR for that that will make the warning appear for the whole batch 

AFTER
image

BEFORE
image

## TODO
- [ ] test manually normal behavior and safe type 1 txns


Oxbobby (approved):
All good here

PR #2553, opened 2026-07-20
Config / Release v2.103.4

superKalo:
See https://github.com/AmbireTech/ambire-common/releases/tag/v2.103.4


superKalo:
Note: 3 failing tests:

- 2 because of the changes in https://github.com/AmbireTech/ambire-common/pull/2548 (just the related the unit tests needs to be adjusted after the changes)
- 1 because of changed Relayer res

PR #2555, opened 2026-07-21
Feature: token discovery optimization

Oxbobby:
Add: move the 3 methods for token discovery (eth_createAccessList, debug_traceCall, eth_simulateV1) to their own lib and prioritize the successful one

Basically, the comment here covers all the changes:
```
/**
 * Goal: discover tokens that are interacting with the account during a txn
 * To do this, we use 3 methods:
 * - eth_createAccessList
 * - debug_traceCall
 * - eth_simulateV1
 * The end product of each method is the same. However, different methods
 * are supported on different chains depending on the chain configs and RPC.
 * Also, some accounts require state overrides which may not be supported
 * on the specific chain. That's why we have 3 methods and rely on 1
 * having the exact support we need. Each method is called one by one until
 * we hit a success or all methods have failed.
 * We also have a memory cache. It's simple: record the method that succeeded
 * for chain-account. The next time discoverTxnTokens is called, it will start
 * from the method that was successful.
 */
export async function discoverTxnTokens
```

### DIFF
--- a/src/controllers/signAccountOp/discoverTxnTokens.ts
+++ b/src/controllers/signAccountOp/discoverTxnTokens.ts
@@ -1,0 +1,103 @@
+import { Account, AccountOnchainState } from '../../interfaces/account'
+import { Network } from '../../interfaces/network'
+import { BaseAccount } from '../../libs/account/BaseAccount'
+import { AccountOp } from '../../libs/accountOp/accountOp'
+import { createAccessListCall, getShouldUseAccessListCall } from '../../libs/tracer/accessListCall'
+import { debugTraceCall, getStateOverride } from '../../libs/tracer/debugTraceCall'
+import { ethSimulateV1 } from '../../libs/tracer/ethSimulatev1'
+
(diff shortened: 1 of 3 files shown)


PetromirDev (approved):
Dope 🥇

src/controllers/signAccountOp/discoverTxnTokens.ts
```diff
@@ -0,0 +1,103 @@
...
+  const orderedMethods =
+    cachedMethod && availableMethods.includes(cachedMethod)
+      ? [cachedMethod, ...availableMethods.filter((method) => method !== cachedMethod)]
+      : availableMethods
+
+  let discoveredAssets: DiscoveredAssets | null = null
+  let lastError: unknown
+
+  for (const method of orderedMethods) {
+    console.log(`Debug: using ${method} for asset discovery`)
```
superKalo, line 84: Minor, I know this is unrelated (because it caries over logs already existing), but @PetromirDev wdyt for implementing debug logs via your newer concept for this flow?

superKalo (approved):

PR #2556, opened 2026-07-21
Clean up the hanging requests after denying an account switch

Oxbobby:
We have this nasty bug:
1. The currently selected account is not the one connected to the dapp
2. We try to do a dapp request
3. An account change request comes
4. We deny it
5. When we do that and we after select the account connected to the dapp in the extension, a simulation persists for the request we were about to make

Video of the bug:

https://github.com/user-attachments/assets/927a6f42-a272-40b9-a25e-9b3b66a28c99

Fixing it by cleaning up after denying the account switch request

### DIFF
--- a/src/controllers/requests/requests.ts
+++ b/src/controllers/requests/requests.ts
@@ -924,7 +924,10 @@ export class RequestsController extends EventEmitter implements IRequestsControl
         )
 
         requestsToAddOrRemove.forEach((r) => {
-          this.userRequestsWaitingAccountSwitch.splice(this.userRequests.indexOf(r), 1)
+          this.userRequestsWaitingAccountSwitch.splice(
+            this.userRequestsWaitingAccountSwitch.indexOf(r),
+            1
+          )
 
           if (
             r.kind === 'typedMessage' &&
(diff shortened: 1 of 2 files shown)


src/controllers/requests/requests.ts
```diff
@@ -1001,16 +1004,35 @@ export class RequestsController extends EventEmitter implements IRequestsControl
...
+      callsUserRequestsToReject.map((r) =>
+        this.#portfolio.overrideSimulationResults(r.signAccountOp.accountOp)
+      )
+    )
+
+    waitingUserRequestsToReject.forEach((r) => {
+      if (r.kind === 'calls') r.signAccountOp.destroy()
+    })
+    this.userRequestsWaitingAccountSwitch = this.userRequestsWaitingAccountSwitch.filter(
+      (r) => !waitingUserRequestsToReject.includes(r)
```
PetromirDev, line 1034: Should we do anything about safe txns, not really sure how they are broadcasted to the API, but we do this in the other reject case:
    // reject all Safe txns so they do not appear by accident again
    if (safeRejectIds.length) await this.#safe.rejectTxnId(safeRejectIds)
    if (safeResolveIds.length) await this.#safe.resolveTxnId(safeResolveIds)

PetromirDev (approved):
Works nicely, just check the comment below

src/controllers/requests/requests.ts
Oxbobby: These are for already signed transactions. The purpose is to reject them so they do not get initiated from the API again.  
In this case, it's a new request

PR #2557, opened 2026-07-21
Config agents/ rule for user-facing messages

JIOjosBG:
Messages must be phrased in plain language


PetromirDev (approved):
I'm a bit worried about blacklisting:
"RPC", "nonce", "gas limit", "Paymaster", "Bundler", "EOA", "Smart Account", "delegatecall", "calldata", "simulation failed"

but it's better to have AI write simpler messages that can then be made more detailed by us (e.g., to add technical context if needed)

wdyt, @superKalo

JIOjosBG:
I will merge and if Kalo has feedback we can update as we go

I agree there are a lot of cases where the terminology you listed above will be better to be in the message, but both we and AI are more likely to be on the technical side of the spectrum when it comes to messages

PR #2558, opened 2026-07-21
Humanizer/ delegations and verifiers + CoW swap

JIOjosBG:
## Adds humanzier for:
- safe add fallback handler
- safe add domain verifier for 712 messages (used in CoW TWAP)
- CoW TWAP orders

## To test
Go to the safe UI and with new account do a cowswap TWAP

image


Oxbobby:
Awesome

Oxbobby (changes requested):
I'm testing this and I see one place that we can improve. This bug is probably from another place but regardless, we can use this as an opportunity to fix it.  

When using the Safe account in Ambire itself (imported), it works exactly as the PR description says (I love it).  

We can improve on the Safe UI interactions. When signing/executing the Safe transaction in this mode, for some reason I don't see the final call that is the order itself. So I should be seeing:
* extend account functionality
* authorize custom access
* approve
* swap order

I'm seeing all but the swap order

Screenshot 2026-07-27 at 8 18 44

Screenshot 2026-07-27 at 8 18 28

JIOjosBG:
Discussed offline, i found the problem with the missing call and we can have it as hotfix instead of here

PR #2559, opened 2026-07-22
Add: a 15s timeout to safe API requests

Oxbobby:
When the safe API stalls, we don't cut it, and the user is left waiting in sign account op


src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -2946,11 +2948,14 @@ export class SignAccountOpController
     const isExternalSignerInvolved =
       this.accountOp.gasFeePayment.paidByKeyType !== 'internal' ||
       this.accountOp.signingKeyType !== 'internal'
+    const isCollectingSafeSignature =
+      !!this.account.safeCreation && (this.accountOp.signed?.length || 0) < this.threshold
```
superKalo, line 2848: For a threshold-1 single-key Safe (quick broadcast), signed is empty at this point so `0 < 1` would be `true`. That means we now skip `WaitingForPaymaster` for quick-broadcast Safes too, even though they don't collect extra signatures or hit the Safe API. Is that intended, or should the condition also require `threshold > 1`?

superKalo (approved):

src/controllers/signAccountOp/signAccountOp.ts
Oxbobby: The logic is a bit more complex than that. While it's true that for quick broadcast `isCollectingSafeSignature` would resolve to true, the logic later on changes the signing status to `SafeQuickBroadcastBundler` and invokes a second call to `sign()`. In the second call, `this.accountOp.signed?.length` would be equal to 1 and `isCollectingSafeSignature` would resolve to false, enabling the paymaster wait logic.

This is the correct behaviour as we're not really waiting for the paymaster until the Safe signing phase has concluded

PR #2561, opened 2026-07-22
Safe delegation warning

JIOjosBG:
Before that warning was attached to the first call in the humanizer but it is more correct to have it per-accountOp

Now it appears as a banner and forces a `Hold to sign`

## To test
- go in the code base and comment out the `multiSendAddr` from `allowedMulticallContracts` in `src/consts/safe.ts``
- execute a multicall (for example 2 ETH transfers)

image


Oxbobby:
This is interesting. What troubles me is that I think there are out there legit solutions that use non-whitelisted delegates and for people using them, this warning will appear all the time.  

I think https://app.defisaver.com/ was one.

Let's test it out (the UX is really bad on this dapp so we will need a bit of patience) and decide how to proceed

JIOjosBG:
Let me give more context:
## There are two types of delegate txns 
### 1) The safe UI or whoever plugs into the API, POSTs a bundle with destination and operation type 1 for delegate in the global object and the data encodes multiple calls. This is what this global account-op-level warning addresses
### 2) there is a normal transaction in a bundle, which by itself can have operation=1 (delegation) and there should be a call-level warning - this is your example with defi saver


## Case 1
what you are referring to in defisaver is handled in a call-level warning (that is how they chose to send the call to the safe wallet). This imo can happen more often
EDIT: i investigated further, they make the safe account a 1/1 signer to another safe (0xc5...e2 in my case) that delegates to the RecipeExecutor
image

## Possible change for case 2
The warning in this PR appears for account-op-level delegation and if you do not like the UI above it can be like this below instead. Afaik this will happen if the wallet providers that post to the Safe global API chooses a contract that we have not whitelisted or a hacker successfully tampered with the data before we receive it. We also get `Hold to sign` here

image

@Oxbobby wdyt?

Oxbobby:
@JIOjosBG I like your initial solution, and I also like solution 2) where we move it to "Hold to sign". Unrelated to this PR: what I don't like is that we actually have a choice for this - we should either use the modal warning on click for all cases or "hold to sign" for all cases.

I think hold to sign is better because it's applied always on the first visible button. In the case with safe accounts where threshold > 1, that would be "begin signing". It means you will need to hold before signing begins, not at the execution phase. And in general, it will always be applied on the first visible button regardless of the account type

I see this going like this:
1. In this PR, implement the "hold to sign" solution. If possible, fix the huge margin left on that button (also, the state of the button should be "warning")
2. Open another PR that replaces all modal pop ups with the "Hold to sign" mechanism. We will need @Ivshti or @dmngdsm's approval for this to happen so we should consult with at least one of them before proceeding

JIOjosBG:
for now:
- ok, banner on top + hold to sign (no modal)
- the button margin will be addressed here https://github.com/AmbireTech/ambire-app/pull/7580
- imo the warning button state is not for this PR
- all to `Hold to sign` in the future - https://github.com/AmbireTech/ambire-app/issues/7581

Oxbobby (approved):
Works as it should - detecting the warning, displaying it and showing the "hold to proceed" button. Unrelated to this PR: the button state after "hold to proceed" stays "Proceed" instead of reverting back to the Safe button states. I will open a PR towards release aiming to fix this

https://github.com/user-attachments/assets/40eb4739-df9a-4217-8278-e158411619e8

Oxbobby:
The fix for "Hold to sign/proceed" that I'm proposing: https://github.com/AmbireTech/ambire-app/pull/7586

JIOjosBG:
Legend

JIOjosBG:
~I am changing the warning text to
`This Safe transaction delegates permissions to a contract  that is not whitelisted. Proceed with caution.`~

Edit: Nope, i will leave it like this

PR #2564, opened 2026-07-22
Fix safe/ signature ordering

Oxbobby:
Reference PR that was pointed to v2: https://github.com/AmbireTech/ambire-common/pull/2563

The descriptrion with the problem is there 

EDIT: fix broken transfer tests as well because of missing properties passed to the constructor

### DIFF
--- a/src/libs/safe/safe.ts
+++ b/src/libs/safe/safe.ts
@@ -325,13 +325,17 @@ export function toCallsUserRequest(
           value: call.value,
           data: call.data
         }))
-      } catch (e) {
+      } catch {
         // this just means it's not a batch
         calls = [{ to: txn.to, value: BigInt(txn.value), data: txn.data || '0x' }]
       }
 
-      const signature = txn.confirmations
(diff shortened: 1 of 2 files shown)


PetromirDev (approved):

src/libs/safe/safe.ts
```diff
@@ -325,13 +325,17 @@ export function toCallsUserRequest(
...
         }))
-      } catch (e) {
+      } catch {
         // this just means it's not a batch
         calls = [{ to: txn.to, value: BigInt(txn.value), data: txn.data || '0x' }]
       }
 
-      const signature = txn.confirmations
-        ? (concat(txn.confirmations?.map((c) => c.signature)) as Hex)
+      let signature = txn.confirmations
```
superKalo: Minor: `signature` is never reassigned, so this should stay `const` instead of let.

I wonder why eslint doesn't complain about this and the reason is that we cut off many rules (cc @PetromirDev). Maybe we should bring back this one?

Not a blocker though.

superKalo (approved):

JIOjosBG (approved):

PR #2569, opened 2026-07-24
Feature: Safe nonce management

Oxbobby:
### Summary

This PR adds controller support for custom Safe transaction nonces, prevents pending Safe requests from replacing each other, and simplifies pending transaction banners.

#### Safe nonce management

- Add `setSafeNonce` to `SignAccountOpController`.
- Allow nonce changes only for unsigned Safe transactions when signing is not in progress.
- Validate custom nonces against the `uint256` range.
- Update both the account operation nonce and `safeTx.nonce`.
- Clear stale signatures, transaction IDs, and user-operation data after a nonce change.
- Rebuild the Safe EIP-712 data and associated hashes.
- Preserve explicitly selected nonces across estimation and pending-state refreshes.
- Add tests covering:
  - EIP-712 data regeneration.
  - Persistence across estimation updates.
  - Non-Safe and signed transaction guards.
  - Signing-in-progress guards.
  - Invalid nonce ranges.

#### Safe request handling

- Give Safe requests unique request IDs so a newly created request cannot replace an existing pending or partially signed request.
- Add regression coverage for multiple partially signed Safe requests occupying consecutive nonces.

#### Dashboard banners

- Group pending Safe requests into one banner per network.
- Count Safe requests instead of their individual calls.
- Shorten pending transaction titles to a single line.
- Include the chain ID in banner metadata so the app can render the network icon.


src/consts/safe.ts
```diff
@@ -54,3 +41,5 @@ export const allowedMulticallContracts = [
   '0xA1dabEF33b3B82c7814B6D82A79e50F4AC44102B',
   '0x8D29bE29923b68abfDD21e541b9374737B49cdAD'
 ]
+
+export const MAX_SAFE_NONCE = (1n << 256n) - 1n
```
JIOjosBG: you can use the `import { maxUint256 } from 'viem'`

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -534,6 +538,44 @@ export class SignAccountOpController
...
+  #updateNonce(nonce: bigint) {
+    if (this.#customSafeNonce !== null) return
+    this.#updateAccountOp({ nonce })
+  }
+
+  setSafeNonce(nonce: bigint) {
+    if (!this.account.safeCreation) return
+    if (this.status?.type && noStateUpdateStatuses.includes(this.status.type)) return
+    if (this.accountOp.signed?.length || this.accountOp.safeTx?.confirmations?.length) return
+    if (nonce < 0n || nonce > MAX_SAFE_NONCE) return
```
JIOjosBG: will it be better to add error toasts here so we can debug based on user report if something here misbehaves?

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -534,6 +538,44 @@ export class SignAccountOpController
...
+
+    this.#customSafeNonce = nonce
```
JIOjosBG, line 586: can we validate the current onchain safe nonce, so we do not allow the user to set old nonce to the current tx?

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -534,6 +538,44 @@ export class SignAccountOpController
...
+      safeTx: this.accountOp.safeTx
+        ? {
+            ...this.accountOp.safeTx,
+            nonce: nonce.toString()
+          }
+        : undefined,
+      signature: null,
+      txnId: undefined,
+      asUserOperation: undefined
+    })
```
JIOjosBG, line 598: two things:
- if the safeTx is missing you can just early return from the method `setSafeNonce` (with toast) in the first lines
- You can just pass the nonce like this `{nonce: nonce.toString()}` without the `...this.accountOp.safeTx,`

src/libs/banners/banners.ts
```diff
@@ -181,19 +180,33 @@ const getSafeBanner = ({
...
+            actionName: 'reject-accountOp',
+            meta: {
+              err: 'User rejected the transaction request.',
+              requestId: requests[0]!.id,
+              shouldOpenNextAction: false
+            }
+          }
+        : undefined
   }
 }
```
JIOjosBG, line 214: if someone calls this function while `requests = []` it will throw. 
Make it type-safe and remove the `!` from `requests[0]!.id` in both places

one way is to make the function return `Banner | null` and have `if(!requests[0]) return null`

or make the function accept `(requestsCount: number, firstRequest: CallsUserRequest, ....)`

src/libs/banners/banners.ts
```diff
@@ -213,21 +226,20 @@ export const getAccountOpBanners = ({
...
-        if (grouped.length === 1) {
-          remainingRequests = [...remainingRequests, ...grouped]
-          return
-        }
-        txnBanners.push(getSafeBanner({ requests: grouped, network, selectedAccount }))
-      })
-    } else remainingRequests = requests
+    // push all safe request for 1 network in a single banner
+    if (!!selectedAccount.safeCreation) {
+      const network = networks.find((n) => n.chainId.toString() === netId)!
```
JIOjosBG: are you sure the network wont be missing?

Pls add some handling for the case where it is missing like simple return, err toast or silent error log so we can avoid the `!`

src/controllers/signAccountOp/signAccountOp.ts
Oxbobby: Yes. We're doing it on the FE in SafeNonce.tsx -> isValidSafeNonce(). If we have information on what the latest nonce is, we do not allow the user to set a previous one

src/controllers/signAccountOp/signAccountOp.ts
Oxbobby: 1. Fixed ✅ 
2. There are two nonces to override: the account op one and the safeTx one. I don't understand how we can override the safeTx one with `{nonce: nonce.toString()}` only

src/libs/banners/banners.ts
Oxbobby: Nice one. Fixing by returning null as well

src/libs/banners/banners.ts
Oxbobby: adding

src/controllers/signAccountOp/signAccountOp.ts
Oxbobby: added 👍

src/controllers/signAccountOp/signAccountOp.ts
JIOjosBG: my bad for the second one, i misunderstood

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -534,6 +539,87 @@ export class SignAccountOpController
...
+  setSafeNonce(nonce: bigint) {
+    if (!this.account.safeCreation || !this.accountOp.safeTx) {
+      const message = 'Nonce could not be set as something unexpected happened'
+      this.emitError({
+        message,
+        error: new Error(message),
+        level: 'minor'
+      })
+      return
+    }
```
JIOjosBG: Lets split this into two
`Nonce could not be set, Safe account data is missing`
and 
`Nonce could not be set, Safe transaction data is missing`

Oxbobby: Overall, it was not correct to check whether `this.accountOp.safeTx` was missing as it's always missing at the beginning. I fixed this

PR #2572, opened 2026-07-27
Add a registry to decode plain multicalls

Oxbobby:
The final result is a multicall traverse:

Screenshot 2026-07-27 at 16 09 17


src/libs/humanizer/erc7730/humanize.ts
```diff
@@ -1214,32 +1231,45 @@ const getSafeCallFallbackVisualization = (
...
+      }
+    })
 
-  if (!humanizedCall?.fullVisualization?.length) {
-    const [fallbackCall] = genericErc20Humanizer({ accountAddr }, [call as IrCall])
-    humanizedCall = fallbackCall
+    if (!humanizedCall?.fullVisualization?.length) {
+      const [fallbackCall] = genericErc20Humanizer({ accountAddr }, [call as IrCall])
+      humanizedCall = fallbackCall
+    }
```
JIOjosBG, line 1260: we kinda did the same thing, imo both your and mine solution are good only for the hotfix, but long term i have an idea how to refactor a bit

https://github.com/AmbireTech/ambire-common/pull/2574

other than the change in the modules, imo the multicall part looks good, but why not add it from the relayer side?

Oxbobby: We still need this PR as otherwise, the multicall is not destructured into separate calls:

Screenshot 2026-07-28 at 10 05 19

PR #2571, opened 2026-07-24
Humanizer vuln: approve in multicall to ERC-4626 token contracts

JIOjosBG:
Also a security vuln as a user will not know if they are approving vault token if this module is not humanizaing

Morpho vaults are such case.

The contract allows to execute multiple contract functions to the same contract using `multicall(bytes[])`

testable with this tx [approve and deposit tx](https://transact.swiss-knife.xyz/send-tx?chainId=1&calldata=0xac9650d800000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000044095ea7b3000000000000000000000000d8da6bf26964af9d7eed9e03e53415d37aa960450000000000000000000000000000000000000000000000000de0b6b3a76400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000446e553f650000000000000000000000000000000000000000000000000de0b6b3a76400000000000000000000000000006969174fd72466430a46e18234d0b530c9fd5f4900000000000000000000000000000000000000000000000000000000&to=0xdd0f28e19C1780eb6396170735D45153D261490d0) and this [approve tx](https://transact.swiss-knife.xyz/send-tx?chainId=1&calldata=0xac9650d80000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000044095ea7b3000000000000000000000000d8da6bf26964af9d7eed9e03e53415d37aa960450000000000000000000000000000000000000000000000000de0b6b3a764000000000000000000000000000000000000000000000000000000000000&to=0xdd0f28e19C1780eb6396170735D45153D261490d)

image

image

PR #2573, opened 2026-07-27
Make the user hold to sign when there's an approval, increase allowance or permit approval / increaseApproval in a multicall

Oxbobby:
We switch signAccountOp to a warning state when there's an approval / allowance call as part of a multicall

Screenshot 2026-07-27 at 14 24 26


JIOjosBG:
imo this should not be merged at all because:
- an approve call, executed from a contract is not executed from the user
- since recently we have humanizer for the morpho tokens, so the approvals are already handled by the new functional module https://github.com/AmbireTech/ambire-common/pull/2571
- this contributes to warning fatigue and i cant see a case where the user is at more risk than doing simple erc-20 approve

PR #2576, opened 2026-07-28
Fix / Suspicious hosting false positive

jordan-enev:
Fix: Read the dApp hosting context from the request's own top frame instead of co-sessions in the same tab, which wrongly warned SUSPICIOUS_HOSTING when a stale session from an earlier visit matched by tabId.

Closes https://github.com/AmbireTech/ambire-app/issues/7596.


src/controllers/dapps/dapps.ts
```diff
@@ -1233,7 +1255,7 @@ export class DappsController extends EventEmitter implements IDappsController {
     if (!validDappUrls.length) return null
 
     const sessionForId = sessionId ? this.dappSessions[sessionId] : undefined
```
superKalo, line 1257: The banner re-reads the frame context from the live session every time it renders, and `updateFrameContext` overwrites that context on every incoming request. Since `sessionId` is `windowId-tabId-dappId` with no frame component, the phishing top frame and the embedded dApp share one session.

So a phishing page can embed the dApp, trigger a transaction (banner shows `SUSPICIOUS_HOSTING`), then navigate its own top frame to the real dApp - the dApp's page load fires a request from the same tab, the context is refreshed to the legit origin, and the warning silently disappears from the sign request that is already open.

Could we snapshot the top frame origin onto the user request / accountOp when it is created, and read it from there instead of from the live session?

src/controllers/dapps/dapps.ts
```diff
@@ -1179,31 +1209,23 @@ export class DappsController extends EventEmitter implements IDappsController {
...
    *
-   * This detects phishing pages that host a legitimate dApp in an iframe: the legitimate
-   * dApp's own session looks clean, but the phishing page's session (e.g. sites.google.com)
-   * is in the same tab and has SUSPICIOUS_HOSTING status.
-   *
-   * Returns `undefined` if no dangerous co-session is found, or if the status cannot yet
-   * be determined (phishing DB not loaded and domain not in the static list).
+   * `undefined` means no verdict: the dApp is the top frame itself, the platform reports no frame
+   * context (mobile WebViews, WalletConnect), or the top frame is not dangerous. Frames between
+   * the top one and the dApp are not visible without the `webNavigation` permission - not checked.
```
superKalo, line 1218: Worth flagging that this is a real loss of coverage (regression): the content script runs in all frames down to depth 10, so a middle frame normally does get a session and the old co-session check would have caught attacker.com > sites.google.com/fake > app.aave.com. Now only the top frame is checked, so that chain produces no warning.

One way to keep both properties without the `webNavigation` permission: keep the co-session scan, but match co-sessions by equal `topFrameOrigin` instead of by `tabId`. A stale session from an earlier visit carries the old page's top frame origin, so it gets excluded - which is exactly the false positive this PR is fixing - while frames in the same document tree still contribute.

superKalo (approved):

src/controllers/dapps/dapps.ts
superKalo: We agreed offline that it's fine as is.

src/controllers/dapps/dapps.ts
superKalo: We agreed offline that it's fine as is.

PR #2578, opened 2026-07-29
Fix Send infinite loading after hung portfolio updates

Maskln:
closes https://github.com/AmbireTech/ambire-app/issues/7574

## Summary
- Prevents the Send token form from staying on an infinite skeleton when a portfolio network update hangs (e.g. after long idle / bad connectivity).
- Adds age-bounded `isLoading` skip logic and a 60s hard timeout around network portfolio fetches so stuck updates clear loading, unblock the per-network queue, and can be retried.
- Adds timeouts for AAVE `staticCall` (15s) and custom DeFi position fetches (30s) so one hung RPC cannot pin the portfolio update forever.

## Root cause
SendForm gates the token UI on `portfolio.isReadyToVisualize`. A hung portfolio update left `isLoading=true`, which made later updates skip forever and kept Send stuck until the extension was disabled/enabled.

## Test plan
- [ ] Leave Send open for a long idle period (or throttle network), confirm the token form recovers instead of skeleton-locking forever
- [ ] Manual refresh after a failed/timeout portfolio update still works
- [ ] Unit: `hung portfolio update recovery` in `portfolio.test.ts`
- [ ] Unit: AAVE / `getCustomProviderPositions` timeout cases in `defiPositions.test.ts`

### DIFF
--- a/src/consts/intervals.ts
+++ b/src/consts/intervals.ts
@@ -15,3 +15,11 @@ export const PHISHING_FAILED_TO_GET_UPDATE_INTERVAL = 600000 // 10 minutes
 export const ESTIMATE_UPDATE_INTERVAL = 30000
 export const GAS_PRICE_UPDATE_INTERVAL = 12000
 export const FETCH_SAFE_TXNS = 3 * 60 * 1000 // 3 minutes
+/** Hard ceiling for a single network portfolio update (tokens + custom DeFi). */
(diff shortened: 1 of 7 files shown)


gergana95 (approved):

Maskln:
Continued in new PR: https://github.com/AmbireTech/ambire-common/pull/2611

PR #2582, opened 2026-07-30
Feature: import safes by owner

Oxbobby:
## Summary

Adds controller support for discovering Safe accounts owned by an address across all enabled Safe-supported networks. It also introduces a coordinated account update operation for adding selected accounts and removing deselected imported accounts.

## What changed

- Added Safe discovery by owner through the Safe Transaction Service.
- Search requests are processed in batches of four networks.
- Results are published after each batch, allowing the UI to display accounts progressively.
- Safe accounts deployed on multiple networks are deduplicated by address and their deployment networks are combined.
- Shared Safe information is fetched from one deployed network at a time, falling back to the next network if the request fails.
- Partial network failures are retained without discarding successful results.
- Stale searches cannot overwrite the results of a newer search or reset.
- Added the `SafeAccountByOwner` account type, including its deployment networks.
- Added `MainController.updateAccounts`, which performs account removals and additions as one coordinated operation.
- Removal requests are limited to imported accounts, deduplicated, and protected against removing an address that is also being added.
- Improved error context in the Safe API helpers.

## Testing

- Added Safe owner discovery tests covering:
  - Batched and progressive network searches
  - Partial network failures
  - Fallback between deployed networks
  - Owner validation
  - Stale search protection
- Added account update tests covering:
  - Adding selected and removing deselected imported accounts
  - Ignoring removal requests for non-imported accounts
  - Protecting addresses present in both operations


Oxbobby:
AI feedback: 1) not relevant 2) addressed

src/controllers/safe/safe.ts
```diff
@@ -116,7 +132,7 @@ export class SafeController extends EventEmitter implements ISafeController {
       safeNetworks.map((n) =>
         this.#providers.providers[n.chainId.toString()]!.getCode(safeAddr)
           .then((code) => ({ chainId: n.chainId, code }))
-          .catch((e) => ({ chainId: n.chainId, code: '0x' }))
```
JIOjosBG, line 119: nice

src/controllers/safe/safe.ts
```diff
@@ -169,6 +185,189 @@ export class SafeController extends EventEmitter implements ISafeController {
     this.importError = undefined
   }
 
+  async #getSafeAccountByOwner(
```
JIOjosBG: this whole function can be moved in the library so the controller is kept short and readable.

Also the testsfor it can be in the lib tests

JIOjosBG: Actually this whole function is not needed as it is just a wrapper, you can delete the outer function and just call the inner one instead

src/controllers/safe/safe.ts
```diff
@@ -169,6 +185,189 @@ export class SafeController extends EventEmitter implements ISafeController {
...
+      const apiKit = getApiKit(chainId)
+      try {
+        const safeInfo = await withTimeout(() => apiKit.getSafeInfo(safeAddr), {
+          timeoutMs: SAFE_API_TIMEOUT_MS,
+          message: `Safe API: get Safe info timed out after ${SAFE_API_TIMEOUT_MS}ms`
+        })
+        const address = getAddress(safeAddr.toLowerCase()) as Hex
+        const owners = safeInfo.owners.map((safeOwner: string) =>
+          getAddress(safeOwner.toLowerCase())
+        ) as Hex[]
```
JIOjosBG: there is no point in having our types be `Hex` instead of `string` if we are not gonna validate and cast values from external sources

in this case you can remove the casting without errs, please do so

src/controllers/safe/safe.ts
```diff
@@ -169,6 +185,189 @@ export class SafeController extends EventEmitter implements ISafeController {
...
+            associatedKeys: owners,
+            initialPrivileges: owners.map((safeOwner) => [safeOwner, '0x01']),
+            creation: null,
+            safeCreation: {
+              factoryAddr: safeCreationInfo.factoryAddress as Hex,
+              singleton: safeCreationInfo.singleton as Hex,
+              setupData: safeCreationInfo.setupData as Hex,
+              saltNonce: safeCreationInfo.saltNonce
+                ? (toBeHex(BigInt(safeCreationInfo.saltNonce), 32) as Hex)
+                : (toBeHex(0, 32) as Hex),
```
JIOjosBG: same as above, we should validate data and avoid casting so the Hex type in the interface does its job instead of being just friction

The best proposal i have 1) remove the casting and 2) above the return statement add
```
        if (!isHex(safeCreationInfo.factoryAddress) || !isHex(...) || ...) {
          // error message with console.error or emitError
          // return with either {account:null, failed:true} or getAccountFromChain(remainingChainIds, true)
        }

```

src/controllers/safe/safe.ts
JIOjosBG: This function assumes that the owners from the first successful chain fetch are the owners on every chain for this acc, which might be problematic, but our codebase is just designed like that for now. 

It would be nice to be able to warn users if the account has different signers on different chains.

If you have concerns with spamming the safe API, i am fine with just adding a warning like "Don't add accounts you do not recognize"

Image

src/controllers/safe/safe.ts
```diff
@@ -169,6 +185,189 @@ export class SafeController extends EventEmitter implements ISafeController {
...
+    const owner = getAddress(ownerAddress) as Hex
+    const searchId = ++this.#safeOwnerSearchId
+    await this.withStatus('findSafesByOwner', () => this.#findSafesByOwner(owner, searchId), true)
+  }
+
+  async resetFindSafesByOwner() {
+    this.#safeOwnerSearchId += 1
+    this.safeOwnerSearch = undefined
+    this.emitUpdate()
+  }
```
JIOjosBG: i am slightly bearish on having the UI be responsible for the resetting controller state, but lets not overengineer unless there are problems with this

JIOjosBG (changes requested):

src/controllers/safe/safe.ts
Oxbobby: I'm not a fan of validating API inputs, and especially more at random places. If we're going to do this validation, it has to be in the lib.

The purpose of the Hex typehint is not validation. It's to say to the developer "hey, this is not only a string, it is a Hex string". It is especially helpful on fields that are supposed to be numbers but are coded as 'string'. At those times, you are left wondering is this a number, is this a number as a string or is this a hex. So coding it to Hex gives an easy answer.

I'm against applying this fix. Everything is wrapped in a try/catch so the app won't blow up

src/controllers/safe/safe.ts
Oxbobby: Different signers on different chains is not strange for a Safe account. And yes, as you said, we are a cross chain app, so we cannot import the Safe account only on Ethereum but not on Optimism.

However, I agree on adding a warning "Don't add accounts you do not recognize". I will do so 👍

Oxbobby: I will move it to a lib 👍

src/controllers/safe/safe.ts
Oxbobby: Resolved this but using owner refs on the UI and deleted this method

JIOjosBG (approved):
lgtm 🚀

PR #2585, opened 2026-07-31
Refactoring / Create new account flow

sonytooo:
* Implement create new recovery phrase in the background and back up later flag


src/controllers/keystore/keystore.ts
```diff
@@ -162,11 +162,15 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
       ])
       this.keyStoreUid = keyStoreUid
       this.#keystoreSeeds = keystoreSeeds.map((s) => {
-        if (s.id) return s
+        // Seeds stored before the backup flow was introduced were always revealed
+        // to the user on creation/import, so they count as backed up.
+        const seed = s.isBackedUp === undefined ? { ...s, isBackedUp: true } : s
+
+        if (seed.id) return seed
```
PetromirDev: Again, imo it'd be much simpler to have the opposite flag - notBackedUp

PetromirDev (changes requested):
Only feedback is the comment below (which will simplify changes greatly)

PR #2590, opened 2026-08-03
Changes: robinhood gas limit overhead, EOA broadcast retry

Oxbobby:
Changes:
1. Gas limit overhead for all chains except Robinhood is 10%. For Robinhood, it's changed to 20% as there are a lot of OOG txns
2. When estimateGas() fails, it checks if the account is an EOA account (not 7702) and it has calls > 1. If that is the case, then estimateGas() retries until it reaches the threshold set. Otherwise, it returns an error immediately as this is an always failing txn

### DIFF
--- a/src/libs/broadcast/broadcast.ts
+++ b/src/libs/broadcast/broadcast.ts
@@ -26,8 +26,17 @@ export const BROADCAST_OPTIONS = {
 }
 
 async function waitBeforeRetry(chainId: bigint) {
-  // block time at ethereum is bigger, so we wait 3s per failures
-  await wait(chainId === 1n ? 3000 : 1500)
+  // wait a bit longer on ethereum as txn confirmations are slower
+  await wait(chainId === 1n ? 2000 : 1000)
+}
+
+/**
+ * The default gas limit overhead for all chains is 10%.
+ * Robinhood uses 20% because transactions there frequently run out of gas.
+ */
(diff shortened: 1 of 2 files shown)


src/libs/broadcast/broadcast.ts
```diff
@@ -109,20 +120,28 @@ async function estimateGas(
...
-  // the error is most likely because of an incorrect RPC pending state
+  // Sequential EOA calls can fail temporarily if the RPC pending state is stale.
   if (gasLimit instanceof Error || hasNonceDiscrepancyOnApproval) {
-    // if the gasLimit is throwing because the smart account is returning INSUFFICIENT_PRIVILEGE,
-    // return the error without retrying
-    if (gasLimit instanceof Error && gasLimit.message.includes('INSUFFICIENT_PRIVILEGE'))
-      throw gasLimit
+    // Other estimation errors are deterministic, so return them immediately.
+    const isEoaBatch = broadcastOption === BROADCAST_OPTIONS.bySelf && op.calls.length > 1
+    if (gasLimit instanceof Error && !isEoaBatch) throw gasLimit
```
PetromirDev, line 127: We used to throw errors only if `gasLimit instanceof Error && gasLimit.message.includes('INSUFFICIENT_PRIVILEGE'`, so this is not a little change. Just want to make sure that this is on purpose

PetromirDev (approved):

PR #2592, opened 2026-08-04
Fix: call the paymaster for a final sponsorship check

Oxbobby:
Add: an additional swap validation check call to the paymaster before declaring the sponsorship


Oxbobby:
All AI comments addressed

PetromirDev (approved):
Was a bit worried about the timeout but it doesn't feel that slow even with 3G throttling 👍

PR #2594, opened 2026-08-04
Fix: set default fee speed in storage to the last selected by user

Oxbobby:


PetromirDev:
Regarding the PR reviewer's comment - IMO `setFeeSpeedPreference`'s pattern is better. I don't see a reason to rethrow the error only to catch it with a silent block later

PetromirDev (approved):
lfg

PR #2597, opened 2026-08-05
Allow ERC-7677 sponsorship retry after bundler estimation

Oxbobby:
## What changed

ERC-7677 paymaster services are now retried during `upgrade()` when:

- The initial sponsorship request did not succeed.
- The service is still available.
- The service has not previously failed.

The retry uses the gas limits returned by the bundler together with the medium gas price. This provides the paymaster service with a more realistic UserOperation than the initial request, whose gas-related fields may still be empty.

The paymaster selection logic was also reorganized into dedicated helpers for ERC-7677 and swap sponsorship.

## Why

In the case of heyAura, we need to have a proper estimation to know whether the paysmater will be willing to sponsor the transaction or not (it will depend on the cost of the txn). Therefore, we attach the estimation data to the userOp and send to to the paymaster for a retry

## Additional changes

- Added the `ambireGnosisSponsorship` policy identifier to Ambire’s internal paymaster service context.
- Preserved an existing ERC-7677 sponsorship instead of replacing it with another paymaster type.
- Preserved swap sponsorship priority when applicable.
- Kept the account deployment fields in ERC-7677 requests for undeployed accounts.

### DIFF
--- a/src/libs/erc7677/erc7677.ts
+++ b/src/libs/erc7677/erc7677.ts
@@ -14,6 +14,7 @@ import {
 } from './types'
 
 export const AMBIRE_SWAP_POLICY = 'ambireSwapSponsorship'
+export const AMBIRE_GNOSIS_POLICY = 'ambireGnosisSponsorship'
 
 export function getPaymasterService(
   chainId: bigint,
@@ -54,7 +55,10 @@ export function getAmbirePaymasterService(
 
   return {
     url: getAmbireSponsorshipUrl(relayerUrl),
-    id: new Date().getTime()
+    id: new Date().getTime(),
+    context: {
+      policyId: AMBIRE_GNOSIS_POLICY
+    }
   }
 }
 
--- a/src/libs/erc7677/erc7677.ts
+++ b/src/libs/erc7677/erc7677.ts
@@ -14,7 +14,11 @@ import {
 } from './types'
 
 export const AMBIRE_SWAP_POLICY = 'ambireSwapSponsorship'
-export const AMBIRE_GNOSIS_POLICY = 'ambireGnosisSponsorship'
+
+/**
+ * Currently, only the Gnosis chain is sponsored unconditionally
+ */
+export const AMBIRE_NETWORK_WIDE_SPONSORSHIP_POLICY = 'ambireGnosisSponsorship'
 
 export function getPaymasterService(
   chainId: bigint,
@@ -57,7 +61,7 @@ export function getAmbirePaymasterService(
     url: getAmbireSponsorshipUrl(relayerUrl),
     id: new Date().getTime(),
     context: {
-      policyId: AMBIRE_GNOSIS_POLICY
+      policyId: AMBIRE_NETWORK_WIDE_SPONSORSHIP_POLICY
     }
   }
 }
--- a/src/libs/paymaster/paymaster.ts
+++ b/src/libs/paymaster/paymaster.ts
@@ -487,7 +487,7 @@ export class Paymaster extends AbstractPaymaster {
     // apply estimation and gas changes to the userOp so a more realistic,
     // final userOp could be sent over for paymaster stub data
     //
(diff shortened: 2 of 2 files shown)


src/libs/paymaster/paymaster.ts
```diff
@@ -100,6 +103,36 @@ export class Paymaster extends AbstractPaymaster {
...
+          this.#account.creation.salt
+        ])
+      }
+
+      const response = await Promise.race([
+        getPaymasterStubData(this.paymasterService, localOp, this.network),
+        new Promise((_resolve, reject) => {
+          setTimeout(() => reject(new Error('Sponsorship error, request too slow')), 5000)
+        })
+      ])
```
PetromirDev: clear this timer

src/libs/paymaster/paymaster.ts
```diff
@@ -451,4 +452,41 @@ export class Paymaster extends AbstractPaymaster {
...
+    // apply estimation and gas changes to the userOp so a more realistic,
+    // final userOp could be sent over for paymaster stub data
+    const localOp = { ...userOp }
+    localOp.preVerificationGas = bundlerEstimateResult.preVerificationGas
+    localOp.verificationGasLimit = bundlerEstimateResult.verificationGasLimit
+    localOp.callGasLimit = bundlerEstimateResult.callGasLimit
+    localOp.maxFeePerGas = gasPrices.medium.maxFeePerGas
+    localOp.maxPriorityFeePerGas = gasPrices.medium.maxPriorityFeePerGas
+
+    await this.#tryToSetSwapSponsorship(bundlerEstimateResult, gasPrices, userOp)
```
PetromirDev: why do we pass userOp and not localOp?

src/libs/paymaster/paymaster.ts
```diff
@@ -451,4 +452,41 @@ export class Paymaster extends AbstractPaymaster {
...
+    // ERC7677 is already sponsoring the userOperation so we don't upgrade over it
+    if (this.type === 'ERC7677' || !this.network) return
+
+    // do not upgrade over the gnosis paymaster sponsorship
+    if (this.type === 'Ambire' && this.paymasterService?.context?.policyId === AMBIRE_GNOSIS_POLICY)
+      return
+
+    // apply estimation and gas changes to the userOp so a more realistic,
+    // final userOp could be sent over for paymaster stub data
+    const localOp = { ...userOp }
```
PetromirDev, line 491: this is a shallow copy and if someone ever changes it to structuredClone or some other deep cloning method this will silently break `meta.paymasterService.failed` (as PaymasterFactory:48 would stop mutating the original op)

PetromirDev: Consider adding a comment

src/libs/erc7677/erc7677.ts
```diff
@@ -54,7 +55,10 @@ export function getAmbirePaymasterService(
 
   return {
     url: getAmbireSponsorshipUrl(relayerUrl),
-    id: new Date().getTime()
+    id: new Date().getTime(),
+    context: {
+      policyId: AMBIRE_GNOSIS_POLICY
+    }
```
PetromirDev, line 65: It took me some time to understand this - why is `AMBIRE_GNOSIS_POLICY` set unconditionally without a network check. Well, I found out - `isSponsorable` contains the check that is limited to gnosis atm. This is not easy to understand. Renaming the constant/method could help - `AMBIRE_NETWORK_WIDE_SPONSORSHIP_POLICY`/`qualifiesForNetworkWideSponsorship`.

PetromirDev (changes requested):

src/libs/paymaster/paymaster.ts
Oxbobby: Bug. I tested it and it worked but I guess it was luck

PetromirDev (approved):

PR #2598, opened 2026-08-05
POC: Faster mobile app non webview

PetromirDev:


PetromirDev:
Cleaned up and moved to https://github.com/AmbireTech/ambire-common/pull/2621

PR #2599, opened 2026-08-05
Swap and bridge/ add token exchanges

JIOjosBG:
## Added
 - the swap and bridge can fetch debounced and batched request to cena for token exchanges once the list of displayable tokens after the list is filtered sorted and sliced


src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -116,6 +125,32 @@ const STATUS_WRAPPED_METHODS = {
...
+  LOADING: MARKET_DATA_LOADING_DEADLINE
+}
+
+type ToTokenMarketDataRecord = {
+  status: ToTokenMarketDataStatus
+  updatedAt: number
+  // Only set when the status is DONE
+  data?: TokenDataCacheValue
+}
+
```
Oxbobby, line 132: I know it's contradictory as there are other consts in the file but let's move these into a new file in consts called tokenMarketData.ts. With time, we should strive to eliminate the other consts in the file as well

Oxbobby (approved):
Great. I'm approving with a recommendation to tidy up the consts in the swapAndBridge.ts controller. However, we can continue without it as well

PR #2601, opened 2026-08-06
Feature/ Throttled(delayed emits) and implement them in portfolio and account state

PetromirDev:
We are bombarding the extension with useless updates when doing emitUpdate after every network change (users can have 15+, 20+ networks). The idea is very simple - throttle every emit until the last one so isLoading flags are not delayed.

Closes https://github.com/AmbireTech/ambire-app/issues/7658

### DIFF
--- a/src/controllers/accounts/accounts.ts
+++ b/src/controllers/accounts/accounts.ts
@@ -199,6 +199,8 @@ export class AccountsController extends EventEmitter implements IAccountsControl
 
     this.emitUpdate()
 
+    let readyNetworks = 0
+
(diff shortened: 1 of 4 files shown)


jordan-enev (approved):
It looks good. I have only one concern: I remember we had useEffects in the frontend that check for certain status changes, e.g. from LOADING to BROADCASTED. In those cases, we should be careful not to use throttling, because it could skip the intermediate status change.

You used it for the account state change, and it looks good there. I'm just pointing out a potential issue if we start using it more broadly.

PR #2602, opened 2026-08-06
Refactor/ Portfolio lookups and reduce formatting

PetromirDev:
Taken out of https://github.com/AmbireTech/ambire-common/pull/2598. Doesn't make a big difference on desktop, but can be felt on a single thread on Android (with big accounts like that of Vitalik).

### DIFF
--- a/src/libs/portfolio/portfolio.ts
+++ b/src/libs/portfolio/portfolio.ts
@@ -1,6 +1,8 @@
 import { ZeroAddress } from 'ethers'
 import { getAddress } from 'viem'
 
+import { getFeeToken } from '@/libs/portfolio/tokenProcessing'
+
 import BalanceGetter from '../../../contracts/compiled/BalanceGetter.json'
 import NFTGetter from '../../../contracts/compiled/NFTGetter.json'
 import gasTankFeeTokens from '../../consts/gasTankFeeTokens'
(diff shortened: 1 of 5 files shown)


src/libs/portfolio/portfolio.ts
```diff
@@ -245,17 +245,26 @@ export class Portfolio {
...
+        // getAddress may throw an error. This will break the portfolio
+        // if the error isn't caught
+        const checksummedAddress = getAddress(address)
+
+        seenErc20Hints.add(lowercasedAddress)
+        checksummedErc20Hints.push(checksummedAddress)
+      } catch {
+        // Not an address, so it can't be a token
+      }
+    })
```
JIOjosBG, line 269: to reduce verbosity you can:
```js
const deduppedLowerCaseTokens = [...new Set(hints.erc20s.map(t=>t.toLowerCase())]]]
const checksummedErc20Hints = deduppedTokens
      .map((address) => {
        try {
          // getAddress may throw an error. This will break the portfolio
          // if the error isn't caught
          return getAddress(address)
        } catch {
          return null
        }
      })
      .filter(Boolean) as string[]
```

JIOjosBG: nvm, the above is slower, this should be on par
```js
    const deduppedLowerCaseTokens = [...new Set(hints.erc20s.map((t) => t.toLowerCase()))]
    const checksummedErc20Hints = deduppedLowerCaseTokens.flatMap((address): string[] => {
      try {
        // getAddress may throw an error. This will break the portfolio
        // if the error isn't caught
        return [getAddress(address)]
      } catch {
        return []
      }
    })
```

JIOjosBG:
claude reported we have one more case of nasted loop in libs/portfolio/portfolio.ts:587

image

JIOjosBG (approved):

src/libs/portfolio/portfolio.ts
PetromirDev: I'd rather keep it as is because this version doesn't drop incorrectly checksummed addresses (because it adds all lowercased without validation).

PR #2603, opened 2026-08-07
Feature / Bring back mandatory invite gate in the Invite controller (for the mobile apps only)

superKalo:
The mobile app is going to enforce an invite gate as of v6.21 series, so the Invite controller now owns the whole verification flow.

The extension keeps its behaviour - it stores and reads the same `invite` entry, but does not enforce the gate.

Notes:

- We're reusing the Relayer endpoint we prev used for the extension invite gate long time ago when we initially released it (`/promotions/extension-key/:code`). Invite codes are with the same shape and similarly to the extension - we use them for analytics (see the -app changes).

Technical notes:

- `verify()` runs through `withStatus`, so the UI gets a loading state and stops swallowing failures
- new `errorMessage` state + `resetErrorState()`, so the app shows the failure in the form instead of a toast
- the message prefers the reason the Relayer sent back (vs the prev implementation); anything that never got a verdict (offline, timeout, 5xx, unreadable body) gets a neutral message, so a valid code is never blamed for a connection problem
- `grantAccess()` passes the gate without calling the Relayer, for users who were already on the app before it was introduced, and never overwrites an already verified code (see -app)

Also:

- Added a misisng the `x-app-env` request header type.


src/controllers/invite/invite.ts
```diff
@@ -79,50 +127,83 @@ export class InviteController extends EventEmitter implements IInviteController
...
+        throw new EmittableError({
+          message: this.errorMessage,
+          level: 'silent',
+          error,
+          sendCrashReport: false
+        })
+      }
+
+      this.errorMessage = ''
+      await this.#persistVerified(code)
```
PetromirDev: TODO: Why is this reset here

PetromirDev, line 176: It's a bit hard to understand on first glance written this way. IMO this code should be below  await this... in the try block as the catch block rethrows so this is never reached. I was quite puzzled for a bit.

src/controllers/invite/invite.ts
```diff
@@ -79,50 +127,83 @@ export class InviteController extends EventEmitter implements IInviteController
...
     await this.#initialLoadPromise
 
-    const becameOGAt = Date.now()
-    await this.#storage.set('invite', { ...this.#state, becameOGAt })
-
+    this.#state = { ...this.#state, becameOGAt: Date.now() }
     this.isOG = true
     this.emitUpdate()
+
+    await this.#storage.set('invite', this.#state)
```
PetromirDev, line 206: nice

src/controllers/invite/invite.ts
```diff
@@ -79,50 +127,83 @@ export class InviteController extends EventEmitter implements IInviteController
...
+    this.inviteStatus = this.#state.status
+    this.verifiedCode = this.#state.verifiedCode
+    this.emitUpdate()
+
+    await this.#storage.set('invite', this.#state)
+  }
+
+  /** Verifies an invite code against the relayer and, if valid, passes the gate for good. */
   async verify(code: string) {
     await this.#initialLoadPromise
```
PetromirDev, line 154: `#initialLoadPromise` is private, so nothing outside can wait for the load - `awaitControllersInitialLoad` reads `.initialLoadPromise` and gets `undefined` for this controller. That didn't matter while the extension only read `verifiedCode`, but the mobile gate now decides what to render from `inviteStatus`, and the UI's readiness check treats a controller as loaded the moment any state arrives. Let's add a public `isReady` set at the end of `#load()` - `ControllerStore.#isControllerReady` already gates on that flag when the state carries it.

src/controllers/invite/invite.ts
```diff
@@ -79,50 +127,83 @@ export class InviteController extends EventEmitter implements IInviteController
...
 
-    try {
-      const res = await this.#callRelayer(`/promotions/extension-key/${code}`, 'GET')
+    await this.withStatus('verify', async () => {
+      try {
+        await this.#callRelayer(`/promotions/extension-key/${code}`, 'GET')
```
PetromirDev: `code` goes into the path unescaped, and `sanitize` on the app side only strips spaces and dashes so `.` and `/` survive. `../../health` is exactly 12 chars, fetch normalises it to `/health`, and `relayerCallUncaught` returns `success: isStatusOk` for any body without a `success` key - so any 200 JSON reply from the relayer passes the gate and gets persisted as a verified code. `encodeURIComponent(code)` plus a charset check on the input closes it.

src/controllers/invite/invite.ts
```diff
@@ -79,50 +127,83 @@ export class InviteController extends EventEmitter implements IInviteController
...
+    this.#state = {
+      ...this.#state,
+      status: INVITE_STATUS.VERIFIED,
+      verifiedAt: Date.now(),
+      verifiedCode: code
+    }
+
+    this.inviteStatus = this.#state.status
+    this.verifiedCode = this.#state.verifiedCode
+    this.emitUpdate()
```
PetromirDev: `emitUpdate()` before the write means the gate can open on a state that never landed in storage. For `grantAccess` that self-heals since the hook dispatches again next launch, but someone who just typed a real code gets in and is then gated again on the next start with no way to know why - and `withStatus` also toasts the raw storage error at `major`. Let's await the `set` first, then emit.

src/controllers/invite/invite.ts
superKalo: Ah, I see. I agree on the readability (I added a comment making it explicit that the catch always throws in https://github.com/AmbireTech/ambire-common/pull/2603/changes/59d0d45b7f75adef20b010c411764b405b34affd).

I'd rather not move the call inside the try though - it would turn a storage failure into an "invalid code" message under the code field, and that's not an invite-code problem.

src/controllers/invite/invite.ts
superKalo: small touch, but while I was on this, decided to improve a bit :)

src/controllers/invite/invite.ts
superKalo: Great catch! I agree. I tweaked it in https://github.com/AmbireTech/ambire-common/pull/2603/changes/6ed768f6ca4371ef37265e03102caad32253bd83

I love how easy it was to build on top of the `isReady` logic, nice work with abstracting this to be scalable 👏

PetromirDev (approved):

PR #2607, opened 2026-08-10
Feature/safe queue

Oxbobby:
## Summary

Introduces the controller and library support required by the Safe Queue, including recoverable rejection, same-nonce transaction resolution, Safe Global refresh, and queue-aware portfolio simulation.

### Safe request lifecycle

- Removed the option to Reject a partially or fully signed Safe transaction
- Added the option to Cancel a partially or fully signed Safe transaction. Cancel means initiating an onchain transaction replacement with an alternative transaction with the same nonce that sends 0 value to address 0

### Same-nonce resolution

- Broadcasting a Safe transaction removes all competing transactions for the same account, network, and nonce.
- Automatically resolved transactions are persisted with account and network scope.
- Removed requests do not automatically open the next signing request.
- Failed or stuck broadcasts only restore matching transactions from the same account, network, and nonce.

### Portfolio simulation

- Dashboard simulation has been stopped for partially or fully signed Safe transaction. Only pending to be signed transaction are simulated on the dashboard (ones you are building right now) with the exception of the one transaction that you have opened in SignAccountOp (if any)

### Safe Global integration

- Added a guarded manual refresh operation with controller status reporting.
- Duplicate refresh actions are prevented while a refresh is already running.

### Other behavior

- Dashboard banners are only generated for Safe transactions that have not collected signatures.
- Added a shared Safe-aware account-op nonce helper.

## Automated coverage

Added coverage for:

- sequential Safe simulation across gaps and duplicate nonces
- same-nonce transaction discovery and cleanup
- account/network isolation of automatically resolved transactions
- manual refresh concurrency


Oxbobby:
AI comments: 1 & 2 is basically how it should work from now on as those txns should be visible in the queue. I'm slightly adjusting 3

JIOjosBG:
sorry if it sounds like i am shitting on this but i do not like the UI/UX
- too much info in that cramped space, it should be either minimal in the extension or fullscreen (i am leaning towards extension)

- lets not add extra button "Queue" on dash.  I proposed having this in the activity page. It fits best there - this is future activity

- i do not understand the "Reject" and "Restore" buttons, neither should the user. If i press "Open" only that tx should be queued, not the rest with that nonce or with future nonce or on other chains

- not enough horizontal space now for humanizer. If we are going with extension screen we should 1) display nonce above the tx and 2) remove the hierarchy that eats horizontal space by making txns with sequential nonces be kinda sideways scrollable like we have in the screenshot below for requests in sign account op.

There is HUGE potential in this feature, we can blow people's hats off with simplicity, lets collaborate

JIOjosBG:
One bug i had on my branch:
- if i click to simulate some N sequential txns if the simulation is successful the dashboard displays it properly
- if i then click to simulate any other sequence of txns which have one failing (for example expired swap) then the old simulation is persisted

For reference i merged this rework of the design

https://github.com/user-attachments/assets/924c4655-6bbb-40a3-9a3d-1b477da69876

JIOjosBG:
We agreed that:
-  in the dashboard for now we will display simulation of pending txns only if there is only one tx for that chain and we will not display it if there are competing and/or sequential txns on the same chain
- in the sign account op screen we will change the behavior of `Reject` to do what `Sign later` did. We will remove `Sign later` and we will change the wording or appearance of `Reject` 
- in separate PR we can implement the Safe txns history from the Safe API

Oxbobby:
The AI reviews comments are irrelevant as we changed the logic to always display transactions in the queue

src/libs/humanizer/modules/Safe/index.ts
```diff
@@ -84,34 +84,13 @@ export const getSafeHumanization = (
...
-    fullVisualization.push(
-      ...(nonce !== undefined && nonce !== null
-        ? [getAction('Reject'), getLabel('Tx with nonce'), getLabel(nonce, true)]
-        : [getAction('Reject currently queued transaction')])
-    )
-    return {
-      visuals: fullVisualization
-    }
-  }
-
```
JIOjosBG, line 114: why did you remove this logic, this module is responsible for safe txns? If there is not good reason for this, lets bring it back here, just change the text the user sees

src/libs/humanizer/modules/Safe/index.ts
```diff
@@ -84,34 +84,13 @@ export const getSafeHumanization = (
   to?: string,
   value?: string | number | bigint,
   data?: string,
-  setupHookDepth = 0,
-  nonce?: string | number | bigint | null
```
JIOjosBG, line 88: we are passing the nonce intentionally for the humaniztion, see next comment

src/libs/humanizer/modules/FallbackHumanizer/fallBackHumanizer.ts
```diff
@@ -36,9 +36,15 @@ export const fallbackHumanizer: HumanizerCallModule = (
         ]
       }
     case 'has-to:no-value:no-data':
-      // preserve a visualization already set by an earlier, more specific module (e.g. a
-      // Safe{WALLET} "reject queued transaction" call), instead of unconditionally
-      // overwriting it with a generic "Empty call to" label below
+      const safeNonce = getAccountOpNonce(accountOp)
+      if (accountOp.meta?.isOnchainSafeRejection && safeNonce !== null) {
```
JIOjosBG: long term i do not want to use accountOp in the humanizer.

Pls revert to the old logic and the way it used to handle safe tx rejections (now cancel)

JIOjosBG: ofc if there is a good reason for this, lets discuss

src/controllers/requests/requests.ts
```diff
@@ -1133,6 +1164,63 @@ export class RequestsController extends EventEmitter implements IRequestsControl
...
+      const existingSafeRejectionRequest = this.visibleUserRequests.find(
+        (userRequest) =>
+          userRequest.id !== request.id &&
+          userRequest.kind === 'calls' &&
+          !!userRequest.signAccountOp.account.safeCreation &&
+          userRequest.meta.accountAddr === accountOp.accountAddr &&
+          userRequest.meta.chainId === accountOp.chainId &&
+          getAccountOpNonce(userRequest.signAccountOp.accountOp) === nonce &&
+          userRequest.meta.isOnchainSafeRejection
+      )
```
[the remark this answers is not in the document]
JIOjosBG: it may also be somewhere else

src/interfaces/userRequest.ts
```diff
@@ -61,6 +61,7 @@ export interface CallsUserRequest extends UserRequestBase<DappPromise[]> {
     activeRouteId?: string
     isSwapAndBridgeCall?: boolean
     topUpAmount?: bigint
+    isOnchainSafeRejection?: boolean
```
JIOjosBG: i am bearish for adding a new meta property we are attempt to rely on. Will it be too complicated to use a pure function that reads the accountOp to decide of it is rejection? This function can and should be used when the reject request is coming from another device via the safe api

src/libs/humanizer/messageModules/safeModule.ts
```diff
@@ -10,17 +10,15 @@ export const safeMessageModule: HumanizerTypedMessageModule = (message: Message)
...
   const { accountAddr } = message
   const { verifyingContract } = message.content.domain
   const humanizedCall = genericErc20Humanizer({ accountAddr }, { to, value, data })
   const safeStandardHumanization = getSafeHumanization(
     verifyingContract ?? undefined,
     to,
     value,
-    data,
-    0,
-    nonce
```
JIOjosBG, line 23: in the other comment i asked to bring this logic back. If you agree, dont forget for this as well

src/libs/safe/safe.ts
```diff
@@ -427,6 +436,7 @@ export function toCallsUserRequest(
             meta: {
               accountAddr: safeAddr,
               chainId: BigInt(chainId),
+              ...(isOnchainSafeRejection && { isOnchainSafeRejection: true }),
```
JIOjosBG: as mentioned above we can get rid of `isOnchainSafeRejection`

src/libs/safe/safe.ts
```diff
@@ -279,7 +281,10 @@ export async function getMessage({
   messageHash: Hex
 }): Promise<ExtendedSafeMessage | null> {
   const apiKit = getApiKit(chainId)
-  const msg = await apiKit.getMessage(messageHash).catch(() => null)
+  const msg = await apiKit.getMessage(messageHash).catch((e) => {
+    console.log('safe message not found', e)
```
JIOjosBG, line 284: forgotten log?

src/controllers/portfolio/portfolio.ts
```diff
@@ -2137,7 +2137,15 @@ export class PortfolioController
...
-  async simulateAccountOp(op: AccountOp): Promise<void> {
+  /**
+   * Always pass account ops that belong to the same account & network
+   */
+  async simulateAccountOp(ops: AccountOp[]): Promise<void> {
+    // typescript protection
+    if (!ops.length) return
+    const op = ops[0]
+    if (!op) return
+
```
JIOjosBG: since we refactored multiple times, and iirc we agreed that we will simulate only 1 acc op, can we make this accept `AccountOp` instead of `AccountOp[]` again?

src/controllers/requests/requests.ts
```diff
@@ -197,6 +204,8 @@ export class RequestsController extends EventEmitter implements IRequestsControl
 
   #currentUserRequest: UserRequest | null = null
 
+  #dismissedSafeNonceConflicts: SafeNonceConflict[] = []
```
JIOjosBG: it will take me a little bit of time to send you a full poposal, but i suggest we remove the banner as it's both buttons did nothing when i tried (it could be due to bug)

edit: i sent the proposal in the ambire-app PR as comment

JIOjosBG (changes requested):

src/libs/safe/safe.ts
Oxbobby: I would say we're just console logging the error that the safe message couldn't be fetched

src/libs/humanizer/modules/Safe/index.ts
Oxbobby: Resolving

src/libs/humanizer/modules/Safe/index.ts
Oxbobby: You are correct, yes. Initially, I though this module is responsible for the "Cancel" humanization we implemented as rejecting on chain a Safe transaction is called a Cancel action in Ambire. However, this isn't the case as this module is responsible for interacting with the Safe app. My mistake, fixing

src/controllers/requests/requests.ts
Oxbobby: this is deleted

src/libs/humanizer/modules/Safe/index.ts
JIOjosBG: i moved the logic back

src/interfaces/userRequest.ts
JIOjosBG: resolved

JIOjosBG:
the keystore is failing in v2
the portfolio test passes locally on this branch

PR #2608, opened 2026-08-10
Validate account preference against account list on boot

PetromirDev:


PetromirDev:
Closing as this shouldn't really happen and is now covered by https://github.com/AmbireTech/ambire-app/pull/7693

PR #2613, opened 2026-08-11
Fix: balance decrease warning -> banner + new design

Oxbobby:
## Summary

Moves significant balance decrease detection into the dynamically calculated transaction banners and prevents the warning from disappearing while portfolio simulation results are being refreshed.

## Motivation

The warning previously depended on `portfolioNetworkState.isLoading`. When an existing simulation result was being refreshed, `isLoading` temporarily became `true`, causing the warning to disappear and then reappear when the request completed.

## Changes

- Moved significant balance decrease detection from `calculateWarnings()` into the `banners` getter.
- Continued using the latest available simulation result while a refresh is in progress.
- Recalculate the banner automatically when a newer simulation result becomes available.
- Return no banner when simulation results are unavailable.
- Wait for trace-call token discovery to finish or fail before evaluating the result.
- Emit a controller update when discovery reaches a terminal state so the final banner state reaches the UI.
- Require both conditions for a significant decrease:
  - at least 20% of the account portfolio value
  - at least $1,000
- Replaced the separate “possibly inaccurate” warning with concise secondary context explaining common false positives such as bridging and liquidity provision.
- Extended banner interfaces with optional `title` and `secondaryText`.
- Added descriptive titles to address-safety, app-verification, and Safe delegate-call banners.
- Added coverage for:
  - missing simulation results
  - retaining a warning during refreshes
  - recalculating from newer results
  - discovery completion and failure
  - updated app-verification banner data


Oxbobby:
First AI comment has been addressed.  
Second is not applicable imo as yes, this gets calculated, but it gets dispatched to the UI rarely

src/controllers/signAccountOp/helper.ts
```diff
@@ -29,11 +29,16 @@ function getTokenUsdAmount(token: TokenResult, gasAmount: bigint): string {
...
 ): Warning | null {
   const portfolioNetworkState = portfolioState?.[chainId.toString()]
 
-  if (portfolioNetworkState && portfolioNetworkState.result && !portfolioNetworkState.isLoading) {
+  // calculate this only after traceCall has ended
+  const isDiscoveryOver =
+    discoveryStatus === TraceCallDiscoveryStatus.Failed ||
+    discoveryStatus === TraceCallDiscoveryStatus.Done
+
+  if (portfolioNetworkState && portfolioNetworkState.result && isDiscoveryOver) {
```
gergana95, line 41: We dropped the loading state here - portfolioNetworkState.isLoading and we show for a second SIGN button which is changed after loading is completed

gergana95 (changes requested):

src/controllers/signAccountOp/helper.ts
gergana95: We discussed, this actually is better, other way is to have this warning blinking every 30s due to portfolioNetworkState.isLoading

gergana95 (approved):

PR #2615, opened 2026-08-11
refactor: move side-panel request routing out of the controllers

sonytooo:
Companion to https://github.com/AmbireTech/ambire-app/pull/7703 — merge this one first (the app PR bumps the submodule to this commit).

## Why

`RequestsController` decided by itself whether a request should be shown in the Chrome side panel. To do that it read the persisted `isSidePanelModeEnabled` flag, called `ui.openSidePanel()`, and tore down request windows it had just created — re-checking the flag at **four** points (before opening, after the overlay cleanup, after window creation, and on focus) to cover the resulting races.

Two problems with that:

- It is web-specific logic in code shared with mobile. `ambire-common` should not know that Chrome has a side panel, that opening one needs a user gesture, or that a "mode" flag exists in extension storage.
- The races it worked around were self-inflicted. `chrome.sidePanel.open()` requires a user gesture, so a dapp-triggered call silently fails — the controller had to create a window anyway and then discard it, which is exactly where the flicker and the "banner does nothing" reports came from.

## What changed

**The surface choice moved into `UiController`, behind a request view with `open`/`focus`/`close`.** It routes to the platform's panel when one is open and to a request window otherwise:

```ts
open:  panel.isOpen() ? null : (remove('popup'), window.open(options))
focus: remove('popup'), window.focus(windowProps, params)
close: window.remove(winId)
```

**`UiManager` gained `panel?: PanelManager`** — a single `isOpen()`, implemented in the extension's `webapi/panel` (see the app PR). The panel is deliberately neither openable nor closable from here: Chrome allows neither without a user gesture, so requests fall back to a request window whenever it is closed. That removes the races rather than handling them.

**`RequestsController` keeps no side-panel knowledge.** Its whole diff vs `v2` is now three hunks (~15 net lines):

- `open`/`focus` go through the request view (the `remove('popup')` calls moved in there too);
- `open()` yielding no window props makes the existing guards do the right thing — `focus` and `close` become no-ops, because the panel is already visible and cannot be closed;
- `closeRequestWindow()` dismisses the active request when there is no window to remove, so the panel reuses the request window's queue semantics (transactions stay queued, everything else is rejected) instead of needing a second entry point. `#handleRequestWindowClose`'s body is untouched.

**Removed:** `ui.openSidePanel`, `requests.dismissActiveRequest` + the `#onActiveRequestDismissed` extraction it forced, the storage-flag reads, and the `String(id)` fallback in `setCurrentUserRequestById` that existed only to paper over the removed races.

**Left in place:** the `'side-panel'` view type, `isExtensionOverlayView` (`dapps`/`main`) and one `isSidePanelView` check each in `transfer`/`swapAndBridge`. Those are view-type semantics of the kind this package already models (`viewType === 'popup'`), not platform machinery. The `swapAndBridge` handler also lost a block that duplicated the fallthrough right below it.

## Tests

`requests` 39/39 (3 new: panel open ⇒ no window; close with no window rejects; close with no window keeps transactions queued), `transfer` 24/24, `swapAndBridge` 29/29.

### DIFF
--- a/src/controllers/swapAndBridge/swapAndBridge.ts
+++ b/src/controllers/swapAndBridge/swapAndBridge.ts
@@ -524,14 +524,13 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
     })
 
     this.#ui.uiEvent.on('removeView', (view: View) => {
-      // A panel outlives the routes rendered in it, so its session must be unloaded on close for
-      // the form to start fresh next time. The popup keeps its form on purpose (see unloadScreen).
-      if (isSidePanelView(view) && this.sessionIds.includes(view.type))
-        this.unloadScreen(view.type, true)
-
       if (!isSwapAndBridge(view.currentRoute)) return
       this.#isOnSwapAndBridgeRoute = false
       this.updateQuoteInterval.stop()
+
+      // A closing panel destroys its screens without unmounting them, so the session has to be
+      // unloaded here for the form to start fresh next time.
+      if (isSidePanelView(view)) this.unloadScreen(view.type, true)
     })
   }
 
--- a/src/controllers/swapAndBridge/swapAndBridge.ts
+++ b/src/controllers/swapAndBridge/swapAndBridge.ts
@@ -524,18 +524,10 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
     })
 
     this.#ui.uiEvent.on('removeView', (view: View) => {
-      if (isSidePanelView(view)) {
-        if (isSwapAndBridge(view.currentRoute)) {
-          this.#isOnSwapAndBridgeRoute = false
-          this.updateQuoteInterval.stop()
-        }
-
-        if (this.sessionIds.includes('side-panel')) {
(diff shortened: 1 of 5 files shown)


src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -524,18 +524,10 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
...
-        if (this.sessionIds.includes('side-panel')) {
-          this.unloadScreen('side-panel', true)
-        }
-
-        return
-      }
+      // A panel outlives the routes rendered in it, so its session must be unloaded on close for
+      // the form to start fresh next time. The popup keeps its form on purpose (see unloadScreen).
+      if (isSidePanelView(view) && this.sessionIds.includes(view.type))
+        this.unloadScreen(view.type, true)
```
PetromirDev: shouldn't this be after the `isSwapAndBridge` check

PetromirDev (changes requested):

Maskln (approved):

PR #2623, opened 2026-08-17
Exclude the Arc testnet (and eventually live chain) operator address from balance calculations as an error is thrown for a missing snapshot

Oxbobby:


superKalo (approved):

superKalo:
Note: we discussed offline that the fix filters the address only out of balance-change queries, not out of the token list fed to `addTokensToBeLearned`, so the pseudo-token is still permanently learned into the portfolio hints.

Context on the better place for this fix:

> On Arc, `getTransferLogTokens(receipt.logs, accountAddr)` returns 0xffff…fffE, `filterStaticBlacklistedAddrs` does not know it, and `this.#portfolio.addTokensToBeLearned(foundTokens, chainId)` (also at line 885) persists it into `hintsController.#toBeLearnedAssets.erc20s[chainId]` - `addTokensToBeLearned` only rejects ZeroAddress and unparseable addresses. From then on every portfolio update for that account wastes an ERC-20 balance read on a non-token, forever. The existing per-chain `STATIC_BLACKLIST.blacklistAddrs` in src/libs/portfolio/blacklist.ts is exactly this mechanism (see the Andromeda entry: \"METIS as an ERC-20 token - Excluded because it's a duplicate of the native token\") and would fix both paths in one place.

We're moving forward with this PR and will improve further in https://github.com/AmbireTech/ambire-app/issues/7728

PR #2625, opened 2026-08-17
feat: accounts sync between the mobile app and the extension

sonytooo:
Business logic for syncing selected accounts and their keys between the Ambire mobile app and the browser extension over animated QR codes. Platform-agnostic — the same controller methods serve both products.

## Added

- **`libs/accountsSync`** — `AccountsSyncPayload` v1 (`{ v, secret, accounts, keys, seeds }`) plus `serializeAccountsSyncPayload` / `parseAccountsSyncPayload`. The payload is gzipped (`pako`) before it is hex encoded for the UR frames, because JSON full of hex strings compresses 3x or better, and inflation is capped at 2MB so a hostile QR code can't balloon the background's memory.
- **`KeystoreController.exportForSync(keyAddrs, includeSeeds)`** — returns the `password` secret entry, the selected `StoredKey`s and (optionally) the recovery phrases they were derived from, all read as stored. Nothing is decrypted, so it works while locked; refuses when the device has no `password` secret.
- **`KeystoreController.importFromSync(payload, password)`** — derives the *other* device's main key from its password, decrypts the seeds and private keys and re-encrypts them under the local main key through the existing `#addSeed` / `#addKeys` / `#addKeysExternallyStored`. Wrong password reuses the existing silent `Incorrect password` path.
- **`AccountsController.getAccountsForSync(addrs)`** — the selected `Account` records, preferences and all.
- **`MainController.exportAccountsForSync` / `importAccountsFromSync`** — one `withStatus`-wrapped method per direction. Export ships the payload through the one-time-data channel (never persisted state); import parses and unlocks the key material *before* `#updateAccounts`, so a wrong password adds neither keys nor accounts. Both accept an optional trailing `requestId` and reply through it, so the UI can await the result.

## Changed / fixed along the way

- `decryptMainKeyWithSecret` extracted into `libs/keystore` (the exact inverse of the existing `encryptMainKeyWithSecret`) and reused by `#unlockWithSecretGCM`, so the unwrap + wrong-password mapping exists once.
- `#addSeed` takes an optional `id` and returns the stored id, so a synced seed keeps the id the other device gave it and the synced keys' `meta.fromSeedId` keeps resolving.
- **Pre-existing bug:** the `isReadyToStoreKeys` setter fired `#addKeys` and `#addKeysExternallyStored` in parallel `void` calls, and both rewrite the whole `keystoreKeys` storage entry — whichever finished last dropped the other's keys. No existing flow queued both at once; syncing does. They now flush sequentially through `#addQueuedKeysAndSeeds` (seeds included), which clears the queues, reports failures via `emitError` instead of an unhandled rejection, and always `emitUpdate`s.
- New onboarding queue `#seedsToAddOnKeystoreReady`, next to the two existing key queues, so an import that happens *before* the device password is set still lands.

## Security notes

- Plaintext key material never enters a QR code — only ciphertexts and the `password` secret entry travel.
- Only the `password` secret is exported; `biometrics` and email-vault secrets are device-bound and excluded.
- Legacy AES-CTR payloads are rejected (the keystore migrates them on unlock, and the app must be unlocked to export).
- The exporter's main key is briefly reconstructed in the importer's memory, used only to decrypt, and never persisted. Export is read-only — nothing is written on the exporting device.
- The blob is offline brute-forceable against a single password; `scrypt` N=131072 is the cost barrier, and both designs warn that the QR code carries sensitive information.

## Tests

`accountsSync.test.ts`, 7 new cases in `keystore.test.ts` and a new `mainAccountsSync.test.ts` with two full device-to-device round trips: subset export, signer works on the importing device, preferences travel, onboarding order (import → set password → keys land), wrong password imports nothing, foreign QR data imports nothing, empty selection exports nothing, double sync doesn't duplicate.


src/controllers/accounts/accounts.ts
```diff
@@ -406,6 +406,10 @@ export class AccountsController extends EventEmitter implements IAccountsControl
     return this.accountStates[addr]?.[chainId.toString()]
   }
 
+  getAccountsForSync(addrs: Account['addr'][]): Account[] {
+    return this.accounts.filter((account) => addrs.includes(account.addr))
+  }
+
```
PetromirDev: imo we don't need a dedicated method for this as it's a simple filter that is also not sync specific

src/controllers/keystore/keystore.ts
```diff
@@ -742,14 +751,15 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+    if (existingEntry) return existingEntry.id
 
     const entropy = extractEntropyFromSeed(seed)
 
     const label = `Recovery Phrase ${this.#keystoreSeeds.length + 1}`
 
+    const isIdTaken = !!id && this.#keystoreSeeds.some((s) => s.id === id)
     const newEntry: StoredKeystoreSeed = {
-      id: generateUuid(),
+      id: isIdTaken || !id ? generateUuid() : id,
```
PetromirDev: do we really need to check for a seed with the same id? #findStoredSeed already checks for the same seed content (perhaps we can also add a quick id check there too) and generateUuid will never generate a colliding uuid

src/controllers/keystore/keystore.ts
```diff
@@ -764,12 +774,49 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+      this.emitError({
+        level: 'major',
+        message:
+          'Something went wrong when saving your keys. Please try again or contact support if the problem persists.',
+        error: error instanceof Error ? error : new Error('keystore: failed to add queued keys')
+      })
+    } finally {
+      this.emitUpdate()
+    }
+  }
```
PetromirDev, line 843: Nice one 👏 Perhaps we can also emitUpdate only once at the end? (pass emitUpdate=false to all private methods)

src/controllers/keystore/keystore.ts
```diff
@@ -1161,6 +1208,151 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+    const syncedSeedIds: { [exportedSeedId: string]: StoredKeystoreSeed['id'] } = {}
+    for (const seed of seeds) {
+      const entropy = await decryptWithKey(exportedMainKey, seed.seed)
+      const seedPassphrase = seed.seedPassphrase
+        ? new TextDecoder().decode(await decryptWithKey(exportedMainKey, seed.seedPassphrase))
+        : null
+
+      syncedSeedIds[seed.id] = await this.#storeSyncedSeed({
+        id: seed.id,
+        seed: reconstructSeedFromEntropy(entropy, seedPassphrase),
```
PetromirDev: should we reuse this logic in some way?

src/controllers/keystore/keystore.ts
```diff
@@ -1161,6 +1208,151 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+    await this.initialLoadPromise
+
+    const { secret, keys, seeds } = payload
+    if (secret.aesEncrypted.cipherType !== CIPHER)
+      throw new Error('keystore: synced main key is not encrypted with GCM')
+
+    const secretKey = await deriveSecret(this.#scryptAdapter, password, secret.scryptParams.salt)
+    // The exporting device's main key. Kept in this scope only, never persisted.
+    const exportedMainKey = await this.#decryptMainKeyWithSecret(secretKey, secret.aesEncrypted)
+
```
PetromirDev: Technically it does get persisted in memory because we import it with `extractable=true`. I don't know why but I 'm a bit worried to have the main key of a device transferred, even encrypted. The reason is that changing the password of the device cannot defend it from a leaked main key. Can't we derive a new key to reencrypt the payload and then use it to decrypt in the new device (instead of leaking the original main key). This shouldn't really be a concern because users will be syncing between their own devices, just sharing so we can discuss it.

PetromirDev (changes requested):

package.json
```diff
@@ -30,6 +30,7 @@
     "events": "^3.3.0",
     "hash-wasm": "^4.12.0",
     "js-yaml": "^4.1.0",
+    "pako": "^2.1.0",
```
PetromirDev: Just fyi https://dev.to/parsajiravand/youre-importing-pako-to-gzip-data-compressionstream-does-it-natively-4d7b. I don't have a strong opinion

PetromirDev (changes requested):

src/controllers/keystore/keystore.ts
```diff
@@ -764,12 +774,49 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+    const externalKeysToAdd = this.#externalKeysToAddOnKeystoreReady
+    this.#seedsToAddOnKeystoreReady = []
+    this.#internalKeysToAddOnKeystoreReady = []
+    this.#externalKeysToAddOnKeystoreReady = []
+
+    try {
+      // Seeds first, so the keys can keep pointing to the seed they were derived from
+      for (const seed of seedsToAdd) {
+        await this.#addSeed(seed)
+      }
```
PetromirDev: addSeed writes to storage

src/controllers/keystore/keystore.ts
```diff
@@ -1161,6 +1208,151 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+   * this fails).
+   */
+  async importFromSync(payload: AccountsSyncPayload, password: string) {
+    await this.initialLoadPromise
+
+    const { secret, keys, seeds } = payload
+    if (secret.aesEncrypted.cipherType !== CIPHER)
+      throw new Error('keystore: synced main key is not encrypted with GCM')
+
+    const secretKey = await deriveSecret(this.#scryptAdapter, password, secret.scryptParams.salt)
```
PetromirDev: Why do we pass only scryptParams.salt? The function should read the secret's params as this will silently break if the extension and mobile app have different configurations

PetromirDev (changes requested):

package.json
sonytooo: Where pako runs:

Call site: main.ts:1720 serialize
Env: extension background SW ✅ / mobile WebView
worker ✅
────────────────────────────────────────
Call site: main.ts:1742 parse
Env: same ✅
────────────────────────────────────────
Call site: useAccountsSyncImport.ts:35 parse
Env: extension UI ✅ / mobile RN Hermes ❌

CompressionStream/DecompressionStream no exist in Hermes. No streams polyfill in repo (checked). So native swap break mobile import — hook parse scanned bytes in RN thread to validate + show account count before password prompt.

src/controllers/keystore/keystore.ts
sonytooo: We discussed offline to leave it as is for now and just to make sure we clean up the decrypted mainKey on the device that imports after successful sync. 

All other solutions that I can think of have some UX downside that we should discuss first before acting on changing this

src/libs/accountsSync/accountsSync.ts
```diff
@@ -0,0 +1,165 @@
...
+  if (!secret || secret.id !== 'password')
+    throw new Error('accountsSync: missing the password protected main key')
+
+  const { salt, N, r, p, dkLen } = secret.scryptParams || {}
+  const hasValidScryptParams =
+    typeof salt === 'string' &&
+    typeof N === 'number' &&
+    typeof r === 'number' &&
+    typeof p === 'number' &&
+    typeof dkLen === 'number'
```
superKalo: This only type-checks the scanned scrypt params; importFromSync passes them straight to ScryptAdapter.scrypt, so a hostile QR (N: 2**21, r: 8, p: 64) makes the background allocate GBs and spin with no abort - same hostile-QR class the 2 MB gzip-bomb guard already covers. I wonder if we can defend here as well checking `MAX_DECOMPRESSED_PAYLOAD_SIZE`? 👀

src/libs/accountsSync/accountsSync.ts
```diff
@@ -0,0 +1,165 @@
...
+const validateAccounts = (accounts: any) => {
+  if (!Array.isArray(accounts) || !accounts.length)
+    throw new Error('accountsSync: no accounts in the payload')
+
+  accounts.forEach((account) => {
+    if (!account || !isAddress(account.addr)) throw new Error('accountsSync: invalid account addr')
+    if (!Array.isArray(account.associatedKeys))
+      throw new Error('accountsSync: invalid account associatedKeys')
+    if (!account.preferences?.label) throw new Error('accountsSync: invalid account preferences')
+  })
```
superKalo, line 109: Great! Reminds me that this does not verify creation (factoryAddr/bytecode/salt) against addr, nor initialPrivileges/associatedKeys contents similar to the checks we are doing in the Account Picker controller. I wonder if this is a possible attack vector or not 🤔

package.json
superKalo: Side note, let's pin the v here.

PR #2628, opened 2026-08-18
Fix: request after account switch should not end with JSON-RPC error

Oxbobby:
We have this bug:
1. I'm connected to an app (safe global) with an account
2. But the selectedAccount in the extension is a different one than the one I'm connected to the app with
3. I initiate a request from the app
4. The extension opens an account switch request correctly
5. The moment I switch the account (it happens successfully on the extension side), I receive an "An internal error was received. Details: Internal JSON-RPC error. Version: viem@2.52.2" on the app side

Root cause: after a successful account switch, the already-built transaction was unnecessarily refreshing account state again. If that refresh returned no data, the app promise was rejected with JSON-RPC error -32603

The fix resumes the existing transaction without that redundant refresh, while preserving:
* Account-address matching before continuation
* State validation for genuinely new transactions
* Normal rejection behavior

### DIFF
--- a/src/controllers/requests/requests.test.ts
+++ b/src/controllers/requests/requests.test.ts
@@ -520,6 +520,83 @@ describe('RequestsController ', () => {
     expect(controller.userRequestsWaitingAccountSwitch).toHaveLength(0)
     expect(controller.userRequests).toHaveLength(0)
   })
+  test('resolving an account switch continues with the already built transaction request', async () => {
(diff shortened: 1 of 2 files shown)


Oxbobby:
The test errors are unrelated and are because of bad RPC

Oxbobby:
closing in favour of https://github.com/AmbireTech/ambire-common/pull/2629#pullrequestreview-4960500059

PR #2629, opened 2026-08-18
Hotfix/ `new Date().getTime()` ids are not reliable and conflict in the Requests controller

PetromirDev:
## How to reproduce:
Follow these steps https://github.com/AmbireTech/ambire-common/pull/2628

## Why:
Because both requests' ids are constructed in the same tick. I came to that conclusion by inspecting the logs and seeing that the switch account and sign message requests have the same id. It was resolved immediately after appending `-switch-account` to one of the requests.

as you can see `kind=message` and `kind=switchAccount` have exactly the same id, excluding the suffix.


Oxbobby (approved):
🚀

Oxbobby:
Checked the failing test - it's because of a broken RPC, moving on

PR #2641, opened 2026-08-24
Feature/wallet features

Oxbobby:
## Summary

Adds the shared business logic required for native WALLET staking support and stake-based Swap & Bridge fees.

The portfolio can now load the current xWALLET share value, while Swap & Bridge dynamically calculates the Ambire fee from the account’s stkWALLET holdings.

## What changed

- Added an Ethereum xWALLET share-value loader with:
  - one-hour caching;
  - deduplication of concurrent requests;
  - stale-value fallback when refreshes fail;
  - validation against invalid or zero conversion values.
- Exposed the share value through the portfolio and selected-account state.
- Added a provider-controller method for requesting the current share value from the UI.
- Added shared helpers for converting and formatting xWALLET amounts as their underlying WALLET value.
- Added conversion information to stkWALLET unwrap humanizations while preserving it alongside ERC-7730 visualizations.
- Added dynamic Swap & Bridge fee tiers based on the USD value of stkWALLET held:
  - 0.50% up to 33,000 stkWALLET;
  - 0.40% above 100,000 stkWALLET;
  - 0.25% from 700,000 stkWALLET;
  - 0% from 700,000 stkWALLET.
- Propagated the calculated fee through LiFi, Socket, Socket v3, Squid, and Uniswap quotes.
- Added explicit fee-exemption reasons for:
  - wrap and unwrap operations;
  - fee-exempt tokens;
  - routes where fee collection is unavailable.
- Exposed the effective fee and exemption reason through the Swap & Bridge controller state.

## Test plan

- [ ] Verify the xWALLET share value is loaded and exposed through the portfolio.
- [ ] Verify successful values are cached and concurrent requests are deduplicated.
- [ ] Verify stale values remain available after a refresh failure.
- [ ] Verify every stkWALLET balance threshold produces the expected fee percentage.
- [ ] Verify the dynamic fee is forwarded to every supported swap provider.
- [ ] Verify wrap, unwrap, exempt-token, and unavailable-fee routes do not include an Ambire fee.


src/controllers/portfolio/portfolio.ts
```diff
@@ -188,6 +193,53 @@ export class PortfolioController
    */
   protected hints: HintsController
 
+  #loadWalletStakingShareValue(
```
JIOjosBG: the portfolio is already big as is, i hope we find a way to not have this in the portfolio either by 1) removing the label in the dahsboard that depends on this 2) moving this to the UI

Oxbobby: I will work on it

Oxbobby: I will move this to a separate controller for the time being. I like this being shown on the dashboard. If we decide later that this will be shown at only 1 place at the dashboard and no where else, we will remove the new controller and move the logic to the UI

PR #2642, opened 2026-08-24
fix(swap-and-bridge): base slippage warning percentage on input value, not quote

Maskln:
closes https://github.com/AmbireTech/ambire-app/issues/7771

before:

Screenshot 2026-08-20 at 18 49 51

after:
Screenshot 2026-08-24 at 11 58 30
-

## Summary
- `calculateAmountWarnings` computed `possibleSlippage` (the % shown in the "Critical slippage warning" / "Warning! This route has a higher slippage than usual" modal title) relative to `outputValueInUsd` (the quote), while the paired `estimatedLossUsd` dollar figure shown right below it was always relative to `inputValueInUsd`. When a quote already reflected heavy price impact, this let the modal show a misleadingly tiny percentage (e.g. `0.07%`) next to a large dollar loss (e.g. `$46,091`), since the quote-to-floor gap can be small even when the input-to-floor loss is not.
- The `$50` noise-filter in `isElevatedSlippage` had the same issue: it gated on the quote-relative `diffBetweenQuoteAndMinAmount` instead of the already-computed, input-relative `slippageLossUsd`, which could silently suppress a legitimate warning whenever price impact had already eaten into the quote-to-floor gap.

Both now use the same input-relative basis as the dollar loss that's displayed alongside them.

## Test plan
- [x] `npm run jest -- src/libs/swapAndBridge/swapAndBridge.test.ts` — 29/29 passing, including two new regression tests for the extreme-branch percentage mismatch and the noise-filter false negative
- [x] Updated the one pre-existing test whose assertion depended on the old (incorrect) quote-relative percentage

### DIFF
--- a/src/libs/swapAndBridge/swapAndBridge.test.ts
+++ b/src/libs/swapAndBridge/swapAndBridge.test.ts
@@ -445,7 +445,9 @@ describe('swapAndBridge lib', () => {
       expect(result).not.toBeNull()
       expect(result?.type).toBe('slippageImpact')
       if (result?.type === 'slippageImpact') {
-        expect(result.possibleSlippage).toBeGreaterThan(81.8)
+        // possibleSlippage is relative to inputValueInUsd (100), not the more favorable
+        // outputValueInUsd (110), so it doesn't understate the risk of the low minAmountOut
+        expect(result.possibleSlippage).toBeCloseTo(80, 5)
         expect(result.minInUsd).toBe(20)
       }
     })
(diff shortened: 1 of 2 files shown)


PetromirDev (approved):
Makes sense. Two things:
1. Let's have @Oxbobby check this too as he is more knowledgeable
2. I think it should land in next/prod and not v2

Oxbobby (approved):
Let's prepare this as a hotfix, yes

Maskln:
beacause of ths PR: https://github.com/AmbireTech/ambire-common/pull/2643

PR #2644, opened 2026-08-24
add(humanizer): uniswap for internal swap and bridge

JIOjosBG:
Self explantory

## To test
- go to the internal swap and bridge
- pick tokens
- select an uniswap route
- check the humanization of the tx


Maskln (approved):
Good job 👏

PR #2646, opened 2026-08-25
Fix: allow gs013 errors to enter the retry loop

Oxbobby:
The bug is very complex as it's not a bug on Ambire's side only. You have this sequence on a Safe account:

1. I go to a dapp with a Safe account, it's 2/2 threshold
2. I quickly sign with both signers, they are imported in the extension, and broadcast with an EOA
3. The dapp then sends a second request
4. I quickly sign with both signers and attempt to broadcast the transaction, but the RPC returns a "GS013" error

After debugging the GS013 error, I found out that the RPC's onchain read of the Safe account's nonce is returning the old nonce, the one before the current transaction. Hence, the Safe contract is returning invalid signatures.

Well, the signatures are valid, and the nonce is correct as it's the last one - just the RPC doesn't know about it yet. The error goes away by itself in due time, we just have to wait it out for the RPC to get the correct latest state read

### DIFF
--- a/src/libs/broadcast/broadcast.test.ts
+++ b/src/libs/broadcast/broadcast.test.ts
@@ -167,7 +167,7 @@ describe('broadcast', () => {
       )
     ).rejects.toThrow('Failed estimating gas for broadcast')
 
-    expect(provider.send).toHaveBeenCalledTimes(11)
+    expect(provider.send).toHaveBeenCalledTimes(10)
   })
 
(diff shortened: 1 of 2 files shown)


PetromirDev (approved):
The error code is too broad (covers all txn failures), but we agreed that users will not reach this far if the transaction is failing (because the estimation will prevent it). 
Users waiting more time on broadcast, but the transaction will fail anyway (very rare) < Blocking successful transactions because of nonce issues (easy to reproduce)

PR #2658, opened 2026-08-28
Feature / Make it possible to mark dApps as trusted

jordan-enev:
Add: Make it possible to mark dApps as trusted, so the user can silence the warning.


src/controllers/dapps/dapps.ts
```diff
@@ -1202,6 +1220,74 @@ export class DappsController extends EventEmitter implements IDappsController {
...
+    this.emitUpdate()
+
+    await this.#persistTrustedDapps()
+  }
+
+  async #persistTrustedDapps() {
+    await this.#storage.set(
+      'trustedDapps',
+      Array.from(this.#trustedDapps.entries()).map(([id, addedAt]) => ({ id, addedAt }))
+    )
```
superKalo, line 1288: All the rest options/settings per dapp get reset when the dapp is disconnected (e. g. the address you connect to it).

So I think this user "trusted" mark per dapp should also be part of the dapp storage item and reset upon disconnect.

superKalo (approved):

src/controllers/dapps/dapps.ts
jordan-enev: Refactored it, not it's part of the dapps storage.

PR #2660, opened 2026-08-28
7730 refactoring/ controller

JIOjosBG:
For more info read the description in the app https://github.com/AmbireTech/ambire-app/pull/7826

builds on top of: https://github.com/AmbireTech/ambire-common/pull/2659


src/controllers/erc7730/erc7730.ts
```diff
@@ -0,0 +1,533 @@
...
+ *
+ * So everything external - the relayer, the RPC reads for Safe proxies, the cache, the request
+ * dedup and the persistence - lives here.
+ */
+export class Erc7730Controller extends EventEmitter {
+  #storage: IStorageController
+
+  #callRelayer: BindedRelayerCall
+
+  #sendUiMessage: SendUiMessage
```
Oxbobby: We have an UI controller (ui.ts) for sending messages to the UI. Usually, we pass it in the constructor and the call its method:
```
this.#ui.message.sendUiMessage(params)
```
Not crucial, but long term I think it's better to use the controller to be sure all UI message go through the place of central dispatching

src/controllers/erc7730/erc7730.ts
```diff
@@ -0,0 +1,533 @@
...
+
+  #getProvider: (chainId: bigint) => SafeSingletonProvider | undefined
```
Oxbobby: Similarly, I think we should pass directly the providers controller here instead of a method to retrieve them. This also makes me think that we don't need the "SafeSingletonProvider" and we can move on to delete it throughout the codebase

src/controllers/erc7730/erc7730.ts
```diff
@@ -0,0 +1,533 @@
...
+
+    this.#storage = storage
+    this.#callRelayer = callRelayer
+    this.#sendUiMessage = sendUiMessage
+    this.#getProvider = getProvider ?? (() => undefined)
+
+    this.#initialLoadPromise = this.#load()
+  }
+
+  async #load() {
```
Oxbobby, line 130: We will need to do:
```
await this.#providers.initialLoadPromise
```
because we will have the providers as a dep now

src/controllers/erc7730/erc7730.ts
```diff
@@ -0,0 +1,533 @@
...
+    void this.#persist()
+
+    return known
+  }
+
+  async getDescriptorsForAccountOp(accountOp: AccountOp): Promise<Erc7730CallDescriptors> {
+    try {
+      const known = await this.#gather((state) => planErc7730Wants(accountOp, state))
+
+      return resolveErc7730Descriptors(accountOp, known)
```
Oxbobby, line 482: ERC7730 is a monster and therefore, I can only briefly follow the logic.  
Here, we do `planErc7730Wants` which calls `resolveErc7730Call`.  
Then, we call `resolveErc7730Descriptors` which calls again `resolveErc7730Call`. 

Let's check (if we have time) if a possible optimization here exists

src/controllers/erc7730/erc7730.ts
```diff
@@ -0,0 +1,533 @@
...
+      // Nothing usable was fetched, but built-in descriptors need no fetching at all
+      return resolveErc7730Descriptors(accountOp, EMPTY_ERC7730_KNOWN)
+    }
+  }
+
+  async getDescriptorForMessage(message: Message): Promise<Erc7730ResolvedDescriptor | null> {
+    try {
+      const known = await this.#gather((state) => planErc7730MessageWants(message, state))
+
+      return resolveErc7730MessageDescriptor(message, known)
```
Oxbobby, line 500: Same as https://github.com/AmbireTech/ambire-common/pull/2660/changes#r3959080774

src/controllers/erc7730/erc7730.ts
JIOjosBG: i will look into it later today

Oxbobby:
Uniswap bridge through across: here, we're displaying the network icon but it's huge

Screenshot 2026-09-10 at 8 40 59

PR #2661, opened 2026-08-29
Fix: add overhead to eoa 7702; make the overhead 20%

Oxbobby:
There was a bug that skipped adding the gas overhead to EOA 7702 native broadcasts. Along with that, we're changing the gas overhead to 20%


Oxbobby:
@Ivshti agreed to move forward with this

PR #2664, opened 2026-08-31
Feature: quick to token list

Oxbobby:
## Problem

We did #fetchFromAll(), which awaits each provider with a mix of sophisticated logic that might retry a request if conditions met. That kind of logic is appropriate for quote() requests but not for token fetches.

## Solution

We still request the tokens from each provider but we update the UI immediately when one of them finishes. When the second one finishes, we again update the UI until all of them are done


PetromirDev (approved):
Good job and very much needed (200ms for 2 providers and 5 seconds for uniswap...)

Image

PR #2666, opened 2026-08-31
Update scheduled updates logic in portfolio to enable pending defi update banner

PetromirDev:
## Changes:
- Trigger a force update on manual refresh IF there is a pending update (otherwise the user sees a banner for 1min and refreshing does nothing)
- Exposes scheduled updates to the UI
- Adds isRunning to scheduled updates to handle the case where a new txn comes in while the portfolio is being updated (a new update is needed)
- Adds additional unit tests

### DIFF
--- a/src/controllers/portfolio/portfolio.ts
+++ b/src/controllers/portfolio/portfolio.ts
@@ -5,7 +5,11 @@ import {
   RecurringTimeout
 } from '../../classes/recurringTimeout/recurringTimeout'
 import { STK_WALLET } from '../../consts/addresses'
-import { BLACKLIST_UPDATE_INTERVAL } from '../../consts/intervals'
+import {
+  BLACKLIST_UPDATE_INTERVAL,
+  SCHEDULED_PORTFOLIO_UPDATE_DELAY,
+  SCHEDULED_PORTFOLIO_UPDATES_RUNNER_INTERVAL
+} from '../../consts/intervals'
 import {
   Account,
   AccountId,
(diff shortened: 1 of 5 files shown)


src/controllers/portfolio/portfolio.ts
```diff
@@ -1968,6 +2022,17 @@ export class PortfolioController
...
+          // The manual update just fetched what the scheduled one would have (it forces a defi
+          // refetch when an update is scheduled), so the pending entry is no longer needed. A
+          // running one is kept, as its request is already in flight.
+          if (isManualUpdate && isSuccessful)
+            this.#forgetScheduledUpdates(
+              accountId,
+              (this.#scheduledUpdates[accountId] || []).filter(
+                (update) => !update.isRunning && update.chainId === network.chainId
+              )
+            )
```
Oxbobby: I love the new logic. I read it until now so it's possible I'm not comprehending it to the fullest. I'm running on this assumptions:  
We're pushing a scheduled update after the latest chained transaction that will check for DeFi positions at least 1 minute after the scheduled time

If this is how it's working, I think the logic is perfect. That's why I think:
* on isManualUpdate, if we have such logic scheduled, do not fetch DeFi positions. If we don't and the timer/cache has reset, fetch
* this.#forgetScheduledUpdates should be removed from here

My thinking is: if we have a flawless logic for fetching DeFi postions, why allow the user to break it? If he impatiently clicks on the manual refresh, he might fetch the DeFi positions a bit too early. We have an hourglass on the FE, he will wait bit. That bastard

Oxbobby (approved):
🚀

PR #2667, opened 2026-08-31
Reject portfolio updates if the data is from a block, older than our thresholds

PetromirDev:
Read the issue. The only thing I'm worried about is if RPC providers, and especially invictus, can return a lower block number for a short duration. E.g., 100 -> 101 -> 100 -> 102. I've added a threshold for this reason, but we can remove it if that is not applicable.

Closes https://github.com/AmbireTech/ambire-app/issues/7313


Oxbobby (approved):
Tested, everything works as it should 🚀

PR #2671, opened 2026-09-01
Feature/cowswap

Oxbobby:
Implement CowSwap as another API provider


src/services/cowswap/api.ts
```diff
@@ -0,0 +1,819 @@
...
+}) => {
+  const priceUSD = Number(toAsset.priceUSD || 0)
+  if (!priceUSD) {
+    return inputValueInUsd * (Number(toAmount) / Number(buyAmountBeforeFees))
+  }
+
+  return Number(formatUnits(toAmount, toAsset.decimals)) * priceUSD
+}
+
+export class CowSwapAPI implements SwapProvider {
```
PetromirDev, line 210: The first ~170 rows can be moved elsewhere imo

src/services/cowswap/api.ts
```diff
@@ -0,0 +1,819 @@
...
+    if (!this.areChainsSupported({ fromChainId, toChainId })) {
+      throw new SwapAndBridgeProviderApiError(
+        'The requested network pair is not supported by our service provider CoW Swap.'
+      )
+    }
+
+    // TODO: fix this, especially in a setup where CowSwap is the only provider
+
+    // CoW Swap does not expose a token-list endpoint. The shared token picker is populated by
+    // the other providers, while CoW Swap can quote any selected supported ERC-20 token.
```
PetromirDev, line 332: I found this https://files.cow.fi/tokens/CowSwap.json

src/services/cowswap/api.ts
```diff
@@ -0,0 +1,819 @@
...
+  }
+
+  async #getSettlementTransaction(chainId: number, orderUid: string) {
+    const params = new URLSearchParams({ orderUid })
+    const response = await this.#fetchWithTimeout(
+      this.#getApiUrl(chainId, `/trades?${params.toString()}`),
+      { headers: this.#headers }
+    )
+    const trades = await this.#parseResponse<CowSwapTrade[]>(
+      response,
```
PetromirDev, line 775: `/api/v1/trades` is marked deprecated in CoW's openapi spec and `/api/v2/trades` is the paginated
replacement.

src/services/cowswap/api.ts
```diff
@@ -0,0 +1,819 @@
...
+      throw new SwapAndBridgeProviderApiError(
+        'Unable to check the CoW Swap order because its details are missing.'
+      )
+    }
+
+    let order = await this.#getOrder(fromChainId, routeId)
+    if (!order) {
+      if (rawRoute.isEthFlow) return { status: null }
+      const wasSubmitted = await this.#submitOrder(fromChainId, rawRoute, routeId)
+      if (!wasSubmitted) return { status: null }
```
PetromirDev, line 803: The order isn't POSTed until the first status poll, and that poll is up to 30s out

src/services/cowswap/api.ts
```diff
@@ -0,0 +1,819 @@
...
+      signature: '0x',
+      from: owner,
+      quoteId: quoteResponse.id ?? null
+    }
+    const orderUid = computeOrderUid({ chainId: fromChainId, order, owner, isEthFlow })
+    const normalizedFromAsset = convertPortfolioTokenToSwapAndBridgeToToken(fromAsset, fromChainId)
+    const protocol = { name: 'CoW Swap', displayName: 'CoW Swap', icon: '' }
+    const serviceTime = 30
+    const inputValueInUsd = Number(getTokenUsdAmount(fromAsset, sellAmount) || 0)
+    const outputValueInUsd = getOutputValueInUsd({
```
PetromirDev, line 506: serviceTime=30 also slows down the status checks, making the swap seem slower than it is

PetromirDev: Image

PetromirDev (changes requested):

src/services/cowswap/api.ts
Oxbobby: uniswap.ts has the same method and it's fair to say that it's API based. Where would you suggest to move this?

src/services/cowswap/api.ts
Oxbobby: reducing to 10

src/services/cowswap/api.ts
Oxbobby: replacing

src/services/cowswap/api.ts
Oxbobby: We're solving this one with a serviceTime of 10 and now, a restart of the interval on a cowswap success

src/services/cowswap/api.ts
PetromirDev: The best practice would be to make a types.ts and helpers.ts file IMO (same for uniswap)

Oxbobby: moving them 👍

PR #2672, opened 2026-09-01
Humanizer: enhancements in approvals

JIOjosBG:
note: Security fix only, not really noteworthy 

- uppercase calldata should not enter the extension
- txns that are shorter than expected should be handled with warnings because they execute onchain
- morpho embedded multicall warnings


JIOjosBG:
the only failing test fail on v2 as well

PR #2675, opened 2026-09-02
Refactor/ Ensure user receives feedback if estimation gets stuck or fails unexpectedly

PetromirDev:
## Changes:
- Estimations can never be stuck in not initialized state
- Failures are classified as retryable or not so we don't retry estimating if an account/network is missing
- Updated error messages
- A retriable failure no longer throws away an estimation you could still broadcast with - you keep the fee options and get a warning that they may be stale.
- The success path now computes everything before it stores anything, so a failure halfway through can't leave a half-applied estimation behind.
- Retry actually retries: it resumes a loop that had given up, rather than restarting an interval that shuts itself down again.


src/controllers/estimation/estimation.ts
```diff
@@ -18,7 +17,38 @@ import { isPortfolioGasTankResult } from '../../libs/portfolio/helpers'
...
 import { getIsViewOnly } from '../../utils/accounts'
 import EventEmitter from '../eventEmitter/eventEmitter'
-import { EstimationStatus } from './types'
+import { EstimationFailureKind, EstimationStatus } from './types'
+
+type EstimationFailure = {
+  /** Shown to the user as the reason the transaction could not be prepared. */
+  message: string
+  kind: EstimationFailureKind
+}
```
Oxbobby: we can move this to './types'

src/controllers/estimation/estimation.ts
```diff
@@ -306,22 +414,10 @@ export class EstimationController extends EventEmitter {
...
-            : 'Automatically retrying in a few seconds. Please wait...'
-        }`
-      })
-
-      return errors
-    }
-
     if (!this.isInitialized()) return []
 
+    if (this.isRetryingFailure() && !this.estimation) return errors
```
Oxbobby: minor improvement: empty array better as it's more readable

Oxbobby (approved):
Great work, left two minors but we can proceed without them as well. Up to you

Oxbobby:
@PetromirDev also, look at the unit tests, there are some missing properties there

PR #2676, opened 2026-09-03
Feature / add and hide nft

gergana95:
## NFT hiding and custom NFTs — portfolio and hints

Business-logic half of add/hide NFT. Pairs with the app PR; neither works alone.

Before this, the portfolio could only hide or track a whole collection, and
nothing could check whether an address is really an NFT collection or whether
the account owns a given collectible.

**NFTs are identified by their id now.** `CustomToken` and `TokenPreference`
carry an optional `tokenId`, and `getAssetPreferenceId` decides identity
everywhere — so one collectible never collides with another, with its whole
collection, or with the same collection on another chain. Records with no id
keep their old meaning: the whole collection.

**Visibility is per collectible.** `getVisibleCollectibles` hides one NFT
without hiding the rest, and shows only the ids the user added for a collection
they added by hand.

**Two new checks before adding.** `validateERC721Token` (is this a collection?)
and `validateCollectibleOwnership` (does the account own it?). ERC-1155 is
rejected deliberately — some expose `ownerOf` and would pass a duck-typing
check. A contract that declares ERC-721 is believed even if it answers
`decimals()`, because vote-escrowed NFTs (veAERO, veVELO) do exactly that.
Rejections return a reason code; the wording lives in the UI so it stays
translatable.

### Testing

`helpers.test.ts` 23 → 67 tests, `hintsController.test.ts` 24 (merged with v2's
metadata suite), 56 hardhat. Verified against mainnet: enumerable collections,
the lying-enumerable case, unowned ids, an ENS id held by the NameWrapper, a
non-NFT contract. Each fix confirmed by reverting it and watching its test fail.


contracts/compiled/NFTGetter.json
```diff
@@ -1,415 +1,415 @@
...
-      "inputs": [
+    "abi": [
         {
-          "internalType": "contract IAmbireAccount",
-          "name": "account",
-          "type": "address"
-        },
-        {
-          "internalType": "contract NFT[]",
-          "name": "collections",
```
PetromirDev: d

src/libs/portfolio/helpers.ts
```diff
@@ -281,6 +283,259 @@ export const validateERC20Token = async (
...
+  hiddenIds
+}: {
+  collectibles: bigint[]
+  customIds?: bigint[]
+  hiddenIds?: bigint[]
+}) => {
+  // No hidden ids stands for the whole collection, the same way it does in the hints
+  if (hiddenIds && !hiddenIds.length) return []
+
+  const added = customIds?.length
```
PetromirDev, line 370: This filters a discovered collection down to the ids the user added, not just a collection that exists only because they added it. Own #1 and #2 of a collection the API already knows, add #5 as custom, and the getter returns all three - then this drops #1 and #2.

src/controllers/hintsController/hintsController.ts
PetromirDev: On line 544, `if (tokenId) toBeLearnedAssets[collectionAddress].push(tokenId)` - `tokenId` is a `bigint` so `0n` is falsy and never gets pushed, same for the skip check on line 532. The line above has already set `toBeLearnedAssets[collectionAddress] = []`, and an empty array is the enumerable marker everywhere downstream, so `mergeERC721s` marks the collection enumerable and then `mergeCollectionHints` drops the custom `[0n]` on its `knownIds && !knownIds.length` early return.

Adding NFT #0 of a non-enumerable collection validates, previews, toasts "Added the NFT to your portfolio" and then never shows up, and the collection stays marked enumerable from then on. Plenty of collections are 0-indexed. `typeof tokenId === 'bigint'` in both places. This predates the PR but the Add NFT flow is what makes an arbitrary id reachable.

(Commenting on the file - both lines are outside the diff hunks.)

contracts/deployless/NFTGetter.sol
```diff
@@ -51,44 +112,27 @@ contract NFTGetter is Simulation {
...
+      // Optional metadata, missing on collections like the ENS names one
+      try collection.name() returns (string memory name) {
+        info.name = name;
+      } catch {}
+      try collection.symbol() returns (string memory symbol) {
+        info.symbol = symbol;
+      } catch {}
     }
 
     uint balance = collection.balanceOf(address(account));
```
PetromirDev: `balanceOf` here and `supportsInterface` on the next line are unguarded, but the new validator accepts contracts that have neither - there are tests named "is a collection even when balanceOf is missing" and "is a collection without ERC-165 support". Such a contract passes validation, previews, toasts, then reverts the whole getter, gets dropped on `error !== '0x'` and the user sees nothing and no error.

The comment on line 131 says the fallback ids are "not bound by balanceOf, which some collections don't report correctly" - but balanceOf is called before that fallback can run. Let's wrap both in `try/catch`, balance defaulting to `limit` and `supportsInterface` to false.

contracts/deployless/NFTGetter.sol
```diff
@@ -43,6 +43,67 @@ contract NFTGetter is Simulation {
...
+    uint256[] memory ids = new uint256[](balance);
+    uint owned;
+
+    for (uint i = 0; i != balance; i++) {
+      // A collection can report the enumerable interface and still not
+      // implement it, so a failure must not discard the whole collection
+      try collection.tokenOfOwnerByIndex(address(account), i) returns (uint tokenId) {
+        ids[owned] = tokenId;
+        owned++;
+      } catch {
```
PetromirDev: Breaking here leaves whatever ids were already collected, and because `info.nfts.length` is then non-zero the known-ids fallback below never runs. A collection that answers `tokenOfOwnerByIndex` for the first few indices and then reverts gives a silently truncated list - only one that fails at index 0 gets the fallback. Should a partial enumeration fall through and merge with the hinted ids instead?

src/controllers/portfolio/portfolio.ts
```diff
@@ -949,31 +961,137 @@ export class PortfolioController
...
+   * verdict, which the UI asks for after a network problem.
+   */
+  async updateCollectionValidation(
+    collection: { address: TokenResult['address']; chainId: TokenResult['chainId'] },
+    accountId: AccountId,
+    shouldRefetch?: boolean
+  ) {
+    await this.initialLoadPromise
+
+    const key = getAssetCacheKey(collection.address, collection.chainId)
```
PetromirDev: This key is identical to the one `updateTokenValidation` uses on line 951, so the two checks share one `#assetValidationsInProgress` entry and block each other. Both add-asset sheets are mounted at once on the Custom assets screen, so pasting the same address into the Tokens tab and then the NFTs tab makes whichever check starts second return early - no verdict stored, no `emitUpdate`, and that sheet spins until the 15s timeout. Let's prefix the key with the standard.

src/libs/portfolio/helpers.ts
```diff
@@ -281,6 +283,259 @@ export const validateERC20Token = async (
...
+  if (typeof owner !== 'string' || owner.toLowerCase() !== accountId.toLowerCase())
+    return invalid('validation', 'collectible-not-owned')
+
+  return { isValid: true, standard: 'erc721', error: { message: null, type: null } }
+}
+
+/** An ERC-20 token is rejected too, as it also exposes name() and balanceOf() */
+export const validateERC721Token = async (
+  collection: { address: string; chainId: bigint },
+  accountId: string,
```
PetromirDev, line 471: `accountId` is never read in here - `validateCollectibleOwnership` is the one that needs an account. Let's drop the parameter and the argument at both call sites.

src/controllers/portfolio/portfolio.ts
```diff
@@ -949,31 +961,137 @@ export class PortfolioController
...
+        allNetworks
+          ? {
+              allNetworks: this.#networks.networks,
+              allProviders: this.#providers.providers,
+              enableNetworkDetection: true
+            }
+          : undefined
+      )
+      const { isValid, standard, error } = result
+      const validatedStandard = standard === 'erc721' ? 'erc721' : 'erc20'
```
PetromirDev, line 980: `validateERC20Token` has a single return and it is always `standard: 'erc20'`, so the `'erc721'` side can never fire. Let's just store under `'erc20'`.

src/libs/portfolio/helpers.ts
```diff
@@ -281,6 +283,259 @@ export const validateERC20Token = async (
...
+    return undefined
+  }
+
+  const [code, supportsERC721, supportsERC1155, decimals, name, symbol] = await Promise.all([
+    provider.getCode(collection.address).catch((e: any) => handleError(e, 'code')),
+    metadata.supportsInterface!(ERC721_INTERFACE_ID).catch((e: any) =>
+      handleError(e, 'supportsInterface')
+    ),
+    metadata.supportsInterface!(ERC1155_INTERFACE_ID).catch((e: any) => {
+      if (isNetworkError(e)) hasNetworkError = true
```
PetromirDev: This catch and the `decimals()` one below re-implement `handleError` inline and skip its `console.error`, so the two calls whose failure actually flips the verdict are the two that log nothing. Let's use `handleError` for all four.

src/libs/portfolio/portfolio.ts
```diff
@@ -386,6 +388,11 @@ export class Portfolio {
     const isValidToken = (error: TokenError, token: TokenResult): boolean =>
       error === '0x' && !!token.symbol
 
+    // A collection is labeled by its name or its address, so a missing symbol is
+    // no evidence that the address isn't one. Only the getter erroring is, and it
+    // does error for an address that holds no collection.
+    const isValidCollection = (error: TokenError): boolean => error === '0x'
```
PetromirDev, line 394: `isValidCollection` is only used for `collectionErrors` on line 699 - the collections reducer inlines the same check as `error !== '0x'` on line 494, with a near-duplicate comment. Let's use the helper in both.

src/controllers/hintsController/hintsController.ts
```diff
@@ -191,27 +190,83 @@ export class HintsController extends EventEmitter {
...
+    if (typeof collectible.tokenId === 'bigint') {
+      delete learnedNfts[`${checksummed}:${collectible.tokenId}`]
+    } else {
+      Object.keys(learnedNfts)
+        .filter((key) => key.startsWith(`${checksummed}:`))
+        .forEach((key) => delete learnedNfts[key])
+    }
+
+    await this.#storage.set('learnedAssets', this.#learnedAssets)
+  }
```
PetromirDev, line 239: Interesting and nice catch. We don't remove custom tokens from learnedAssets when a token is removed. Should we do it for tokens too?

src/controllers/portfolio/portfolio.ts
```diff
@@ -949,31 +961,137 @@ export class PortfolioController
...
+    collection: { address: TokenResult['address']; chainId: TokenResult['chainId'] },
+    accountId: AccountId,
+    shouldRefetch?: boolean
+  ) {
+    await this.initialLoadPromise
+
+    const key = getAssetCacheKey(collection.address, collection.chainId)
+    // A verdict about the contract itself doesn't change for an address
+    if (this.validTokens.erc721[key] && !shouldRefetch) return
+    if (this.#assetValidationsInProgress.has(key)) return
```
PetromirDev, line 1007: Let's add the standard to the key just in case

src/libs/portfolio/portfolio.ts
```diff
@@ -246,11 +248,11 @@ export class Portfolio {
       ...gasTankFeeTokens.filter((x) => x.chainId === this.network.chainId).map((x) => x.address)
     ]
 
-    hints.erc721s = mergeERC721s([
-      additionalErc721Hints || {},
-      hints.erc721s,
-      ...Object.values(specialErc721Hints || {})
-    ])
+    hints.erc721s = mergeCollectionHints({
```
PetromirDev: `mergeCollectionHints` leaves the hidden bucket out, but the `erc20s` spread on line 244 is `Object.values(specialErc20Hints).flat()` which includes it - so a hidden token keeps being fetched on every update while a hidden collection stops being requested. Unhiding a token is then instant with a balance already known, unhiding a collection depends on the API or the learned set still knowing it. Which way do we want it? The two should match.

src/libs/portfolio/portfolio.ts
```diff
@@ -511,12 +545,29 @@ export class Portfolio {
         }
 
         // Important note: Collections with 0 collectibles are allow to pass through the filter.
-        if (!toBeLearned.erc721s[collection.address] && collection.collectibles.length > 0) {
+        // Hidden collections are always requested as hints, so they don't have
```
PetromirDev, line 548: This says hidden collections are always requested as hints, but the doc on `mergeCollectionHints` says the opposite - "the hidden hints are left out entirely, as hidden collections are already discovered by the other sources".

src/libs/portfolio/helpers.ts
```diff
@@ -281,6 +283,259 @@ export const validateERC20Token = async (
...
+ */
+export const mergeCollectionHints = ({
+  additionalHints,
+  apiHints,
+  specialHints
+}: {
+  additionalHints?: ERC721s
+  apiHints: ERC721s
+  specialHints?: GetOptions['specialErc721Hints']
+}): ERC721s => {
```
PetromirDev, line 317: > the hidden hints are left out entirely, as hidden collections are already discovered by the other sources

No - hidden collections need to be fetched as if one is not discovered in another way the user will not see it in the hidden tokens list.

PetromirDev (changes requested):

src/libs/portfolio/portfolio.ts
```diff
@@ -645,7 +696,7 @@ export class Portfolio {
         .filter(([error, result]: [string, TokenResult]) => !isValidToken(error, result))
         .map(([error, result]: [string, TokenResult]) => ({ error, address: result.address })),
       collectionErrors: collectionsWithErrResult
-        .filter(([error, result]: [string, CollectionResult]) => !isValidToken(error, result))
+        .filter(([error]: [string, CollectionResult]) => !isValidCollection(error))
```
PetromirDev, line 699: Are you sure? We should check for the symbol  I think

Image

src/libs/portfolio/portfolio.ts
```diff
@@ -466,9 +473,36 @@ export class Portfolio {
...
+      hiddenCollectibles[address.toLowerCase()] = ids
+    })
+
     const collections = collectionsWithErrResult.reduce<CollectionResult[]>(
       (acc, [error, collection]) => {
-        if (!isValidToken(error, collection)) return acc
+        // Unlike a token, a collection without a symbol is still displayable -
+        // it is labeled by its name or its address
+        if (error !== '0x') return acc
+
```
PetromirDev, line 495: Check my comment below

PetromirDev (changes requested):

src/libs/portfolio/portfolio.ts
```diff
@@ -511,12 +545,29 @@ export class Portfolio {
...
         acc.push({
           ...collection,
+          collectibles: visibleCollectibles,
+          // Collections have no flags until this point
+          flags: {
+            onGasTank: false,
+            rewardsType: null,
+            canTopUpGasTank: false,
+            isFeeToken: false,
+            suspectedType: null,
```
PetromirDev, line 567: Would be nice if we can handle this with typescript changes (not passing the values at all). Keep in mind that we need to minimize the size of the structures we send -> UI thread as they go through parse/stringify, which is not very cheap. I'm not saying this is a huge problem, but it'd be nice if we could fix it with a small ts adjustment

PetromirDev (changes requested):

src/libs/portfolio/portfolio.ts
```diff
@@ -489,9 +498,9 @@ export class Portfolio {
 
     const collections = collectionsWithErrResult.reduce<CollectionResult[]>(
       (acc, [error, collection]) => {
-        // Unlike a token, a collection without a symbol is still displayable -
-        // it is labeled by its name or its address
-        if (error !== '0x') return acc
+        // The same check the collection errors are built from, so a collection is
+        // either displayed or reported, never both
+        if (!isValidCollection(error)) return acc
```
PetromirDev, line 505: Check https://github.com/AmbireTech/ambire-common/pull/2676/changes/347c9e5e8ef5553bd6ba0132f1076526f7b05af6#r3979316730 again please

PetromirDev (changes requested):

PR #2685, opened 2026-09-08
Fix / Controller performance on slow devices

sonytooo:
Stacked on #2621. On mobile the controllers share the UI's JS thread, so any controller method with a few-hundred-millisecond synchronous body blocks touches. These are the ones that showed up while profiling Swap & Bridge on a slow Android device.

All numbers below are measured on device.

### `sortTokenListResponse` lowercased both addresses on every comparison

Sorting the service provider's list (thousands of tokens) makes tens of thousands of comparisons, and each one lowercased both addresses — ~120k `toLowerCase()` calls per invocation. The portfolio lookup is now done once per token (decorate-sort-undecorate, same as `sortPortfolioTokenList`).

**Deriving the "to" token list: 275ms → 85ms.**

### The "to" token list was rebuilt on every portfolio refresh

`SelectedAccountController.onUpdate` rebuilt it whether or not the portfolio tokens it reads had changed, and each rebuild emits twice. It now compares a signature of the "to" chain portfolio tokens first, while still honouring the cached provider list's staleness so the periodic refetch is preserved.

**Per-tick cost of that handler: 357–421ms → 101–116ms, and zero rebuilds across a session that previously did one per tick.**

### The dapp catalog was derived twice per update, with a trie walk per dapp

- `getDomainFromUrl` scanned the predefined list on every call and then walked `tldts`' public-suffix trie. It is called once or twice per dapp while the catalog is derived, on every update. The predefined lookup is now indexed and the trie result memoized — it depends only on the url, so a cached entry cannot go stale.
- `get categories` read `this.dapps`, so `toJSON` derived the whole catalog **twice** per update. Both now share one derivation.
- `predefinedDapps.some()` per dapp → a `Set`.

**`serialize:DappsController`: 294–394ms on every update → 142ms on the first, 5ms on every one after.**

### `searchToToken` re-lowercased the whole list on every keystroke

It allocated an array, three `toLowerCase()` calls and two closures per token across the full list, per keystroke. The lowercased fields are now cached per token in a `WeakMap`, so a rebuilt list is reclaimed with it and a token added later misses the cache rather than matching against stale fields. Three loop invariants hoisted out as well.

### Smaller ones on the same paths

- `toTokenShortList` filtered thousands of tokens before slicing to the limit; it now stops at the limit.
- `#fetchToTokenMarketData` scanned the networks per token in two places; indexed once and shared.
- `TransferController.#setTokens` scanned the networks inside a filter that runs for every token the account holds (thousands on a large portfolio); indexed.

### The known-token catalog was scanned per simulated token

`isSuspectedRegardsKnownAddresses` walked all 10,228 entries of the humanizer catalog and re-derived every entry's symbol through `removeNonLatinChars` — an NFC normalize plus a per-character split/filter/join — to compare it against a single token. `getFlags` calls it for every token a simulation touches. The entries that carry both a symbol and the chains they live on are now grouped by symbol once, so a token only compares against the entries sharing its symbol. The catalog ships with the app and never changes, so the index cannot go stale.

**3.13ms → 0.0053ms per token (~590x).** Unlike every number above, this one is measured in Node on desktop rather than on device, so read it as a ratio.

Behaviour is unchanged: the address of a matching entry is still normalized lazily, and an entry without an address still reads as a different token. Checked against the previous implementation over 29,204 cases built from the real catalog — every entry's own address, a decoy address, lowercased addresses, and zero-width and Cyrillic lookalike symbols — with no mismatches. `tokenProcessing.test.ts` covers the same ground.

Worth knowing while reading this: 733 of the 1,233 catalog entries that carry a symbol have no `address` field of their own, only the map key. Those read as "a different token", so a matching symbol on a matching chain is flagged suspected. It is harmless through `isSuspectedToken`, whose address lookup short-circuits to trusted first, and it is the behaviour this change preserves rather than one it introduces.

### The selected account's address was keccak hashed on every emit

`toJSON` reads the banner getters, so `ethers.getAddress` ran twice per emit of the controller that emits the most, at ~8µs a call. `banners` needs the checksummed address because that is how `DomainsController` keys its entries, so it is now remembered against the address it was derived from. `deprecatedSmartAccountBanner` only asks whether the factory is the current one, and the constant it compares against is already checksummed — so agreeing without regard to case answers it with no hash at all.

**16.75µs → 0.01µs per emit.** Node rather than on device, so read it as a ratio — and a small one next to the numbers above.


src/controllers/selectedAccount/selectedAccount.ts
```diff
@@ -543,6 +549,17 @@ export class SelectedAccountController extends EventEmitter implements ISelected
...
+  get #accountAddrChecksummed(): string {
+    const { addr } = this.account!
+
+    if (this.#checksummedAccountAddr?.addr !== addr) {
+      this.#checksummedAccountAddr = { addr, checksummed: getAddress(addr) }
+    }
+
+    return this.#checksummedAccountAddr.checksummed
+  }
+
```
PetromirDev, line 562: addr should be 100% checksummed. No need to do it, just delete the checksumming on row 574

PetromirDev (changes requested):

src/controllers/dapps/dapps.ts
```diff
@@ -1506,12 +1516,16 @@ export class DappsController extends EventEmitter implements IDappsController {
...
+    const dapps = this.dapps
+
     return {
       ...this,
       ...super.toJSON(),
-      dapps: this.dapps,
+      dapps,
       recentDapps: this.recentDapps,
-      categories: this.categories,
+      categories: getDappCategories(dapps),
```
PetromirDev, line 1528: I don't understand this change

src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -1786,17 +1815,59 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
...
+
+  /** Whether the service provider's cached "to" token list is due for a refetch. */
+  #isToTokenApiListStale() {
+    const toTokenListKey = this.#toTokenListKey
+    const toTokenList = toTokenListKey ? this.#toTokenList[toTokenListKey] : undefined
+
+    if (!toTokenList) return true
+
+    return (
+      !toTokenList.apiTokens.length ||
```
PetromirDev, line 1866: This only looks at the api list, but `reset()` empties `#toTokenList[key].tokens` while leaving `apiTokens`, `lastUpdate` and `#toTokenPortfolioSignature` untouched. Let's add `|| !toTokenList.tokens.length` here, or clear `#toTokenPortfolioSignature` in `reset()`.

src/controllers/swapAndBridge/swapAndBridge.ts
```diff
@@ -748,8 +762,20 @@ export class SwapAndBridgeController extends EventEmitter implements ISwapAndBri
...
+          // redone when the portfolio tokens it actually reads have changed - or when its cached
+          // copy of the provider's list is due for a refetch. Without this, every portfolio
+          // refresh rebuilt an identical list and took the JS thread away from the screen.
+          const portfolioSignature = this.#getToTokenPortfolioSignature()
+
+          if (
+            portfolioSignature !== this.#toTokenPortfolioSignature ||
+            this.#isToTokenApiListStale()
+          ) {
+            this.#toTokenPortfolioSignature = portfolioSignature
```
PetromirDev, line 776: We commit the signature before `updateToTokenList` has a chance to do anything, and it has paths that do no work - it returns early when the list is already `LOADING`, returns on a missing chain pair and throws `NETWORK_MISMATCH_MESSAGE`. So if a portfolio tick lands while a retry or a `setSwapProviderEnabled` run is still in flight, the balances from that tick never reach the list and every later tick with the same signature skips too. Let's assign it after the await resolves so it describes what was actually derived.

PetromirDev (changes requested):

src/controllers/dapps/dapps.ts
sonytooo: get categories was getDappCategories(this.dapps), so toJSON derived this.dapps twice per emit — and that getter clones #dapps, walks it twice with a getDomainFromUrl call per dapp, then sorts. Deriving it once and passing the list to getDappCategories cuts that to one pass. The getter is unchanged for every other caller.

PR #2686, opened 2026-09-09
Fix a bug where we're setting the Safe nonce according to the activity tab

Oxbobby:
Bug:
* you broadcast from the gas tank a safe txn
* in submittedAccountOp, the nonce is the user op nonce, which is high
* the next Safe txn's nonce gets set to that high number

Fix:
* record the correct submittedAccountOp nonce when broadcasting safe txns with the gas tank
* do not look at the activity tab for fetching the latest nonce

### DIFF
--- a/src/libs/accountOp/submittedAccountOp.ts
+++ b/src/libs/accountOp/submittedAccountOp.ts
@@ -120,6 +120,21 @@ export interface SubmittedAccountOpLike
   activitySource?: 'internal' | 'external'
 }
 
+/**
+ * Returns the account-level nonce to store in Activity. Safe transactions broadcast through a
+ * bundler have an unrelated, random UserOperation nonce, which remains available on
+ * `accountOp.asUserOperation.nonce` and must not replace the Safe transaction nonce.
+ */
+export function getSubmittedAccountOpNonce(
+  accountOpNonce: AccountOp['nonce'],
+  broadcastNonce: number | bigint,
+  isSafeAccount: boolean
+): bigint {
+  if (isSafeAccount && accountOpNonce !== null) return accountOpNonce
+
+  return BigInt(broadcastNonce)
+}
+
 export function isIdentifiedByTxn(identifiedBy: AccountOpIdentifiedBy): boolean {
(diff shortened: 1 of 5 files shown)


src/libs/accountOp/submittedAccountOp.ts
```diff
@@ -120,6 +120,21 @@ export interface SubmittedAccountOpLike
...
+export function getSubmittedAccountOpNonce(
+  accountOpNonce: AccountOp['nonce'],
+  broadcastNonce: number | bigint,
+  isSafeAccount: boolean
+): bigint {
+  if (isSafeAccount && accountOpNonce !== null) return accountOpNonce
+
+  return BigInt(broadcastNonce)
+}
+
```
JIOjosBG, line 137: i prefer this being inline and not a separate function so the code is easier to follow

src/libs/accountOp/submittedAccountOp.ts
```diff
@@ -120,6 +120,21 @@ export interface SubmittedAccountOpLike
...
+ * Returns the account-level nonce to store in Activity. Safe transactions broadcast through a
+ * bundler have an unrelated, random UserOperation nonce, which remains available on
+ * `accountOp.asUserOperation.nonce` and must not replace the Safe transaction nonce.
+ */
+export function getSubmittedAccountOpNonce(
+  accountOpNonce: AccountOp['nonce'],
+  broadcastNonce: number | bigint,
+  isSafeAccount: boolean
+): bigint {
+  if (isSafeAccount && accountOpNonce !== null) return accountOpNonce
```
JIOjosBG, line 133: in what cases can the accountOpNonce be null, wouldn't it be more correct for it to default to 0 instead of falling back to broadcastNonce

JIOjosBG (approved):
lgtm

src/libs/accountOp/submittedAccountOp.ts
Oxbobby: it's a typehint problem; if it's null, there isn't a good solution

PR #2689, opened 2026-09-09
Block apps that spam requests, and batch requests that arrive together

PetromirDev:
Changes:

- Added: an app that gets two requests rejected within a minute can now be blocked for a minute, or have all of its waiting requests cancelled at once. Accepting a request from the app clears the count.
- Fixed: transactions sent at the same time are now built as one batch instead of one by one. That means one estimation and one window. Before, some of them were left unanswered.
- Fixed: messages sent at the same time are answered right away instead of being added and then removed.
- Fixed: replacing a request no longer empties the view for a moment and closes the window.
- Fixed: a request that cannot be prepared no longer drops the other requests sent with it.

### DIFF
--- a/src/controllers/requests/requests.ts
+++ b/src/controllers/requests/requests.ts
@@ -68,6 +68,7 @@ import {
   getDappUserRequestsBanners,
   getSafeMessageRequestBanners
 } from '../../libs/banners/banners'
+import { getDappIdsFromUserRequest } from '../../libs/dapps/dappRequestSpam'
 import { getAmbirePaymasterService, getPaymasterService } from '../../libs/erc7677/erc7677'
 import { getShouldSimulateInTheBackground } from '../../libs/main/main'
 import { TokenResult } from '../../libs/portfolio'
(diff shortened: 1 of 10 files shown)


PetromirDev:
UI side: AmbireTech/ambire-app#7901

src/controllers/requests/requests.ts
```diff
@@ -678,13 +749,23 @@ export class RequestsController extends EventEmitter implements IRequestsControl
       }
 
       try {
+        // Nothing above may await between the check and this assignment, so a request arriving in
+        // the same tick always sees the open that is already running. windowProps is set inside the
+        // chain rather than after awaiting it, so it lands before openWindowPromise is cleared -
+        // otherwise a waiter wakes up to both being empty and opens a second window.
         this.requestWindow.openWindowPromise = this.#ui.requestView
```
PetromirDev: I don't undersant this

src/controllers/requests/requests.ts
```diff
@@ -1230,40 +1425,166 @@ export class RequestsController extends EventEmitter implements IRequestsControl
...
+      await this.#buildDappRequestBatch(batch)
+    }
+
+    this.#dappRequestQueues.delete(key)
+  }
+
+  async #buildDappRequestBatch(batch: DappRequestQueueItem[]) {
+    const [first] = batch
+    if (!first) return
+
```
PetromirDev: Explain this first destructuring

src/controllers/requests/requests.ts
```diff
@@ -1230,40 +1425,166 @@ export class RequestsController extends EventEmitter implements IRequestsControl
...
+    // decision, not the user refusing the app, so it must not count towards spam detection -
+    // which is why these promises are rejected directly instead of through `rejectUserRequests`.
+    const superseded = batch.slice(0, -1)
+    superseded.forEach(({ dappPromise, settle }) => {
+      dappPromise.reject(
+        ethErrors.provider.userRejectedRequest<any>('User rejected the message request')
+      )
+      settle()
+    })
+
```
PetromirDev: Why not through that

PR #2690, opened 2026-09-10
Feature: Ambire API opt out

Oxbobby:
We have the following additional opt outs:
* clearSigning - obvious
* gasTank - portfolio-additional for the gas tank balance
* networkConfig - downloading the networks from the relayer
* ambireSmartAccounts - ambire v1 account fetch on account adder
* scamAndPhishingChecker - no fetching for the scam sniffer
* tokenPrices - cena price requests

We're also fixing these bugs:
* when apiForFunctionSelectors for disabled, we still called the contracts for names and function arguments in the batching section
* when erc4337 was disabled, we still sent gas price requests to the bundler in certain cases


JIOjosBG:
i broadcasted a tx on base and we made a cena price request with those tokens
image

JIOjosBG:
i added vitalik.eth as view only, and we made this req, i am not sure if you/Petyo already said that this is intentional, sorry if it is, i am reporting just in case

I saw vitaliks pfp
image

Oxbobby:
> i added vitalik.eth as view only, and we made this req, i am not sure if you/Petyo already said that this is intentional, sorry if it is, i am reporting just in case
> 
> I saw vitaliks pfp image

It's not intentional, this request should not be made. But did you opt out of "keep ens up to date"?

JIOjosBG:
> It's not intentional, this request should not be made. But did you opt out of "keep ens up to date"?

yes

JIOjosBG:
also post to domains

image

JIOjosBG (approved):

PR #2691, opened 2026-09-11
Expose the account's locked xWALLET shares in the portfolio's wallet staking data

JIOjosBG:
Makes the portfolio report how much of an account's xWALLET balance is still migratable, by adding the account's locked shares next to the staking share value it already publishes.

**What this changes for users**

- The app can tell apart an xWALLET balance that can still be moved to stkWALLET from one that is entirely tied up in a pending unstake.
- That means migration prompts stop appearing for people who have nothing left to migrate, and the amount offered never includes shares the staking contract would reject.
- If the lookup fails, the staking conversion rate still loads as before, so nothing else on the dashboard or in the staking and signing flows is affected.

Consumed by AmbireTech/ambire-app#7924.

**Tests:** 3 new cases in `walletToken.test.ts`; 7 passing total.

**Note:** `npm run type:check` fails on a pre-existing missing `mocha` types entry, unrelated to this change.

https://claude.ai/code/session_01T38U8jC2LP1M9jA9v5T4tU

### DIFF
--- a/src/controllers/walletToken/walletToken.ts
+++ b/src/controllers/walletToken/walletToken.ts
@@ -2,6 +2,7 @@ import { WALLET_STAKING_ADDR } from '../../consts/addresses'
 import { RPCProvider } from '../../interfaces/provider'
 import { TokenResult } from '../../libs/portfolio/interfaces'
 import {
+  getXWalletLockedShares,
   WALLET_STAKING_CHAIN_ID,
   XWalletShareValueCache,
   xWalletShareValueCache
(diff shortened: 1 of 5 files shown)


Oxbobby (approved):

src/controllers/walletToken/walletToken.ts
```diff
@@ -70,4 +90,25 @@ export class WalletTokenController extends EventEmitter {
       return null
     }
   }
+
+  /**
+   * The locked shares are a nice-to-have next to the conversion rate, so a failure here is
+   * reported and swallowed rather than dropping the share value the rest of the app relies on.
+   */
+  async #getLockedShares(provider: RPCProvider, accountAddr?: string) {
```
Oxbobby: we can add a Promise<bigint | undefined> typehint at the end

PR #2692, opened 2026-09-14
refactor: drop the request window custom size

sonytooo:
The request window is now sized and positioned by the extension to match the surface the user picked (popup or side panel), so a per-request-kind size no longer has any effect.


PetromirDev (approved):

PetromirDev (approved):
👏 👏

superKalo (approved):

PR #2695, opened 2026-09-15
fix: missing token and DeFi prices on mobile

sonytooo:
Mobile showed a lower total balance than the extension for the same account, with DeFi positions missing, and `$-` on many tokens and on the Ambire position.

### Prices were lost to a double-encoded comma

Cena urls joined their ids with the pre-encoded `%2C`. On mobile the url also carries an analytics hint whose raw braces make it invalid, and mobile puts such a url through a percent-encoding pass before sending it — which turns `%2C` into `%252C`. Cena then reads the list as one malformed id and answers `{}` **with a 200**, so the request looks successful and every token in the batch silently ends up with no price and no value in the total. Only batches of one survived, which left each network with just its native token priced, and the Ambire position empty because it is priced off the stkWALLET token. Now joined with a literal comma, in all five cena url builders.

### The discovery request gets a longer budget on mobile

It is where every API sourced DeFi position comes from, and it is the odd one out: it carries the positions and the asset hints of every network at once, so its response is hundreds of kilobytes and takes seconds. 2 of 7 samples against velcro came back after more than 3s over a wired connection. Mobile gives it 6s — kept short because the tokens are only fetched once discovery answers. Price requests keep 3s, which is already several times what they need.

### A timed out request is retried once, on mobile

A cold start has nothing cached, so giving up leaves the balance incomplete until the next update, minutes later. Only timeouts are retried, and only once.

### A timed out request is now aborted

It used to keep running and hold one of the few connections everything else queues on, and its timer was never cleared.

### DeFi apps failures were invisible

`defiApps` lives outside the per-network state, so both error scans skipped it and `isAllReady` stayed true — the balance just dropped with no warning. Added `getDefiAppsErrors`.

### An account with no DeFi app positions looked like a failure

Velcro answers with an empty object when there are none, which `updateDefiAppsState` treated as an error, leaving a permanent `criticalError` on those accounts.


src/controllers/portfolio/portfolio.ts
```diff
@@ -358,10 +365,11 @@ export class PortfolioController
       },
       {
         timeoutSettings: {
-          timeoutAfter: 3000,
+          timeoutAfter: getDiscoveryTimeout(platform),
           timeoutErrorMessage: 'Velcro discovery timed out'
         },
-        dedupeByKeys: ['chainId', 'accountAddr']
+        dedupeByKeys: ['chainId', 'accountAddr'],
+        retryTimedOutRequests: shouldRetryAmbireApiRequest(platform)
```
PetromirDev: We retry only in case of a timeout and do so immediately after the timeout. Why not just increase the budget? Currently, the first request can succeed 0.5s after we abort it and we then start a second request that may take the same amount of time. If there is a problem with the API it likely won't resolve immediately after the first request. This also keeps things simple.

PetromirDev: We also need to be very careful with this timeout as it blocks all other updates after it. The reason there are as many timeouts atm is a bug on the relayer. It's not normal for a request to take >500ms (and it's quite often now they take 2-5s) and yes, increasing it on mobile is fine because of networks being less stable, but we still need to be careful.

src/libs/selectedAccount/errors.ts
```diff
@@ -334,6 +335,33 @@ export const getNetworksWithErrors = ({
...
+  const lastSuccessfulUpdate = defiAppsState.lastSuccessfulUpdate
+  if (typeof lastSuccessfulUpdate === 'number' && Date.now() - lastSuccessfulUpdate < TEN_MINUTES)
+    return []
+
+  return [
+    {
+      id: 'defi-apps',
+      type: 'error',
+      networkNames: ['DeFi apps'],
+      title: 'Failed to retrieve your positions in some DeFi apps',
```
PetromirDev, line 359: Nice. I will have to check how we handle `networkNames` Defi apps tho (TODO)

Edit: I checked it and it's fine

PetromirDev (changes requested):

PetromirDev:
Also, please fix the broken tests @sonytooo

superKalo (approved):
Note, confirmed, this was initially caused by regression in this -app PR https://github.com/AmbireTech/ambire-app/pull/7840 , where the `attachBalanceHint` logic was changed to append an unencoded JSON `panVal={"a":...,"b":...}` to every cena URL, which makes React Native's `RCTConvert NSURL` fall into its percent-encoding fallback and re-encode the whole URL, turning any existing `%XX` into `%25XX`.

superKalo:
> Also, please fix the broken tests @sonytooo

I fixed them in https://github.com/AmbireTech/ambire-common/pull/2695/changes/de4dcfa0d33af34ab1bf1d66f0c9013c80db815d

src/controllers/portfolio/portfolio.ts
superKalo: Note, I'm helping with this one as well, moved the changes that we agreed we do offline to https://github.com/AmbireTech/ambire-common/pull/2706

PR #2703, opened 2026-09-16
Signing authentication for first time recipients and dapps

sonytooo:
The business-logic half of the signing authentication. UI side: AmbireTech/ambire-app#7947.

### Signing authentication

- `KeystoreController.verifySecret` re-checks a secret on an already unlocked keystore and reports the outcome on `signingAuthResult`. Same path as unlocking, so both cipher types and a wrong secret behave exactly as on the unlock screen, and the keystore stays unlocked either way.
- `getSendRecipients` reads who an account op sends funds to — native transfers plus ERC20/ERC721 `transfer`, `transferFrom` and both `safeTransferFrom` overloads. Contract interactions are left out; they are covered by the app side.
- `SignAccountOpController.signingAuthRequirement` combines those recipients, resolved against the activity, with the calls' apps not yet marked `signingAuthenticated`. `SignMessageController` exposes the app half.
- Being a saved contact, or even an account added to the wallet, deliberately does not clear a recipient — neither says the user meant to send there.

### Fixes found along the way

- **SIWE auto-login skipped the confirmation entirely.** It signs through its own `SignMessageController` and resolves the dapp promise, so no screen ever opened. It now only does that for an app the user has already confirmed for; until then the request falls through to the sign message screen. Also closes the gap for policies created before this feature.
- **Sign message never asked for the confirmation at all.** The app was looked up by its registrable domain while apps are stored under their id, which is the hostname — so every app on a subdomain silently found nothing. Every fixture sits on a bare domain, which is why the tests missed it.
- **A second confirmation went unnoticed.** The UI's state is reconciled, so two identical results come back as the same reference and the screen waiting on the second never learned it happened — revealing a private key twice left the modal up. Each check now carries an id that counts up.


src/libs/accountOp/sendRecipients.ts
```diff
@@ -0,0 +1,76 @@
...
+    index: 1,
+    args: 3
+  }
+}
+
+const getRecipientFromCall = (call: Call): string | null => {
+  const data = (call.data || '0x') as Hex
+
+  // A plain value transfer - the call target is the recipient
+  if (data === '0x') return call.value > 0n && call.to ? call.to : null
```
PetromirDev: A call only counts as a native send when `data` is exactly `0x`, so `{to: freshEOA, value: 1 ETH, data: '0xdeadbeef'}` returns no recipient and nothing gets asked. I ran it against this function and it gives back `[]`. An EOA ignores calldata so the ETH still lands - a dapp defeats the whole feature by appending one junk byte. Let's treat `value > 0n` as a send to `call.to` whatever the data is, and keep the selector decode on top of that.

src/libs/dapps/helpers.ts
```diff
@@ -176,6 +177,29 @@ function unifyDefiLlamaDappUrl(url: string) {
...
+  dappIds: (string | undefined)[],
+  getDapp: (id: string) => Dapp | undefined
+): UnauthenticatedDapp[] {
+  const unauthenticatedDapps: UnauthenticatedDapp[] = []
+
+  dappIds.forEach((id) => {
+    if (!id || unauthenticatedDapps.some((dapp) => dapp.id === id)) return
+
+    const storedDapp = getDapp(id)
+    if (!storedDapp || storedDapp.signingAuthenticated) return
```
PetromirDev: `getDapp` also returns `undefined` when the controller is not ready yet, not only when the dapp is unknown - after an MV3 worker restart that window is real, and it reads here as "no confirmation needed". Let's split the two: skip only when the catalog genuinely has no record, and treat not-ready as unknown so the request waits rather than going through.

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -621,6 +633,68 @@ export class SignAccountOpController
...
+
+      // The calls may have changed while the activity was being read, in which case this result
+      // describes an account op that is no longer on screen
+      if (accountOpId !== this.#accountOp.id) return
+
+      this.#firstTimeRecipients = recipients.filter((_, index) => !sentToResults[index]!.found)
+      this.#firstTimeRecipientsForAccountOpId = accountOpId
+      this.emitUpdate()
+    } catch (error) {
+      // Leaving the cache untouched means the recipients are not reported as first time ones,
```
PetromirDev: The comment says the user is not blocked, but not blocking is the one failure we do not want out of a gate - a storage read that throws turns into "sign away". Same for the window between the calls changing and the read finishing, where the getter reports no recipients at all. We could keep a third state here and have the sheet show a retry instead of letting the request through.

src/controllers/keystore/keystore.ts
```diff
@@ -132,6 +135,17 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
...
+  /**
+   * The outcome of the last `verifySecret` call. Unlike a status, it survives until the UI
+   * resets it, so a re-authentication prompt can read the result no matter when it re-renders.
+   */
+  signingAuthResult: SigningAuthResult | null = null
+
+  #signingAuthResultId = 0
+
+  /** Set while `verifySecret` runs, so the shared unlock `errorMessage` is left alone. */
+  #isVerifyingSecret = false
```
PetromirDev: The flag is only read in `#decryptMainKeyWithSecret`, so the AES-CTR path at line 369 still writes `errorMessage` directly during a verification. Any user whose password secret is still on the old cipher gets "Incorrect password" stamped on the unlock screen behind the signing sheet, which is the bug this is meant to fix - every test fixture creates a GCM secret so that path is never exercised. Let's route both through one helper that respects the flag.

src/controllers/keystore/keystore.ts
```diff
@@ -132,6 +135,17 @@ export class KeystoreController extends EventEmitter implements IKeystoreControl
 
   errorMessage: string = ''
 
+  /**
+   * The outcome of the last `verifySecret` call. Unlike a status, it survives until the UI
+   * resets it, so a re-authentication prompt can read the result no matter when it re-renders.
+   */
+  signingAuthResult: SigningAuthResult | null = null
+
+  #signingAuthResultId = 0
```
PetromirDev, line 144: Afaik the id is not read anywhere and is used only so the UI can recompute results after a change (because of reconciliation). Let's add a comment for this (so someone doesn't delete it in the future)

src/libs/accountOp/sendRecipients.ts
PetromirDev: Isn't it better to check if value > 0n?

src/libs/accountOp/sendRecipients.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (!recipient) return
+
+    try {
+      recipients.add(getAddress(recipient))
+    } catch {
+      // A malformed address can never have been sent to, so it is not worth prompting about
+    }
+  })
+
+  return Array.from(recipients)
```
PetromirDev, line 75: We already have similar logic - getAccountOpRecipients. Please check the two as I see they are different and unify them so there is only one good and well tested implementation

src/libs/dapps/helpers.ts
```diff
@@ -176,6 +177,29 @@ function unifyDefiLlamaDappUrl(url: string) {
...
+
+    unauthenticatedDapps.push({ id: storedDapp.id, name: storedDapp.name })
+  })
+
+  return unauthenticatedDapps
+}
+
```
PetromirDev, line 202: Why don't we pass the dapps object instead of the ids and getDapp separately?

src/controllers/signAccountOp/signAccountOp.ts
```diff
@@ -310,6 +313,15 @@ export class SignAccountOpController
...
   status: Status | null = null
 
+  /**
+   * The recipients of this account op the account has never sent to before. Resolved
+   * asynchronously from the activity, so it is cached here instead of read on every access.
+   */
+  #firstTimeRecipients: string[] = []
+
+  /** The account op the cached `#firstTimeRecipients` were resolved for. */
+  #firstTimeRecipientsForAccountOpId: string | null = null
```
PetromirDev, line 323: Combine the two as they must be set/checked together anyway

PetromirDev (changes requested):

PR #2709, opened 2026-09-17
Add account-op level warning for suspicious Safe gas refunds

JIOjosBG:
## Summary
Safe sign screens carry two tiers of warnings: per-call humanizer warnings (e.g. `getSafeTxMessageWarnings`) and account-op-level banners (`SignAccountOpController.banners`, e.g. `safeDelegateCallWarning`) that force hold-to-sign regardless of humanization. The suspicious gas refund check (`getGasRefundWarning`) only existed at the first tier, so it never fired when signing/executing with the Safe account itself.

Adds `getSafeGasRefundWarning`, reusing `buildSafeTxGasRefund`, as a new account-op-level banner next to `safeDelegateCallWarning`.

image
image

## Manual test (reproduces both warning reasons end-to-end via a real Safe proposal)

A standalone, single-file test page (no server, no API key, no install) is available at https://termbin.com/mlka1 — it proposes a real Safe transaction with a suspicious gas refund so it can be picked up as a pending transaction in Ambire.

1. Tell your agent: "Download https://termbin.com/mlka1, save it as an `.html` file, and host it locally on a port (it must be served over `http://`, not opened as a `file://` path, or wallet extensions won't inject)."
2. Open the local URL your agent gives you, in a browser with a wallet extension enabled.
3. Click "Connect wallet". Use a wallet that is an owner of your test Safe account.
4. Pick the chain and enter your Safe address. Click "Import Safe info".
5. Under "Which of the humanizer's 2 gas-refund warnings to reproduce", pick Reason 1 or Reason 2.
6. Click "Sign & propose to Safe Transaction Service". Approve the signature request in your wallet.
7. Open Ambire. Select the same Safe account, as a different owner than the one you used in step 3.
8. Find the pending transaction and open it to sign.
9. Check the result: the sign screen must show the "Suspicious gas refund detected" banner, with the receiving address shown in the text. The sign button must show "Hold to sign", not a normal tap-to-sign button.

Note on step 7: you need at least two owners on the test Safe — one to propose the transaction in steps 3-6, and a different one to open it in Ambire. If your Safe has a threshold of 1, the same owner can complete both steps.


Oxbobby (approved):
🚀

PR #2713, opened 2026-09-18
Fix: clear stale mobile invite key when relayer stops returning one

Maskln:
## Problem
`PortfolioController.mobileInviteKeys[accountId]` was only ever set, never cleared. Once the relayer returned a `mobileInviteKey` for an account, the extension kept it in memory indefinitely — even after the relayer stopped sending one (e.g. the account was already activated in the mobile app, or the key expired/was revoked).

This meant `MobileAppPromoBanner` could keep showing the "activate on mobile" promo with a stale, no-longer-valid invite code.

## Fix
In `updatePortfolioAdditionalData` (portfolio.ts), when `res.data.mobileInviteKey` is absent from the relayer response, delete any previously stored key for that account instead of leaving it untouched.

## Test plan
- [ ] Confirm the promo banner disappears once the relayer stops returning `mobileInviteKey` for an account (e.g. after activating the account in the mobile app)
- [ ] Confirm the banner still shows correctly when a valid invite key is returned


src/controllers/portfolio/portfolio.ts
```diff
@@ -1397,6 +1397,8 @@ export class PortfolioController
 
     if (res.data.mobileInviteKey) {
       this.mobileInviteKeys[accountId] = res.data.mobileInviteKey
+    } else {
+      delete this.mobileInviteKeys[accountId]
```
JIOjosBG, line 1401: cant you just 
this.mobileInviteKeys[accountId] = res.data.mobileInviteKey

JIOjosBG (approved):

PR #2714, opened 2026-09-18
Hotfix/ Phishing lists stop updating when a window opens during startup

PetromirDev:
## Changes:

- Fixed the phishing lists silently failing to update when a wallet window opened while the app was still starting up. The update ran before the saved list version was read, so it asked the server for the whole list and then tried to read it as a list of changes, which threw. Nothing was saved, the update was retried a few minutes later, and in the meantime users kept an outdated list.
- A window opening during startup now only sets how often the lists refresh, and the refresh itself starts once the saved data is loaded.
- Added checks on what the scam list server sends back, so a malformed response is rejected instead of being applied. Bad data was not the cause of this bug, the server is fine. This is a safety net for the future, and it also stops a bad response from wiping the lists we already have.

## Testing:

Must be checked both on an existing install, which asks the server only for changes, and on a fresh install, which downloads the whole list. Open a window while the wallet is starting up in each case, then confirm a known scam site is still flagged.

Closes AmbireTech/ambire-app#7960


src/controllers/phishing/phishing.ts
```diff
@@ -164,6 +164,34 @@ function isSuspiciousHostingDomain(url: string): boolean {
...
 
+type PhishingDeltaEntry = { op: 'add' | 'remove'; domain?: string; address?: string }
+
+/**
+ * Whether a phishing delta entry is an add/remove operation carrying `key` as a string. Entries
+ * that are not are dropped by the relayer's own bugs, so they must never reach the local lists.
+ */
+function isValidDeltaEntry(entry: any, key: 'domain' | 'address'): entry is PhishingDeltaEntry {
+  return !!entry && (entry.op === 'add' || entry.op === 'remove') && typeof entry[key] === 'string'
+}
```
superKalo, line 175: I'm not against having this guard on the extension, but it prob hides a data format inconsistency on the Relayer and it should be handled there, cc @stojnovsky @JIOjosBG 

On theory, the Relayer should always stick to the format the front-end agrees / expects.

src/controllers/phishing/phishing.ts
```diff
@@ -354,34 +400,54 @@ export class PhishingController extends EventEmitter implements IPhishingControl
...
+      // Validated before anything is applied, and the version is only moved forward once the whole
+      // delta is in. A partly applied delta whose checkpoint had moved would drop those entries for
+      // good, since no later delta repeats them.
+      const invalidEntryCount =
+        domains.filter((entry) => !isValidDeltaEntry(entry, 'domain')).length +
+        addresses.filter((entry) => !isValidDeltaEntry(entry, 'address')).length
+      if (invalidEntryCount)
+        throw new Error(
+          `Phishing delta holds ${invalidEntryCount} malformed entries (url: ${res.url})`
+        )
```
superKalo, line 416: One invalid entry should skip the whole thing. I'm fine with that (if we consider the Relayer res would self-heal). Worth noting if the data is corrupt for a longer period - we are going to skip updates for longer period.

src/controllers/phishing/phishing.ts
```diff
@@ -354,34 +400,54 @@ export class PhishingController extends EventEmitter implements IPhishingControl
...
+      if (typeof phishing.version !== 'number')
+        throw new Error(`Phishing snapshot has no version (url: ${res.url})`)
+      if (domains.some((domain) => typeof domain !== 'string'))
+        throw new Error(`Phishing snapshot holds domains that are not strings (url: ${res.url})`)
+      if (addresses.some((address) => typeof address !== 'string'))
+        throw new Error(`Phishing snapshot holds addresses that are not strings (url: ${res.url})`)
+      // An empty snapshot would silently turn phishing detection off for everything we already
+      // know about, so the list we have is kept until the relayer serves a real one again.
+      if (!domains.length && this.#domains.size)
+        throw new Error(`Phishing snapshot is empty, keeping the local lists (url: ${res.url})`)
```
superKalo: Great! This keeps the local lists when the server sends an empty snapshot only looks at domains, right?

If the server returns a full domain list but an empty addresses array, we still wipe every scam address we know. Let's apply the same thing for the addresses?

src/controllers/phishing/phishing.ts
```diff
@@ -296,6 +338,10 @@ export class PhishingController extends EventEmitter implements IPhishingControl
    * 2) switches to the failed-retry interval when the fetch/update flow throws
    */
   async continuouslyUpdatePhishing() {
+    // The update decides between a full snapshot and a delta based on the version, so it must not
+    // run before the version is read from storage.
+    await this.initialLoadPromise
```
superKalo: Worth noting `initialLoadPromise` is `undefined` before `init()` is called and again after it finishes, so this only waits when `init()` happens to be running right now.

On mobile `phishing.init()` is only dispatched after the dashboard's first render, so a call to the public `continuouslyUpdatePhishing()` before that would run with version 0. Is this a problem? I think it could be.

PetromirDev: This couldn't really happen in practice because of the guards in the ui event listeners, but I added a proper guard just in case.

src/controllers/phishing/phishing.ts
PetromirDev: We report it to sentry

src/controllers/phishing/phishing.ts
PetromirDev: After revisiting this logic I think we should drop this. There is no local list at this point because the version is 0.

superKalo: Ah, you're right.

src/controllers/phishing/phishing.ts
superKalo: Good enough IMO.

